PR #5, opened 2015-12-19
use django version '1.8.7' for base requirements

zubair-arbi:
@awais786 

Update base requirements for django, `django==1.8.7`


awais786:
:+1:

PR #8, opened 2015-12-29
add logger config in settings

zubair-arbi:
Add logger settings in `settings/utils.py` and import/use it in `local`, `test` and `production` settings.
@awais786 @ahsan-ul-haq @tasawernawaz @waheedahmed


ahsan-ul-haq:
:+1:

PR #9, opened 2015-12-29
remove unused logger settings

zubair-arbi:
Remove unused logger settings.
Related to: https://github.com/edx/credentials/pull/8
@ahsan-ul-haq @tasawernawaz @awais786 @waheedahmed


ahsan-ul-haq:
:+1:

PR #13, opened 2015-12-31
update duplicate attribute exists in db then update its value.

awais786:
Implement the https://openedx.atlassian.net/browse/ECOM-3324.
update the value of existing attribute or add a new one.

### DIFF
--- a/credentials/apps/api/tests/test_views.py
+++ b/credentials/apps/api/tests/test_views.py
@@ -37,6 +37,7 @@ class UserCredentialViewSetTests(APITestCase):
         self.user_credential = factories.UserCredentialFactory.create(credential=self.program_cert)
         self.user_credential_attribute = factories.UserCredentialAttributeFactory.create(
             user_credential=self.user_credential)
+        self.username = "test_user"
         self.request = APIRequestFactory().get('/')
 
     def _attempt_update_user_credential(self, data):
@@ -130,7 +131,7 @@ class UserCredentialViewSetTests(APITestCase):
         required fields are missing.
         """
         data = {
-            "username": "test_user",
+            "username": self.username,
             "credential": {"program_id": self.program_id},
             "attributes": [
                 {
@@ -157,9 +158,8 @@ class UserCredentialViewSetTests(APITestCase):
         """ Verify no UserCredential is created, and HTTP 400 is returned, if
         required fields are missing.
         """
-        username = "test_user"
         data = {
-            "username": username,
+            "username": self.username,
             "credential": {"program_id": self.program_id},
             "attributes": [
                 {
@@ -180,9 +180,8 @@ class UserCredentialViewSetTests(APITestCase):
     def test_create_with_programcertificate(self):
         """ Verify the endpoint supports issuing a new ProgramCertificate credential. """
         program_certificate = factories.ProgramCertificateFactory()
-        username = 'user2'
         data = {
-            "username": username,
+            "username": self.username,
             "credential": {
                 "program_id": program_certificate.program_id
             },
@@ -201,7 +200,7 @@ class UserCredentialViewSetTests(APITestCase):
         }
         response = self._attempt_create_user_credentials(data)
         self.assertEqual(response.status_code, 201)
-        user_credential = UserCredential.objects.get(username=username)
+        user_credential = UserCredential.objects.get(username=self.username)
         self.assertEqual(
             dict(response.data),
             dict(UserCredentialSerializer(user_credential, context={'request': self.request}).data)
@@ -221,9 +220,8 @@ class UserCredentialViewSetTests(APITestCase):
         """ Verify no UserCredential is created, and HTTP 400 is returned, if
         there are duplicated attributes.
         """
-        username = 'test_user'
         data = {
-            "username": username,
+            "username": self.username,
             "credential": {"program_id": self.program_id},
             "attributes": [
                 {
@@ -253,15 +251,14 @@ class UserCredentialViewSetTests(APITestCase):
         self.assertEqual(response.data, {'attributes': ['Attributes cannot be duplicated.']})
 
         self.assertEqual(response.status_code, 400)
-        self.assertFalse(UserCredential.objects.filter(username=username).exists())
+        self.assertFalse(UserCredential.objects.filter(username=self.username).exists())
 
     def test_create_with_empty_attributes(self):
         """ Verify no UserCredential is created, and HTTP 400 is returned, if
         there are some attributes are null.
         """
-        username = 'test_user_duplicate_attr'
         data = {
-            "username": username,
+            "username": self.username,
             "credential": {"program_id": self.program_id},
             "attributes": [
                 {
@@ -278,48 +275,11 @@ class UserCredentialViewSetTests(APITestCase):
         }
         response = self._attempt_create_user_credentials(data)
         self.assertEqual(response.status_code, 400)
-        self.assertFalse(UserCredential.objects.filter(username=username).exists())
+        self.assertFalse(UserCredential.objects.filter(username=self.username).exists())
         self.assertEqual(
             response.data.get('attributes')[1]['namespace'][0],
             'This field may not be blank.'
         )
-
-    def test_create_with_existing_usercredential(self):
-        """ Verify that, if a user has already been issued a credential, further
-        attempts to issue the same credential will NOT create a new credential
-        and will NOT modify the existing credential.
-        """
-        username = 'test_user'
-        data = {
-            "username": username,
-            "credential": {
-                "program_id": self.program_id
-            },
-            "attributes": [
-                {
-                    "namespace": "white",
-                    "name": "grade",
-                    "value": "0.5"
-                }
-            ]
-        }
-
-        user_credential_1 = factories.UserCredentialFactory.create(username=username, credential=self.program_cert)
-
-        # 2nd attempt to create credential with attributes.
-        msg = 'User [{username}] already has a credential for program [{program_id}].'.format(
-            username=username,
-            program_id=self.program_id
-        )
-
-        # Verify log is captured.
-        with LogCapture(LOGGER_NAME) as l:
-            response = self._attempt_create_user_credentials(data)
-            l.check((LOGGER_NAME, 'WARNING', msg))
-
-        self.assertEqual(response.status_code, 201)
-        user_credential_2 = UserCredential.objects.get(username=username)
-        self.assertEqual(user_credential_1, user_credential_2)
 
     def test_list_with_username_filter(self):
         """ Verify the list endpoint supports filter data by username."""
@@ -364,10 +324,9 @@ class UserCredentialViewSetTests(APITestCase):
         """ Verify no UserCredential is created, and HTTP 400 is return if credential
         id does not exists in db.
         """
-        username = 'test_user'
         cred_id = 10
         data = {
-            "username": username,
+            "username": self.username,
             "credential": {
                 "program_id": cred_id
             },
@@ -385,3 +344,88 @@ class UserCredentialViewSetTests(APITestCase):
             l.check((LOGGER_NAME_SERIALIZER, 'ERROR', msg))
 
         self.assertEqual(response.status_code, 400)
+
+    def test_reissue_the_user_credentials(self):
+        """ Verify that, if a user has already been issued a credential, further
+        attempts to issue the same credential will NOT create a new credential,
+        but its attributes will be updated if provided.
+        """
+        attributes = [
+            {"namespace": "whitelist", "name": "Program whitelist user", "value": "Reason for whitelisting."},
+            {"namespace": "grade", "name": "Final grade", "value": "0.5"}
+        ]
+
+        data = {
+            "username": self.username,
+            "credential": {
+                "program_id": self.program_cert.program_id
+            },
+            "attributes": attributes
+        }
+
+        # issue first credential for the user
+        response = self._attempt_create_user_credentials(data)
+        self.assertEqual(response.status_code, 201)
+        self._assert_usercredential_fields(response, self.username, attributes)
+
+        # change the attributes value
+        data["attributes"][0]["value"] = "New reason for whitelisting."
+        data["attributes"][1]["value"] = "0.8"
+
+        # try to issue credential again for the same user but with different attribute values and
+        # test that the existing record for user credential has been updated with new attribute values
+        response = self._attempt_create_user_credentials(data)
+        self.assertEqual(response.status_code, 201)
+        self._assert_usercredential_fields(response, self.username, attributes)
+
+    @ddt.data(
+        [{"namespace": "grade", "name": "Final grade", "value": "0.5"}],
+        [
+            {"namespace": "grade", "name": "Final grade", "value": "0.9"},
+            {"namespace": "grade", "name": "MidTerm grade", "value": "0.5"}
+        ],
+    )
+    def test_create_with_duplicate_attrs(self, attributes):
+        """ Verify that, if a user has a credential with attributes
+        then its values can be updated.
+        """
+        # create credential with attributes
+        user_credential = factories.UserCredentialFactory.create(
+            username=self.username,
+            credential=self.program_cert
+        )
+        factories.UserCredentialAttributeFactory(
+            user_credential=user_credential, namespace="grade", name="Final grade", value="0.75"
+        )
+        self.assertTrue(user_credential.attributes.exists())
+
+        data = {
+            "username": self.username,
+            "credential": {
+                "program_id": self.program_id
+            },
+            "attributes": attributes
+        }
+
+        # 2nd attempt to create credential with attributes.
+        response = self._attempt_create_user_credentials(data)
+        self.assertEqual(response.status_code, 201)
+        self._assert_usercredential_fields(response, self.username, attributes)
+
+    def _assert_usercredential_fields(self, response, username, expected_attrs):
+        """ Verify the fields on a UserCredential object match expectations. """
+
+        user_credential = UserCredential.objects.filter(username=username)
+        self.assertEqual(user_credential.count(), 1)
+        self.assertEqual(
+            dict(response.data),
+            dict(UserCredentialSerializer(
+                user_credential[0], context={'request': self.request}
+            ).data)
+        )
+
+        actual_attributes = [
+            {"namespace": attr.namespace, "name": attr.name, "value": attr.value}
+            for attr in user_credential[0].attributes.all()
+        ]
+        self.assertEqual(actual_attributes, expected_attrs)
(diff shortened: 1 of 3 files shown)


awais786:
@clintonb this is what u mentioned about replacing old attributes and replacing them with the new ones ?

awais786:
If that's the right behaviour than i will add few more test cases at view level also. If not that the right implementation than please explain.

credentials/apps/credentials/issuers.py
```diff
@@ -101,19 +101,9 @@ def set_credential_attributes(self, user_credential, attributes):
...
 
         for attr in attributes:
-            __, created = UserCredentialAttribute.objects.get_or_create(
+            UserCredentialAttribute.objects.get_or_create(
                 user_credential=user_credential,
                 namespace=attr.get('namespace'),
                 name=attr.get('name'),
                 defaults={'value': attr.get('value')}
             )
-            if not created:
```
clintonb: This works, but we need to decide if values should be updated.

awais786:
I will check this thoroughly tomorrow. defaults will not update the existing record i think. I will fix it tomorrow.

awais786:
@zubair-arbi please review this PR.

credentials/apps/api/tests/test_views.py
```diff
@@ -369,3 +332,118 @@ def test_create_with_non_existing_credential(self):
...
 
         self.assertEqual(response.status_code, 400)
+
+    def test_reissue_the_user_credentials(self):
+        """ Verify that, if a user has already been issued a credential, further
+        attempts to issue the same credential will NOT create a new credential.
+        But it's attributes will modify if provided.
+        """
+        username = 'user2'
+        attributes = [
```
zubair-arbi: Use attributes input like below for testing:
```python
attributes = [
    {"namespace": "whitelist", "name": "Program/Course whitelist user", "value": "Reason for whitelisting."},
    {"namespace": "grade", "name": "Final grade", "value": "80.50"}
]
```

credentials/apps/api/tests/test_views.py
```diff
@@ -369,3 +332,118 @@ def test_create_with_non_existing_credential(self):
...
+            },
+            "attributes": attributes
+        }
+
+        # first credential issue
+        response = self._attempt_create_user_credentials(data)
+        self.assertEqual(response.status_code, 201)
+        self._assert_usercredential_fields(response, username, attributes)
+
+        # change the attribute value
```
zubair-arbi: `# change the attributes value`

credentials/apps/api/tests/test_views.py
```diff
@@ -369,3 +332,118 @@ def test_create_with_non_existing_credential(self):
...
+
+        data = {
+            "username": username,
+            "credential": {
+                "program_id": self.program_cert.program_id
+            },
+            "attributes": attributes
+        }
+
+        # first credential issue
```
zubair-arbi: `# issue credentials for the user`

credentials/apps/api/tests/test_views.py
```diff
@@ -369,3 +332,118 @@ def test_create_with_non_existing_credential(self):
...
+        response = self._attempt_create_user_credentials(data)
+        self.assertEqual(response.status_code, 201)
+        self._assert_usercredential_fields(response, username, attributes)
+
+        # change the attribute value
+        data['attributes'][0]['value'] = '0.7'
+        data['attributes'][1]['value'] = '0.8'
+
+        # create 2nd credential issued. No credential will issue but attributes values
```
zubair-arbi: `# try to issue credential again for the same user but with different attribute values and test that the existing record for user credential has been updated with new attribute values`

credentials/apps/api/tests/test_views.py
```diff
@@ -369,3 +332,118 @@ def test_create_with_non_existing_credential(self):
...
+        actual_attributes = [
+            {'namespace': attr.namespace, 'name': attr.name, 'value': attr.value}
+            for attr in user_credential[0].attributes.all()
+        ]
+        self.assertEqual(actual_attributes, expected_attrs)
+
+    def test_create_credentials_with_multiple_programs(self):
+        """ Verify that, different programs credentials issued to a user."""
+
+        username = 'user2'
```
zubair-arbi: if there is only one user needed for a test then use `user` of `test_user` instead of `user2`

credentials/apps/api/tests/test_views.py
```diff
@@ -369,3 +332,118 @@ def test_create_with_non_existing_credential(self):
...
+        data['attributes'][1]['value'] = '0.8'
+
+        # create 2nd credential issued. No credential will issue but attributes values
+        # will be change.
+        response = self._attempt_create_user_credentials(data)
+        self.assertEqual(response.status_code, 201)
+        self._assert_usercredential_fields(response, username, attributes)
+
+    @ddt.data(
+        [{"namespace": "white", "name": "grade", "value": "0.5"}],
```
zubair-arbi: please use attributes input values like I mentioned above

credentials/apps/api/tests/test_views.py
```diff
@@ -369,3 +332,118 @@ def test_create_with_non_existing_credential(self):
...
+            {"namespace": "white1", "name": "grade", "value": "0.9"}
+        ],
+    )
+    def test_create_with_duplicate_attrs(self, attributes):
+        """ Verify that, if a user has a credential with attributes
+        then its values can be updated.
+        """
+        username = 'test_user'
+        # create credential with attributes
+        user_credential_1 = factories.UserCredentialFactory.create(
```
zubair-arbi: `user_credential_1` -> `user_credential`

credentials/apps/api/tests/test_views.py
```diff
@@ -369,3 +332,118 @@ def test_create_with_non_existing_credential(self):
...
+                {"namespace": "white1", "name": "grade", "value": "0.5"}
+            ]
+        }
+
+        # first credential issued
+        response = self._attempt_create_user_credentials(data)
+        self.assertEqual(response.status_code, 201)
+        self._assert_usercredential_fields(response, username, data["attributes"])
+
+        # create 2nd credential of different program.
```
zubair-arbi: `# create second credential for different program against the same user`

credentials/apps/api/tests/test_views.py
```diff
@@ -369,3 +332,118 @@ def test_create_with_non_existing_credential(self):
...
+        program_cert_2 = factories.ProgramCertificateFactory()
```
zubair-arbi: program_2_certificate = factories.ProgramCertificateFactory()

credentials/apps/api/tests/test_views.py
```diff
@@ -373,3 +336,118 @@ def test_create_with_non_existing_credential(self):
...
+            "attributes": attributes
+        }
+
+        # issue first credential for the user
+        response = self._attempt_create_user_credentials(data)
+        self.assertEqual(response.status_code, 201)
+        self._assert_usercredential_fields(response, username, attributes)
+
+        # change the attributes value
+        data['attributes'][0]['value'] = '0.7'
```
zubair-arbi: Since this value is related to `whitelist` namespace so it should be like `"New reason for whitelisting."`

credentials/apps/api/tests/test_views.py
```diff
@@ -373,3 +336,118 @@ def test_create_with_non_existing_credential(self):
...
+        data['attributes'][1]['value'] = '0.8'
```
zubair-arbi: `'0.8'` -> `85`

credentials/apps/api/tests/test_views.py
```diff
@@ -373,3 +336,118 @@ def test_create_with_non_existing_credential(self):
...
+        # try to issue credential again for the same user but with different attribute values and
+        # test that the existing record for user credential has been updated with new attribute values
+        response = self._attempt_create_user_credentials(data)
+        self.assertEqual(response.status_code, 201)
+        self._assert_usercredential_fields(response, username, attributes)
+
+    @ddt.data(
+        [{"namespace": "grade", "name": "Final grade", "value": "0.5"}],
+        [
+            {"namespace": "grade", "name": "Final grade", "value": "0.9."},
```
zubair-arbi: "0.9." -> "0.9"

credentials/apps/api/tests/test_views.py
```diff
@@ -373,3 +336,118 @@ def test_create_with_non_existing_credential(self):
...
+        then its values can be updated.
+        """
+        username = 'test_user'
+        # create credential with attributes
+        user_credential = factories.UserCredentialFactory.create(
+            username=username,
+            credential=self.program_cert
+        )
+        factories.UserCredentialAttributeFactory(
+            user_credential=user_credential, namespace="grade", name="Final grade", value=0.3
```
zubair-arbi: value is string `0.3` -> `"0.3"`

credentials/apps/api/tests/test_views.py
```diff
@@ -373,3 +336,118 @@ def test_create_with_non_existing_credential(self):
...
+        response = self._attempt_create_user_credentials(data)
+        self.assertEqual(response.status_code, 201)
+        self._assert_usercredential_fields(response, username, attributes)
+
+    def _assert_usercredential_fields(self, response, username, expected_attrs):
+        """ Verify the fields on a UserCredential object match expectations. """
+
+        user_credential = UserCredential.objects.filter(username=username)
+        self.assertEqual(user_credential.count(), 1)
+
```
zubair-arbi: remove this newline

credentials/apps/api/tests/test_views.py
```diff
@@ -373,3 +336,118 @@ def test_create_with_non_existing_credential(self):
...
+            {'namespace': attr.namespace, 'name': attr.name, 'value': attr.value}
+            for attr in user_credential[0].attributes.all()
+        ]
+        self.assertEqual(actual_attributes, expected_attrs)
+
+    def test_create_credentials_with_multiple_programs(self):
+        """ Verify that, different programs credentials issued to a user."""
+
+        username = 'test_user'
+
```
zubair-arbi: remove this newline

credentials/apps/api/tests/test_views.py
```diff
@@ -373,3 +336,118 @@ def test_create_with_non_existing_credential(self):
...
+        # create second credential for different program against the same user
+        program_2_certificate = factories.ProgramCertificateFactory()
+        data["credential"]["program_id"] = program_2_certificate.program_id
+        response = self._attempt_create_user_credentials(data)
+        self.assertEqual(response.status_code, 201)
+
+        user_credential = UserCredential.objects.filter(username=username)
+        self.assertEqual(user_credential.count(), 2)
+
```
zubair-arbi: rm newline

credentials/apps/credentials/tests/test_issuer.py
```diff
@@ -31,7 +29,7 @@ def setUp(self):
         self.user_program_cred = self.issuer.issue_credential(self.program_certificate, self.username)
         self.attributes = [
             {"namespace": "whitelist1", "name": "grade", "value": "0.5"},
-            {"namespace": "whitelist2", "name": "grade", "value": "0.5"}
+            {"namespace": "whitelist2", "name": "grade", "value": "0.6"}
```
zubair-arbi: Please update with correct dummy values

zubair-arbi:
:+1:

awais786:

credentials/apps/credentials/issuers.py
```diff
@@ -101,19 +90,9 @@ def set_credential_attributes(self, user_credential, attributes):
             raise DuplicateAttributeError("Attributes cannot be duplicated.")
```
jimabramson: @awais786 this is minor and slightly off-topic but I noticed while reviewing that in this line https://github.com/edx/credentials/blob/awais786/ECOM-2999-attrs-update/credentials/apps/credentials/utils.py#L18 the list of dictionaries does not seem to be getting sorted by the same critieria as the groupby key function, which appears to be needed according to https://docs.python.org/2/library/itertools.html#itertools.groupby .   I suspect this could lead to certain invalid inputs not getting caught depending on their original ordering.  If you agree this is an issue I think it's worth addressing (but in a separate PR would be fine)

credentials/apps/credentials/issuers.py
jimabramson: @clintonb do you mean deleted?  (is that supported?)

credentials/apps/api/tests/test_views.py
```diff
@@ -373,3 +332,111 @@ def test_create_with_non_existing_credential(self):
             l.check((LOGGER_NAME_SERIALIZER, 'ERROR', msg))
 
         self.assertEqual(response.status_code, 400)
+
+    def test_reissue_the_user_credentials(self):
+        """ Verify that, if a user has already been issued a credential, further
+        attempts to issue the same credential will NOT create a new credential.
+        But it's attributes will modify if provided.
```
jimabramson: ", but its attributes will be updated if provided."

credentials/apps/api/tests/test_views.py
```diff
@@ -373,3 +332,111 @@ def test_create_with_non_existing_credential(self):
...
+        self._assert_usercredential_fields(response, self.username, attributes)
+
+        # change the attributes value
+        data["attributes"][0]["value"] = "New reason for whitelisting."
+        data["attributes"][1]["value"] = "0.8"
+
+        # try to issue credential again for the same user but with different attribute values and
+        # test that the existing record for user credential has been updated with new attribute values
+        response = self._attempt_create_user_credentials(data)
+        self.assertEqual(response.status_code, 201)
```
jimabramson: 201 is a weird response status for a PATCH request.

credentials/apps/api/tests/test_views.py
```diff
@@ -373,3 +332,111 @@ def test_create_with_non_existing_credential(self):
...
+        """
+        # create credential with attributes
+        user_credential = factories.UserCredentialFactory.create(
+            username=self.username,
+            credential=self.program_cert
+        )
+        factories.UserCredentialAttributeFactory(
+            user_credential=user_credential, namespace="grade", name="Final grade", value="0.75"
+        )
+        self.assertTrue(user_credential.attributes.all().exists())
```
jimabramson: minor: `.all()` isn't needed, and is a little confusing

credentials/apps/api/tests/test_views.py
```diff
@@ -373,3 +332,111 @@ def test_create_with_non_existing_credential(self):
...
+
+        # create second credential for different program against the same user
+        program_2_certificate = factories.ProgramCertificateFactory()
+        data["credential"]["program_id"] = program_2_certificate.program_id
+        response = self._attempt_create_user_credentials(data)
+        self.assertEqual(response.status_code, 201)
+
+        user_credential = UserCredential.objects.filter(username=self.username)
+        self.assertEqual(user_credential.count(), 2)
+        self.assertEqual(dict(response.data), dict(UserCredentialSerializer(user_credential[1]).data))
```
jimabramson: since this makes an assumption about the ordering of results, I think you should explicitly order the queryset above on L440

credentials/apps/api/tests/test_views.py
```diff
@@ -373,3 +332,111 @@ def test_create_with_non_existing_credential(self):
...
+        self.assertEqual(dict(response.data), dict(UserCredentialSerializer(user_credential[0]).data))
+
+        actual_attributes = [
+            {"namespace": attr.namespace, "name": attr.name, "value": attr.value}
+            for attr in user_credential[0].attributes.all()
+        ]
+        self.assertEqual(actual_attributes, expected_attrs)
+
+    def test_create_credentials_with_multiple_programs(self):
+        """ Verify that, different programs credentials issued to a user."""
```
jimabramson: docstring is unclear - what's the motivation for this test?  it seems to be checking that it's possible to create more than one program credential for the same username, though that doesn't seem directly related to the purpose of this PR.

jimabramson:
@awais786 done reviewing - a few minor things inline.   two main concerns:

* if attributes will be updated via merge (patch) operations from now on, how can a client delete / unset existing attributes if desired?  Or is this even needed?  If not needed I don't think we should add handling, but I do think the behavior needs to be explicitly documented either way.
* maybe not introduced by this PR but it seems that PATCH requests are responding with 201/Created, which would be a misuse of REST conventions - let me know if I read that wrong.

credentials/apps/credentials/issuers.py
clintonb: Deleted and updated. My go-to use case is for a certificate re-issue. Say a student passes a course with a grade of 70% and we issue a certificate. If the certificate completes additional coursework and increases his grade to 80%, should that be reflected in the updated attributes? What if the grade somehow decreases (which I don't think we support)?

credentials/apps/api/tests/test_views.py
awais786: Yes. It's not related with this PR directly. Removing this test case now.

credentials/apps/api/tests/test_views.py
awais786: @jimabramson this is not patch request. This is post request but to handle the existing credentials attributes api updates the record. Instead of showing the error message.

credentials/apps/api/tests/test_views.py
awais786: I have removed this test case. It was not required in this PR. Reason i did dict() because both of the responses return ReturnDict which is subclass [OrderDict](https://docs.python.org/3/library/collections.html#collections.OrderedDict) of  and with same data and order they were not equal. That's why I used dict.

credentials/apps/api/tests/test_views.py
jimabramson: @awais786 sorry, my mistake.  was looking at the wrong method.

credentials/apps/api/tests/test_views.py
```diff
@@ -385,3 +344,88 @@ def test_create_with_non_existing_credential(self):
             l.check((LOGGER_NAME_SERIALIZER, 'ERROR', msg))
 
         self.assertEqual(response.status_code, 400)
+
+    def test_reissue_the_user_credentials(self):
+        """ Verify that, if a user has already been issued a credential, further
+        attempts to issue the same credential will NOT create a new credential
+        , but its attributes will be updated if provided.
```
jimabramson: nit: move the initial `,` to the end of the previous line

credentials/apps/credentials/issuers.py
jimabramson: right now it's simply last-update-wins, which is straightforward enough, but I agree with getting direction from stakeholders if things should work differently in some cases.

To throw out a case for deletion, I don't think you can presently un-whitelist someone after they've been whitelisted, because you wouldn't be able to delete that attribute.

jimabramson:
@awais786 i'm good to go on everything except the issue of updates / deletes which @clintonb also raised.

PR #14, opened 2016-01-01
add boto in prod requirements + add s3 storage settings for prod

zubair-arbi:
ECOM-3323

### DIFF
--- a/credentials/settings/base.py
+++ b/credentials/settings/base.py
@@ -37,6 +37,7 @@ THIRD_PARTY_APPS = (
     'social.apps.django_app.default',
     'waffle',
     'compressor',
+    'storages',
 )
 
 PROJECT_APPS = (
--- a/credentials/settings/production.py
+++ b/credentials/settings/production.py
@@ -12,6 +12,9 @@ ALLOWED_HOSTS = ['*']
 
 LOGGING = get_logger_config()
 
+# AMAZON S3 STORAGE CONFIGURATION
+# See: https://django-storages.readthedocs.org/en/latest/backends/amazon-S3.html#settings
+
 CONFIG_FILE = get_env_setting('CREDENTIALS_CFG')
 with open(CONFIG_FILE) as f:
     config_from_yaml = yaml.load(f)
@@ -28,3 +31,6 @@ DB_OVERRIDES = dict(
 
 for override, value in DB_OVERRIDES.iteritems():
     DATABASES['default'][override] = value
+
+if AWS_STORAGE_BUCKET_NAME and AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY:
+    DEFAULT_FILE_STORAGE = 'storages.backends.s3boto.S3BotoStorage'
--- a/credentials/urls.py
+++ b/credentials/urls.py
@@ -17,6 +17,7 @@ import os
 
 from django.conf import settings
 from django.conf.urls import include, url
+from django.conf.urls.static import static
 from django.contrib import admin
 from django.contrib.auth.views import logout
 from django.core.urlresolvers import reverse_lazy
@@ -42,6 +43,10 @@ urlpatterns = [
     url(r'^logout/$', logout, name='logout'),
     url('', include('social.apps.django_app.urls', namespace='social')),
 ]
+
+# Add media and extra urls
+if settings.DEBUG:  # pragma: no cover
+    urlpatterns += static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
 
 if settings.DEBUG and os.environ.get('ENABLE_DJANGO_TOOLBAR', False):  # pragma: no cover
     import debug_toolbar  # pylint: disable=import-error, wrong-import-position, wrong-import-order
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -23,6 +23,27 @@ If you need such overrides, create a file :file:`credentials/settings/private.py
 read by :file:`credentials/settings/local.py`, but ignored by Git.
 
 
+Configure the Amazon S3 Storage Backend
+---------------------------------------
+When you deploy the credentials on a staging or production server, you must change the settings to use the
+`Amazon S3 storage backend`_ instead of Django's default file storage backend
+(``django.core.files.storage.FileSystemStorage``) so that you do not commit changes to the repository.
+
+To change the settings, SSH into the system, and then update the :file:`/edx/etc/credentials.yml` file.
+This file's values are read by :file:`credentials/settings/production.py`, but ignored by Git.
+
+.. highlight:: yaml
+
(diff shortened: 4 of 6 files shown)


zubair-arbi:
@awais786 @ahsan-ul-haq @tasawernawaz I have tested this PR on my [sandbox](https://credentials-zubair-arbi.sandbox.edx.org/admin/) and its working. To override settings on `sandbox` and `prod` we just need to add following settings variables in the file `/edx/etc/credentials.yml` and restart worker `sudo /edx/bin/supervisorctl restart credentials:`
```yml
AWS_STORAGE_BUCKET_NAME: credentials-s3-bucket-name
AWS_ACCESS_KEY_ID: AAAAAAAAAAAAAAA
AWS_SECRET_ACCESS_KEY: BBBBBBBBBBBBBBBBBBBBB
AWS_DEFAULT_ACL: ''
```

awais786:
:+1:

zubair-arbi:
@clintonb please review.
fyi @feanil

clintonb:
@zubair-arbi, please find another reviewer. I'm maxed out on reviews and other development.

zubair-arbi:
@clintonb sure.
@feanil @rlucioni @jimabramson can any of you please review?

rlucioni:
@jimabramson should be able to take a look. Thank you!

jimabramson:
LGTM.   is this only for storing generated static assets or does it need to integrate with collectstatic as well?

zubair-arbi:
@jimabramson this is only for using S3 as default storage on prod so I believe its not needed to integrate it with collect static.

jimabramson:
:+1:

clintonb:
The storages app is not installed: http://django-storages.readthedocs.org/en/latest/#installation.

credentials/settings/production.py
```diff
@@ -12,6 +12,13 @@
 
 LOGGING = get_logger_config()
 
+# AMAZON S3 STORAGE CONFIGURATION
```
clintonb: These are coming from the YAML file. Why do they need to be declared here?

credentials/urls.py
```diff
@@ -41,6 +42,9 @@
     url('', include('social.apps.django_app.urls', namespace='social')),
 ]
 
+# Add media and extra urls
+urlpatterns += static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
```
clintonb: This should only be used for development: https://docs.djangoproject.com/en/1.9/howto/static-files/#serving-files-uploaded-by-a-user-during-development. Static files should be served from nginx for all other uses.

credentials/settings/base.py
```diff
@@ -127,6 +127,8 @@
     'compressor.finders.CompressorFinder',
 )
 
+DEFAULT_FILE_STORAGE = 'django.core.files.storage.FileSystemStorage'
```
clintonb: This is the default value. There is no need to set it.

credentials/settings/production.py
zubair-arbi: Just to give reference to the app users that these values will be need.

credentials/settings/base.py
zubair-arbi: Since we are using changing the default storage for S3 so I thought its useful to mention the default storage too

credentials/settings/production.py
clintonb: All you need is a comment with a URL. No need for unused variables.

credentials/settings/base.py
clintonb: If it's worth calling out, it should be in the documentation. Speaking of which...this change needs documentation.

credentials/urls.py
jimabramson: according to those docs, this won't actually work unless `DEBUG=True` anyway, but I agree with putting `if DEBUG` around it for clarity's sake.

docs/getting_started.rst
```diff
@@ -23,6 +23,26 @@ If you need such overrides, create a file :file:`credentials/settings/private.py
 read by :file:`credentials/settings/local.py`, but ignored by Git.
 
 
+Configure Amazon S3 storage backend on stage/production
+-------------------------------------------------------
+When deploying the credentials on stage/production you will need to override settings to use `Amazon S3 storage backend`_
```
zubair-arbi: @srpearce Please take a look on this piece of documentation.

zubair-arbi:
@srpearce Please review the documentation so that I can merge this PR.

docs/getting_started.rst
```diff
@@ -23,6 +23,26 @@ If you need such overrides, create a file :file:`credentials/settings/private.py
...
+instead of django's default file storage backend ``django.core.files.storage.FileSystemStorage`` that you do not wish
+to commit to the repository.
+If you need such overrides ssh into the system, update the file :file:`/edx/etc/credentials.yml`. This file's values are
+read by :file:`credentials/settings/production.py`, but ignored by Git.
+
+.. highlight:: yaml
+
+Here is an example for using Amazon S3 storage backend::
```
srpearce: The following example shows how to use the Amazon S3 storage backend

docs/getting_started.rst
```diff
@@ -23,6 +23,26 @@ If you need such overrides, create a file :file:`credentials/settings/private.py
 read by :file:`credentials/settings/local.py`, but ignored by Git.
 
 
+Configure Amazon S3 storage backend on stage/production
+-------------------------------------------------------
+When deploying the credentials on stage/production you will need to override settings to use `Amazon S3 storage backend`_
+instead of django's default file storage backend ``django.core.files.storage.FileSystemStorage`` that you do not wish
+to commit to the repository.
+If you need such overrides ssh into the system, update the file :file:`/edx/etc/credentials.yml`. This file's values are
+read by :file:`credentials/settings/production.py`, but ignored by Git.
```
srpearce: Suggested revision (please let me know if it's not technically accurate):

When you deploy the credentials on a staging or production server, you must change the settings to use the `Amazon S3 storage backend`_ instead of Django's default file storage backend (``django.core.files.storage.FileSystemStorage``) so that you do not commit changes to the repository. 

To change the settings, SSH into the system, and then update the :file:`/edx/etc/credentials.yml` file. This file's values are read by :file:`credentials/settings/production.py`, but ignored by Git.

docs/getting_started.rst
```diff
@@ -23,6 +23,26 @@ If you need such overrides, create a file :file:`credentials/settings/private.py
 read by :file:`credentials/settings/local.py`, but ignored by Git.
 
 
+Configure Amazon S3 storage backend on stage/production
```
srpearce: Configure the Amazon S3 Storage Backend

zubair-arbi:
@srpearce thanks for the review. I have made the suggested changes.

PR #15, opened 2016-01-01
change url from hyphen to underscore

ahsan-ul-haq:
@clintonb I have tried to fetch the data using this URL in lms [here](https://github.com/edx/edx-platform/pull/11064/files#diff-ad7d61197fa94d2095dccbb82e6ab748R39) but I got error as hyphen is not allowed so I change it. Is that okay or would you suggest something else?
FYI: @awais786 @zubair-arbi


clintonb:
Fine with me; but, you should add a comment explaining why "user-credentials" cannot be used, and linking to https://github.com/samgiles/slumber/issues/44.

PR #19, opened 2016-01-04
Added filter for program and course credentials APIs.

tasawernawaz:
ECOM-3263
Added following filters end points for program and course credentials.

* `/api/v1/program_credentials/?program_id=<program_id>`
* `/api/v1/program_credentials/?program_id=<program_id>&status=<status>`
* `/api/v1/course_credentials/?course_id=<course_id>`
* `/api/v1/course_credentials/?course_id=<course_id>&status=<status>`
* `/api/v1/course_credentials/?course_id=<course_id>&certificate_type=<certificate_type>`

Please review this @zubair-arbi @ahsan-ul-haq @awais786

### DIFF
--- a/credentials/apps/api/filters.py
+++ b/credentials/apps/api/filters.py
@@ -1,0 +1,28 @@
+"""
+Reusable queryset filters for the REST API.
+"""
+import django_filters
+from credentials.apps.credentials.models import UserCredential
+
+
+class ProgramFilter(django_filters.FilterSet):
+    """ Allows for filtering program credentials by their program_id and status
+    using a query string argument.
+    """
+    program_id = django_filters.NumberFilter(name="program_credentials__program_id")
+
+    class Meta:
+        model = UserCredential
+        fields = ['program_id', 'status']
+
+
+class CourseFilter(django_filters.FilterSet):
+    """ Allows for filtering course credentials by their course_id, status and
+    certificate_type using a query string argument.
+    """
+    course_id = django_filters.CharFilter(name="course_credentials__course_id")
+    certificate_type = django_filters.CharFilter(name="course_credentials__certificate_type")
+
+    class Meta:
+        model = UserCredential
+        fields = ['course_id', 'status', 'certificate_type']
--- a/credentials/apps/api/tests/test_views.py
+++ b/credentials/apps/api/tests/test_views.py
@@ -211,8 +211,7 @@ class UserCredentialViewSetTests(APITestCase):
         the unauthorized users to create a new user credential for the program.
         """
         self.client.logout()
-        path = self.list_path
-        response = self.client.post(path=path, data={}, content_type=JSON_CONTENT_TYPE)
+        response = self.client.post(path=self.list_path, data={}, content_type=JSON_CONTENT_TYPE)
 
         self.assertEqual(response.status_code, 403)
 
@@ -284,9 +283,7 @@ class UserCredentialViewSetTests(APITestCase):
     def test_list_with_username_filter(self):
         """ Verify the list endpoint supports filter data by username."""
         factories.UserCredentialFactory(username="dummy-user")
-        path = self.list_path + "?username=" + self.user_credential.username
-
-        response = self.client.get(path)
+        response = self.client.get(self.list_path, data={'username': self.user_credential.username})
         self.assertEqual(response.status_code, 200)
 
         # after filtering it is only one related record
@@ -299,16 +296,12 @@ class UserCredentialViewSetTests(APITestCase):
     def test_list_with_status_filter(self):
         """ Verify the list endpoint supports filtering by status."""
         factories.UserCredentialFactory.create_batch(2, status="revoked", username=self.user_credential.username)
-        path = self.list_path + "?status={}" + self.user_credential.status
-
-        response = self.client.get(path)
+        response = self.client.get(self.list_path, data={'status': self.user_credential.status})
         self.assertEqual(response.status_code, 400)
 
         # username and status will return the data.
-        path = self.list_path + "?username={user}&status={status}".format(
-            user=self.user_credential.username,
-            status=UserCredential.AWARDED)
-        response = self.client.get(path)
+        response = self.client.get(self.list_path,
+                                   data={'username': self.user_credential.username, 'status': UserCredential.AWARDED})
 
         # after filtering it is only one related record
         expected = UserCredentialSerializer(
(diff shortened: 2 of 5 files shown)


credentials/apps/api/filters.py
```diff
@@ -0,0 +1,35 @@
...
+    """
+    program_id = django_filters.NumberFilter(name="program_credentials__program_id")
+
+    class Meta:
+        model = UserCredential
+        fields = ['program_id', 'status']
+
+
+class CourseFilter(django_filters.FilterSet):
+    """ Allows for filtering course credentials by their course_id and status
```
zubair-arbi: add `course_credentials` too

credentials/apps/api/filters.py
```diff
@@ -0,0 +1,35 @@
...
+    using a query string argument.
+    """
+    def filter_course_id(self, value):
+        """ Helper method that replace + signs with spaces."""
```
zubair-arbi: Please add comment why you are doing this.

credentials/apps/api/filters.py
```diff
@@ -0,0 +1,35 @@
...
+
+class CourseFilter(django_filters.FilterSet):
+    """ Allows for filtering course credentials by their course_id, status and
+    certificate_type using a query string argument.
+    """
+    def filter_course_id(self, value):
+        """ Helper method that replace + signs with space, for correct course_id format."""
+        # pylint: disable=no-member
+        return self.filter(
+            course_credentials__course_id=value.replace(' ', '+')
```
zubair-arbi: Can we do this in some better way like decoding the values?

zubair-arbi:
Just one question, otherwise :+1:

credentials/apps/api/filters.py
tasawernawaz: Django convert '+' sign into ' ' by url encoding. I think we need to convert it back into '+' sign for matching course_id.

tasawernawaz:
@jimabramson Please review this.

credentials/apps/api/filters.py
jimabramson: I agree with @zubair-arbi - we should require that the parameter value is urlencoded on the incoming request - is there a reason we can't do that?

credentials/apps/api/tests/test_views.py
```diff
@@ -373,3 +373,137 @@ def test_create_with_non_existing_credential(self):
...
+            'error': 'A program_id query string parameter is required for filtering program credentials.'
+        })
+
+    def test_list_with_program_id_filter(self):
+        """ Verify the list endpoint supports filter data by program_id."""
+
+        program_cert = factories.ProgramCertificateFactory(program_id=001)
+        factories.UserCredentialFactory.create(credential=program_cert)
+
+        path = self.list_path + "?program_id=" + str(self.program_id)
```
jimabramson: instead of manually assembling a query string you can pass the parameters as a dict in a `data` kwarg to `self.client.get` on the next line

credentials/apps/api/tests/test_views.py
```diff
@@ -373,3 +373,137 @@ def test_create_with_non_existing_credential(self):
...
+        course_id filter. Otherwise it will return 400.
+        """
+        response = self.client.get(self.list_path)
+        self.assertEqual(response.status_code, 400)
+        self.assertEqual(response.data, {
+            'error': 'A course_id query string parameter is required for filtering course credentials.'
+        })
+
+    def test_list_with_course_id_filter(self):
+        """ Verify the list endpoint supports filter data by program_id."""
```
jimabramson: s/program_id/course_id/

credentials/apps/api/v1/views.py
```diff
@@ -33,3 +34,37 @@ def list(self, request, *args, **kwargs):
...
         return super(UserCredentialViewSet, self).create(request, *args, **kwargs)
+
+
+class ProgramsCredentialsViewSet(mixins.ListModelMixin, viewsets.GenericViewSet):
+    """It will return the all credentials for programs."""
+    queryset = UserCredential.objects.all()
+    filter_backends = (filters.DjangoFilterBackend,)
+    filter_class = ProgramFilter
+    serializer_class = UserCredentialSerializer
+    permission_classes = (DjangoModelPermissionsOrAnonReadOnly,)
```
jimabramson: should anonymous read-only access be allowed for this data?

credentials/apps/api/tests/test_views.py
```diff
@@ -373,3 +373,137 @@ def test_create_with_non_existing_credential(self):
             l.check((LOGGER_NAME_SERIALIZER, 'ERROR', msg))
 
         self.assertEqual(response.status_code, 400)
+
+
+class ProgramCredentialViewSetTests(APITestCase):
```
jimabramson: @tasawernawaz most of the two test cases for the two viewsets added here are basically the same, could you look into a way of reusing the test code?   For example, the methods except for `test_list_with_certificate_type` could be implemented in a Mixin and you could have both test cases inherit from it.

credentials/apps/api/v1/views.py
tasawernawaz: @jimabramson I have discussed it with @zubair-arbi, We don't need permissions for list end-points. So I am removing `permission_classes` from  `ProgramsCredentialsViewSet` and `CourseCredentialsViewSet`.

credentials/apps/api/filters.py
tasawernawaz: @jimabramson @zubair-arbi 
I am replacing space with plus because when I try to make request from browser, in view I receive `course_id` like this `course-v1:ASUx AST111 3T2015` instead of `course-v1:ASUx+AST111+3T2015`. 
But in tests if I try to send request with client.get() I receive correct parameter value like this `course-v1:ASUx+AST111+3T2015`.
Can you help me on which level django is doing this, where I can replace or prevent it from converting plus into space.

Reference: [same problem](http://stackoverflow.com/questions/29703930/testing-request-parameters-in-django-behaves-differently).

tasawernawaz:
@jimabramson Made suggested changes, Please review this.

credentials/apps/api/filters.py
jimabramson: @tasawernawaz when you make the request from a browser, are you urlencoding the parameter?  Since course keys (course ids) are not strictly URL-safe, this is necessary.  Your browser request for that course id should look like this:
```
?course_id=course-v1%3AASUx%2BAST111%2B3T2015
```
This is how all clients should be making HTTP requests to this API.

jimabramson: And, in fact, as long as the request parameter is correctly urlencoded as in the example in my previous comment, you should not need to do any manual urlencoding or urldecoding in tests or in filter code.  Django will automatically do this.

credentials/apps/api/tests/test_views.py
```diff
@@ -429,3 +422,121 @@ def _assert_usercredential_fields(self, response, username, expected_attrs):
...
+
+    def assert_list_without_id_filter(self, path, expected):
+        """Helper method used for making request and assertions. """
+        response = self.client.get(path)
+        self.assertEqual(response.status_code, 400)
+        self.assertEqual(response.data, expected)
+
+    def assert_list_with_id_filter(self, data):
+        """Helper method used for making request and assertions. """
+
```
jimabramson: remove blank line

credentials/apps/api/tests/test_views.py
```diff
@@ -429,3 +422,121 @@ def _assert_usercredential_fields(self, response, username, expected_attrs):
...
+        """
+        self.assert_list_without_id_filter(path=self.list_path, expected={
+            'error': 'A program_id query string parameter is required for filtering program credentials.'
+        })
+
+    def test_list_with_program_id_filter(self):
+        """ Verify the list endpoint supports filter data by program_id."""
+        program_cert = factories.ProgramCertificateFactory(program_id=001)
+        factories.UserCredentialFactory.create(credential=program_cert)
+
```
jimabramson: remove blank line

jimabramson:
@tasawernawaz tiny nitpicks.   :+1: after removing the `' '` -> `'+'` replacement (unless we need to discuss it more)

PR #20, opened 2016-01-05
Adding sorting using same keys on list and groupby.

awais786:
ECOM-3343
Adding sorting using same keys on list and groupby.

### DIFF
--- a/credentials/apps/credentials/utils.py
+++ b/credentials/apps/credentials/utils.py
@@ -15,7 +15,11 @@ def validate_duplicate_attributes(attributes):
         Boolean: Return True if data has no duplicated namespace and name otherwise False
 
     """
-    for __, group in groupby(sorted(attributes), lambda x: (x['namespace'], x['name'])):
+    def keyfunc(attribute):  # pylint: disable=missing-docstring
+        return attribute['namespace'], attribute['name']
+
+    sorted_data = sorted(attributes, key=keyfunc)
+    for __, group in groupby(sorted_data, key=keyfunc):
         if len(list(group)) > 1:
             return False
     return True


awais786:
@zubair-arbi please review this PR.

credentials/apps/credentials/tests/test_utils.py
```diff
@@ -9,21 +9,38 @@
 @ddt.ddt
 class ValidateDuplicateAttributesTests(TestCase):
     """ Tests for Validate the attributes method """
-
-    def test_with_non_duplicate_attributes(self):
+    @ddt.data(
+        [
+            {"namespace": "whitelist", "name": "programs", "value": "0.9"},
```
zubair-arbi:
```python
[
    {"namespace": "whitelist", "name": "program", "value": "Reason for whitelist"},
    {"namespace": "whitelist", "name": "course", "value": "Reason for whitelist"},
]
```

credentials/apps/credentials/tests/test_utils.py
```diff
@@ -9,21 +9,38 @@
...
+            {"namespace": "whitelist", "name": "courses", "value": "0.9"},
+        ],
+        [
+            {"name": "programs", "namespace": "whitelist", "value": "0.9"},
```
zubair-arbi: order the keys for input

awais786: This is what I am testing.

zubair-arbi:
:+1:

awais786:

credentials/apps/credentials/tests/test_utils.py
```diff
@@ -9,21 +9,38 @@
...
+        ],
+        [
+            {"name": "program", "namespace": "whitelist", "value": "Reason1"},
+            {"namespace": "whitelist", "name": "program", "value": "Reason2"},
+        ],
+        [
+            {"value": "Reason1", "name": "program", "namespace": "whitelist"},
+            {"namespace": "whitelist", "name": "program", "value": "Reason2"},
+        ],
+    )
```
jimabramson: @awais786 these inputs are redundant because the pairs of dict values are equal.   The way Python compares dictionaries is described here https://docs.python.org/2/reference/expressions.html#not-in ... the order in which the keys/values are declared when creating the dictionary makes no difference to how it will be sorted.

here is an example set of inputs that would have unexpected results without your fix:

```
>>> dicts = [
...     {'name': 'a', 'namespace': 'n1', 'value': 'first'},
...     {'name': 'b', 'namespace': 'n1', 'value': 'second'},
...     {'name': 'a', 'namespace': 'n2', 'value': 'third'},
...     {'name': 'b', 'namespace': 'n2', 'value': 'fourth'},
... ]
>>> keyfunc = lambda d: (d['namespace'], d['name'])
>>>
>>> # buggy
>>> [list(v[1]) for v in groupby(sorted(dicts), key=keyfunc)]
[[{'namespace': 'n1', 'name': 'a', 'value': 'first'}], [{'namespace': 'n2', 'name': 'a', 'value': 'third'}], [{'namespace': 'n1', 'name': 'b', 'value': 'second'}], [{'namespace': 'n2', 'name': 'b', 'value': 'fourth'}]]
>>>
>>> # fixed
>>> [list(v[1]) for v in groupby(sorted(dicts, key=keyfunc), key=keyfunc)]
[[{'namespace': 'n1', 'name': 'a', 'value': 'first'}], [{'namespace': 'n1', 'name': 'b', 'value': 'second'}], [{'namespace': 'n2', 'name': 'a', 'value': 'third'}], [{'namespace': 'n2', 'name': 'b', 'value': 'fourth'}]]

Comment (jimabramson) [inline credentials/apps/credentials/tests/test_utils.py]: @awais786 because of the natural sorting of the key names in these dicts, a bug wouldn't happen (the previous example I posted wouldn't actually break the logic even though the sort order is different).  However I was able to come up with an example which would actually cause a bug (i.e. it would fail to detect duplicate attributes) if the dictionaries might contain other keys:

```
>>> dicts = [
...     {'a': 1, 'name': 'a', 'namespace': 'n1', 'value': 'first'},
...     {'a': 1, 'name': 'b', 'namespace': 'n1', 'value': 'second'},
...     {'a': 0, 'name': 'a', 'namespace': 'n2', 'value': 'third'},
...     {'a': 0, 'name': 'b', 'namespace': 'n2', 'value': 'fourth'},
...     {'a': 2, 'name': 'b', 'namespace': 'n2', 'value': '?'},
... ]
>>> # misses duplicate
>>> [len(list(v[1])) for v in groupby(sorted(dicts), key=keyfunc)]
[1, 1, 1, 1, 1]
>>> # catches duplicate
>>> [len(list(v[1])) for v in groupby(sorted(dicts, key=keyfunc), key=keyfunc)]
[1, 1, 1, 2]
```
I admit this is a far-off edge case.  With your change in `utils.py` we are safe from such a bug even if new keys are added to the UserCredentialAttributeSerializer, so my recommendation is actually just to revert your modifications the tests, but leave the fix in utils.py in place.  (It doesn't seem worth it to test an unrealistic scenario just to prove that the usage of `groupby` is correct and bug-proof.)

Thanks for indulging my concern over this obscure detail...

jimabramson:
:+1: after removing changes to tests (assuming you agree with this recommendation)

awais786:
@jimabramson thanks for sending the buggy scenario data. I have reverted the test changes.

PR #23, opened 2016-01-06
fix the NameError on aws s3 settings

zubair-arbi:
@awais786 @tasawernawaz @ahsan-ul-haq

### DIFF
--- a/credentials/settings/production.py
+++ b/credentials/settings/production.py
@@ -14,6 +14,9 @@ LOGGING = get_logger_config()
 
 # AMAZON S3 STORAGE CONFIGURATION
 # See: https://django-storages.readthedocs.org/en/latest/backends/amazon-S3.html#settings
+AWS_STORAGE_BUCKET_NAME = ''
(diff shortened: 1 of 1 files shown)


zubair-arbi:
@clintonb we have to define these variables somewhere. If these variables are not in `credentials.yml` file then we will get exception `NameError: name 'AWS_STORAGE_BUCKET_NAME' is not defined`.

clintonb:
If they need to be in credentials.yml, make sure they are always there. I would much rather the app fail to start when settings are missing than discover we are using empty strings later in production.

zubair-arbi:
Thanks for the feedback. I have already added these variables in config PR: https://github.com/edx/configuration/commit/e1ea1165c864361b726d9ce335b76a7b573d46ac
But just wanted to make sure if we need to suppress this error or not.

clintonb:
No. Errors are a good thing when caught early. I actually want more of our apps to run startup checks and fail if data does not meet expectations. If you need the data in an existing sandbox, manually add it.

PR #26, opened 2016-01-09
get organization and user data from in lms

zubair-arbi:
ECOM-3282

**WIP**

* Create settings variables required for communication with the `organizations` API in `LMS`
* Add `get_organization` method to get an organization from edx rest api client with the provided `organization_key`

**Remaining changes**
* Create settings variables required for communication with the `users` API in `LMS`
* Add `get_user` method to get a user data from edx rest api client with the provided `username`
* Add tests for data retrieval from `organizations` api
* Add tests for data retrieval from`users` api

### DIFF
--- a/credentials/apps/credentials/utils.py
+++ b/credentials/apps/credentials/utils.py
@@ -1,7 +1,19 @@
 """
-Helper methods for data validation.
+Helper methods for credential app.
 """
+import datetime
+import logging
 from itertools import groupby
+from urlparse import urljoin
+
+import jwt
+from django.conf import settings
+from django.core.cache import cache
+from edx_rest_api_client.client import EdxRestApiClient
+
+
+log = logging.getLogger(__name__)
+PROGRAMS_CACHE_KEY = 'programs.api.data'
 
 
 def validate_duplicate_attributes(attributes):
@@ -9,10 +21,10 @@ def validate_duplicate_attributes(attributes):
     Validate the attributes data
 
     Arguments:
-            attributes (list): List of dicts contains attributes data.
+        attributes (list): List of dicts contains attributes data
 
     Returns:
-        Boolean: Return True if data has no duplicated namespace and name otherwise False
+        Boolean: Return True only if data has no duplicated namespace and name
 
     """
     def keyfunc(attribute):  # pylint: disable=missing-docstring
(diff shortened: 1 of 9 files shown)


zubair-arbi:
fyi: @jimabramson @clintonb

credentials/apps/credentials/utils.py
```diff
@@ -7,6 +7,7 @@
 from urlparse import urljoin
 
 import jwt
+import requests
```
clintonb: This is not used.

credentials/settings/base.py
```diff
@@ -213,8 +213,18 @@
 PROGRAMS_API_URL = '/api/v1/'
 # Specified in seconds. Enable caching by setting this to a value greater than 0.
 PROGRAMS_CACHE_TTL = 0
-# Credentials service user in Programs service
-PROGRAMS_SERVICE_USER = 'credentials_service_user'
+
+# Absolute URL used to construct URLs pointing to the LMS
+LMS_URL_ROOT = 'http://127.0.0.1:8000'
+# Organizations API url in edx-platform
+ORGANIZATIONS_API_URL = '/api/organizations/v0/'
```
clintonb: This should be a complete URL, not a path. That way, if we ever move the service to an IDA, your code does not need to change. I would do the same for `PROGRAMS_API_URL` as well.

credentials/apps/credentials/utils.py
```diff
@@ -122,3 +183,24 @@ def _get_service_user_api_client(api_url, service_username):
...
+    Arguments:
+        api_url (str): Absolute url of the api
+
+    """
+    try:
+        api_client = EdxRestApiClient(
+            get_organizations_api_url(),
+            oauth_access_token=settings.CREDENTIALS_OAUTH_ACCESS_TOKEN
+        )
+    except Exception:  # pylint: disable=broad-except
```
clintonb: When/why would an exception be raised? Are there less-broad exceptions you can catch?

credentials/apps/credentials/utils.py
```diff
@@ -122,3 +183,24 @@ def _get_service_user_api_client(api_url, service_username):
...
+    Helper method to get edx rest api client for the credentials service user
+    which is present on the LMS.
+
+    Arguments:
+        api_url (str): Absolute url of the api
+
+    """
+    try:
+        api_client = EdxRestApiClient(
+            get_organizations_api_url(),
```
clintonb: This should be `api_url`. Remove the `branches` directive in .travis.yml so that tests are run on this PR: https://github.com/edx/credentials/blob/master/.travis.yml#L22-L24.

credentials/apps/credentials/utils.py
```diff
@@ -122,3 +183,24 @@ def _get_service_user_api_client(api_url, service_username):
         return
 
     return api_client
+
+
+def _get_lms_service_user_api_client(api_url):
```
clintonb: Why is this necessary? Do you plan to use it elsewhere?

credentials/settings/base.py
```diff
@@ -213,8 +213,18 @@
...
+# Specified in seconds. Enable caching by setting this to a value greater than 0.
+ORGANIZATIONS_CACHE_TTL = 0
```
clintonb: This should default to a sane production default (e.g. 30 minutes). You can set it to a lower development default in local.py.

credentials/apps/credentials/utils.py
```diff
@@ -77,6 +79,49 @@ def get_programs():
...
+
+    Arguments:
+        organization_key (str): Unique key of the organization for retrieval
+
+    Returns:
+        dict, representing organization data returned by the LMS.
+    """
+    no_organization = {}
+
+    # Bypass caching if the programs cache ttl is not set.
```
clintonb: Remove this to eliminate complexity. Django already supports what you're doing: https://docs.djangoproject.com/en/1.9/topics/cache/#basic-usage.

credentials/apps/credentials/utils.py
```diff
@@ -77,6 +79,49 @@ def get_programs():
...
+    use_cache = False
+    if settings.ORGANIZATIONS_CACHE_TTL > 0:
+        use_cache = True
+
+    if use_cache:
+        cached = cache.get(ORGANIZATIONS_CACHE_KEY)
```
clintonb: Why are all organizations being cached as a single key? That seems unnecessarily complicated. Consider caching one value per organization.

```python
# TODO Ensure the cache keys are ASCII and less than 250 characters
cache_key = 'organization_{}'.format(hash(organization_key))

organization = cache.get(cache_key)

if not organization:
  # TODO Attempt to retrieve info from API

  # Do not cache empty values
  if organization:
    cache.set(cache_key, organization)

return organization
```

credentials/apps/credentials/utils.py
zubair-arbi: I have seen `ConnectionError`, `HttpClientError` and `AttributeError`. Handling this broad exception is consistent with the programs code.

zubair-arbi:
Closing this PR in favor of https://github.com/edx/credentials/pull/34

PR #29, opened 2016-01-09
changes in make file to remove test_root

ahsan-ul-haq:
@zubair-arbi @tasawernawaz @awais786


zubair-arbi:
thanks for fixing this :+1:

PR #30, opened 2016-01-11
Added JWT authentication, initial roles

ahsan-ul-haq:
@zubair-arbi @tasawernawaz @awais786 

ECOM-3413

### DIFF
--- a/credentials/apps/api/authentication.py
+++ b/credentials/apps/api/authentication.py
@@ -1,0 +1,65 @@
+"""
+Authentication logic for REST API.
+"""
+
+import logging
+
+from django.contrib.auth.models import Group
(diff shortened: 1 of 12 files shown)


ahsan-ul-haq:
@jimabramson please review

jimabramson:
@ahsan-ul-haq I recognize most of this code from the programs repo.  So I'm comfortable with the implementation generally since we've seen it work, but I'm not totally comfortable that we are duplicating all this.  Please create a story to move all the shared JWT auth code to its own repo - we will need to prioritize it immediately after the current milestone to ensure duplicated code doesn't begin to drift apart.

:+1: after follow up story created and squash/rebase.

PR #21, opened 2016-01-05
Awais786/ecom 3027 credential rendering

awais786:
https://openedx.atlassian.net/browse/ECOM-3284
default template rendering code.

https://github.com/edx/web-certificates/blob/master/templates/verified-base.html

PR #32, opened 2016-01-12
Awais786/ecom 3417 integrate ux pattern

awais786:
https://openedx.atlassian.net/browse/ECOM-3417

Integrating UX pattern library.  
Compiling scss.


awais786:
@marcotuts , @AlasdairSwan I have created separated PR for UX integration. Please review this.

credentials/templates/base.html
```diff
@@ -12,7 +12,8 @@
     <title>{% block title %}{% endblock title %}</title>
 
     {% compress css %}
-        <link rel="stylesheet" href="{% static 'sass/main.scss' %}" type="text/x-scss">
+        <link rel="stylesheet" href="{% static 'sass/edx-web-certificates-ltr.scss' %}" type="text/x-scss">
+        <link rel="stylesheet" href="{% static 'sass/edx-web-certificates-rtl.scss' %}" type="text/x-scss">
```
AlasdairSwan: You should not be loading both of these files, the second will supercede the first. There needs to be a mechanism which determines which file to load based on current language. @clintonb how is the Ecommerce IDA handling this?

awais786:
@AlasdairSwan I have added that condition while loading the credential.  
https://github.com/edx/credentials/pull/21/files#diff-04d0ea3f5a4c10801bb49e36591260e8R13

credentials/static/sass/_build.scss
```diff
@@ -0,0 +1,33 @@
...
+// ------------------------------
+// #CONFIG
+// ------------------------------
+@import 'config';
+
+// ------------------------------
+// #LIB
+// ------------------------------
+@import 'lib';
+@import '../../../node_modules/edx-pattern-library/pattern-library/sass/edx-pattern-library';
```
awais786: I have changed this path for credentials repo.

credentials/static/sass/_config.scss
```diff
@@ -0,0 +1,13 @@
+// ------------------------------
+// edX Web Certificates: Config
+
+// About: variable and configuration overrides
+
+// ------------------------------
+// #VARIABLES
+// ------------------------------
+ $font-path: '../bower_components/edx-pattern-library/pattern-library/fonts'; // LOCALDEV
```
awais786: Changed this path too.

AlasdairSwan:
@awais786 I am not seeing the font files in this PR. Are you using the fonts from the Pattern Library?

credentials/static/sass/mitpe/_utilities.scss
```diff
@@ -0,0 +1,43 @@
...
+// #MIXINS
+// #HELPERS
+
+// ------------------------------
+// #VARIABLES
+// ------------------------------
+ $font-path: '/node_modules/edx-pattern-library/pattern-library/fonts';
+// $font-path: '/media/certificate_template_assets/'; // CERTPROD
+
+$font-family-serif: Georgia, Times, 'Times New Roman', serif;
```
AlasdairSwan: @marcotuts is it a conscious decision to not use OpenSans?

awais786:
I have installed pattern Library. And it added the fonts inside bower folder. Do I need to commit that folder too ?

AlasdairSwan:
@awais786 in your output files you are attaching a cache-busting hash to the font files, e.g. ```url('/static/bower_components/edx-pattern-library/pattern-library/fonts/OpenSans/OpenSans-Light-webfont.woff2?e6b070bd17fd')```
I don't think this is necessary because the font files shouldn't really change. Can you confirm whether this hash updates after every compilation, or just when the font files themselves are updated?

credentials/templates/base.html
```diff
@@ -12,7 +12,8 @@
...
     {% endcompress %}
 
     {% compress css %}
```
AlasdairSwan: Note that you also need to load some JavaScript from the Pattern Library (e.g. https://github.com/edx/ux-pattern-library/blob/master/public/js/edx-icons.js) because of the accessibly icon approach being used.

awais786: @AlasdairSwan with each new compilation hash key updated.
http://127.0.0.1:8005/static/bower_components/edx-pattern-library/pattern-library/fonts/OpenSans/OpenSans-Light-webfont.woff2?408ef97722bc

AlasdairSwan: Can you please update to not do that because these assets will not change.

AlasdairSwan:
@awais786 I can confirm that the font files in the bower_components folder should be copied and added to this repo. The bower_components are used to generate the CSS to be used in production, so any other assets needed in production that are contained within that folder need to be hosted by the IDA.

awais786:
@AlasdairSwan I have added OpenSans and edx-icons files. Please let me know if i have to add any thing else too.

credentials/templates/base.html
```diff
@@ -12,7 +12,8 @@
     <title>{% block title %}{% endblock title %}</title>
 
     {% compress css %}
-        <link rel="stylesheet" href="{% static 'sass/main.scss' %}" type="text/x-scss">
+        <link rel="stylesheet" href="{% static 'sass/edx-web-certificates-ltr.scss' %}" type="text/x-scss">
```
clintonb: There may be other non-certificate pages served by this app. `base.html` should be reserved for content that will be loaded for *every* page. I would use `certificates/base.html` (which extends `base.html`) for certificate-specific pages.

AlasdairSwan:
@awais786 please move the Pattern Library fonts into a folder ```credentials/static/fonts``` instead of adding the bower_components folder.

credentials/templates/base.html
```diff
@@ -31,6 +32,7 @@
 {% compress js %}
     <script src="{% static 'bower_components/requirejs/require.js' %}"></script>
     <script src="{% static 'js/config.js' %}"></script>
+    <script src="{% static 'bower_components/edx-pattern-library/pattern-library/js/edx-icons.js' %}"></script>
```
AlasdairSwan: @awais786 this file needs to be added to ```credentials/static/js``` along with its dependencies (i.e. aFontGarde.js)

requirements/base.txt
```diff
@@ -13,3 +13,4 @@ edx-auth-backends==0.1.3
 pytz==2015.7
 Pillow==3.0.0
 git+https://github.com/edx/opaque-keys.git@9f07da1abf699d10bc3252d3081017e8aa37c302#egg=opaque-keys
+libsass==0.9.2
```
clintonb: This is already included with `django-libsass`. Also, the latest version is 0.10.0.

clintonb:
The purpose of using bower is to *not* include the various components in the source. Why are they being included?

AlasdairSwan:
@clintonb the full roll out approach for the pattern library hasn't been completely finalized yet. I think this is why JS and fonts inclusion has not been fully addressed. Currently these will need to be hosted by the IDA.

clintonb:
@AlasdairSwan what will it take to finalize? How long? edx/ecommerce does not require the files to be present in the repo: https://github.com/edx/ecommerce/blob/master/bower.json#L22.

AlasdairSwan:
@clintonb is there a way Ecommerce is handling there bower components in production?

AlasdairSwan:
@awais786 apologies, there was some confusion on my part re. the hosting of the bower_components folder in production. @clintonb is correct, you do *not* need to include the font files or JS files in this repo. Sorry.

credentials/static/sass/_config.scss
awais786: @AlasdairSwan sorry for asking this question again. But I have a confusions about these path changes. Reason is if we get the css from webcerts than it will have different path for the fonts etc. Then how IDA will serve the assets in prod ?

AlasdairSwan: @awais786 I think I need more background around how this CSS will integrate with the webcerts CSS. Will there be 2 different CSS files hosted in this IDA (one internal and one external)?

awais786: @AlasdairSwan initially we thought we only need ux-pattern-library for fonts. But we need more information from @marcotuts about how things will works.

AlasdairSwan: @awais786 so currently you are only loading the pattern library to use the edx-icons font? Or for OpenSans as well?

awais786: @AlasdairSwan check this certificate's.
https://github.com/edx/web-certificates/blob/master/templates/verified-base.html
It is using following css and all fonts are on s3.
https://edxuploads.s3.amazonaws.com/certificate_template_assets/1/edx-web-certificates-ltr.css

awais786:
@AlasdairSwan @clintonb Please review this PR. Main purpose is just compiling scss.

Makefile
```diff
@@ -60,7 +60,7 @@ quality:
 static:
 	$(NODE_BIN)/r.js -o build.js
 	python manage.py collectstatic --noinput -v0
-	python manage.py compress -v0
+	python manage.py compress -v0 --force
```
clintonb: This should not be necessary.

credentials/static/sass/_base.scss
```diff
@@ -0,0 +1,22 @@
...
+// #BASE
+// ------------------------------
+html, body {
+    height: 100%;
+    margin: 0;
+    padding: 0;
+    background: palette(grayscale-cool, xx-light);
+}
+
+// ------------------------------
```
clintonb: Remove.

credentials/static/sass/_base.scss
```diff
@@ -0,0 +1,22 @@
...
+// // edX Web Certificates: Base
+
+// About: Base resets and definitons (using shared resources from elements when appropriate).
+
+// ------------------------------
```
clintonb: Remove.

credentials/static/sass/_components-rendering.scss
```diff
@@ -0,0 +1,305 @@
+// ------------------------------
+// // edX Web Certificates: Certificate Rendering - Components
+
+// About: (specifically for certificate rendering) styling for specific UI components ranging from global to modular.
+
+// #RENDERING
```
clintonb: Remove.

credentials/static/sass/_config.scss
```diff
@@ -0,0 +1,13 @@
...
+// $font-path: '/media/certificate_template_assets/'; // SANDBOX
```
clintonb: Remove all of the commented-out code.

credentials/static/sass/_config.scss
clintonb: Remove the LOCALDEV comment.

credentials/static/sass/mitpe/_utilities.scss
```diff
@@ -0,0 +1,43 @@
...
+
+// #VARIABLES
+// #FUNCTIONS
+// #MIXINS
+// #HELPERS
+
+// ------------------------------
+// #VARIABLES
+// ------------------------------
+ $font-path: '/node_modules/edx-pattern-library/pattern-library/fonts';
```
clintonb: This path seems wrong. It would be best to setup a variable to the edx-pattern-library root path rather than duplicate this path everywhere.

credentials/static/sass/mitpe/_utilities.scss
```diff
@@ -0,0 +1,43 @@
...
+// $font-path: '/media/certificate_template_assets/'; // CERTPROD
```
clintonb: Remove dead code.

credentials/templates/base.html
clintonb: This needs to be addressed. Otto does not currently support RTL.

credentials/templates/base.html
clintonb: This needs to be addressed.

credentials/templates/base.html
```diff
@@ -34,6 +35,10 @@
 
     {# Note: django-compressor does not recognize the data-main attribute. Load the main script separately. #}
     <script src="{% static 'js/common.js' %}"></script>
+
+    <script src="{% static 'bower_components/edx-pattern-library/pattern-library/js/edx-icons.js' %}"></script>
```
clintonb: If these are needed for every page, they should be loaded via `require` in common.js.

AlasdairSwan: ```edx-icons.js``` is needed on any page that has an icon on it.

credentials/static/sass/_components-rendering.scss
```diff
@@ -0,0 +1,305 @@
...
+
+// ------------------------------
+// #RENDERING
+// ------------------------------
+.wrapper-accomplishment-title {
+
+  .accomplishment-title {
+    @extend %hd-1;
+    font-weight: 300;
```
AlasdairSwan: You should be able to use the font-weight function (https://github.com/edx/ux-pattern-library/blob/2ad7677936b469dd8a445070d47a619075de9d49/pattern-library/sass/global/_functions.scss#L38) and pass it light (https://github.com/edx/ux-pattern-library/blob/288d3155c80819d2c7094f6d27175756ff5477bd/pattern-library/sass/global/_settings.scss#L273) - i.e. ```font-weight(light)```

credentials/static/sass/_layouts-rendering.scss
```diff
@@ -0,0 +1,170 @@
...
+  @extend %depth-2;
+  margin: 2%;
+  border: 1px solid palette(grayscale-cool, light);
+  padding: 7% 5% 5% 5%;
+  background-color: palette(grayscale, white);
+  text-align: center;
+
+  @include susy-breakpoint($bp-screen-md, $susy) {
+    @include text-align(left);
+    padding: 5% 5% 3% 5%;
```
AlasdairSwan: Nitpick: this is the same as the slightly shorter version ```padding: 5% 5% 3%;```

credentials/static/sass/_layouts-rendering.scss
```diff
@@ -0,0 +1,170 @@
...
+  text-align: center;
+
+  @include susy-breakpoint($bp-screen-md, $susy) {
+    display: inline-block;
+    vertical-align: middle;
+    width: span(6 of 12);
+    margin-top: 0;
+    margin-bottom: spacing-vertical(base);
+    @include margin-left(gutter(inside));
+    @include text-align(right);
```
AlasdairSwan: Please move these includes to the top of the selector (line 50). Generally speaking includes should be at the top, but I'm okay with the susy-breakpoints being at the bottom because that makes more sense for readability.

credentials/static/sass/_layouts-rendering.scss
```diff
@@ -0,0 +1,170 @@
...
+  padding-bottom: spacing-vertical(small);
+
+  @include susy-breakpoint($bp-screen-lg, $susy) {
+    display: inline-block;
+    vertical-align: top;
+    width: span(4 of 12);
+    @include margin-left(gutter(inside));
+    border-bottom: 0;
+    padding-bottom: 0;
+    @include text-align(right);
```
AlasdairSwan: Please move this include to the top of the selector (line 94)

credentials/static/sass/_layouts-rendering.scss
```diff
@@ -0,0 +1,170 @@
...
+
+.accomplishment-signatories {
+  border-bottom: 1px solid palette(grayscale-cool, light);
+  padding-bottom: spacing-vertical(small);
+
+  @include susy-breakpoint($bp-screen-lg, $susy) {
+    display: inline-block;
+    vertical-align: top;
+    width: span(4 of 12);
+    @include margin-left(gutter(inside));
```
AlasdairSwan: Please move this include to the top of the selector (line 94)

credentials/static/sass/_layouts-rendering.scss
```diff
@@ -0,0 +1,170 @@
...
+.accomplishment-stamps {
+  @extend %list-unstyled;
+
+  .accomplishment-stamp-platform {
+    margin-bottom: spacing-vertical(small);
+
+    .logo-img {
+      width: 125px;
+      margin-top: 2px;
+      max-width: 100%;
```
AlasdairSwan: If you are explicitly setting a width you shouldn't need a max-width too.

credentials/static/sass/_layouts-rendering.scss
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    @include susy-breakpoint($bp-screen-lg, $susy) {
+      display: inline-block;
+      vertical-align: middle;
+      width: span(2 of 12);
+      @include margin-right(gutter(inside));
```
AlasdairSwan: Please move this include to the top of the selector (line 135)

credentials/static/sass/_layouts-rendering.scss
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  .accomplishment-stamp-date {
+    margin-bottom: spacing-vertical(x-small);
+
+    @include susy-breakpoint($bp-screen-md, $susy) {
+      display: inline-block;
+      vertical-align: middle;
+      width: span(4 of 12);
+      @include margin-right(gutter(inside));
```
AlasdairSwan: Please move this include to the top of the selector (line 147)

credentials/static/sass/_print-rendering.scss
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    .accomplishment-stamps {
+        margin-top: 0;
+
+        .accomplishment-stamp-platform {
+            display: inline-block;
+            vertical-align: top;
+            width: span(1 of 12);
+            @include margin-right(gutter(inside));
```
AlasdairSwan: Please move this include to the top of the selector (line 86)

credentials/static/sass/_print-rendering.scss
```diff
@@ -0,0 +1,235 @@
...
+            margin-bottom: 0;
+            padding-top: 0;
+        }
+
+        .accomplishment-stamp-date {
+            width: span(3 of 12);
+            display: inline-block;
+            vertical-align: top;
+            @include margin-right(gutter(inside));
```
AlasdairSwan: Please move this include to the top of the selector (line 95)

credentials/static/sass/_print-rendering.scss
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    // certificate - verified
+    .certificate-verified {
+
+        .accomplishment-rendering {
+            border: 1px solid palette(grayscale-cool, light);
+            background-image: url("/static/images/verified-seal-grayscale-darker.png"); // LOCALDEV
+//            background-image: url("/media/certificate_template_assets/18/verified-seal-grayscale-darker.png"); // SANDBOX
+//            background-image: url("https://edxuploads.s3.amazonaws.com/certificate_template_assets/2/verified-seal-grayscale-darker.png"); // PROD
```
AlasdairSwan: There needs to be a better way of doing this than commenting/uncommenting per environment. cc @clintonb

credentials/static/sass/_print.scss
```diff
@@ -0,0 +1,138 @@
...
+// #LAYOUT
+// #ELEMENTS
+
+// ------------------------------
+// #UTILITIES
+// ------------------------------
+
+// page set up - only available to modern browsers now
+@page {
+    size : landscape;
```
AlasdairSwan: Nitpick: there should not be a space between ```size``` and ```:```

credentials/static/sass/_views.scss
```diff
@@ -0,0 +1,103 @@
...
+// ------------------------------
+// #VERIFIED CERT
+// ------------------------------
+.certificate-verified {
+
+  .accomplishment-rendering {
+    background-color: palette(grayscale, white-t);
+    background-image: url("/static/images/verified-seal-grayscale.png"); // LOCALDEV
+//    background-image: url("/media/certificate_template_assets/11/verified-seal-grayscale.png"); // SANDBOX
+//    background-image: url("https://edxuploads.s3.amazonaws.com/certificate_template_assets/5/verified-seal-grayscale.png"); // PROD
```
AlasdairSwan: There should be a better way of doing this.

credentials/static/sass/_views.scss
```diff
@@ -0,0 +1,103 @@
...
+// ------------------------------
+// #PROFESSIONAL EDUCATION CERT
+// ------------------------------
+.certificate-professional {
+
+  .accomplishment-rendering {
+    background-color: palette(grayscale, white-t);
+    background-image: url("/static/images/professionaled-seal-grayscale.png"); // LOCALDEV
+//    background-image: url("/media/certificate_template_assets/19/professionaled-seal-grayscale.png"); // SANDBOX
+//    background-image: url("https://edxuploads.s3.amazonaws.com/certificate_template_assets/32/professionaled-seal-grayscale.png"); // PROD
```
AlasdairSwan: There should be a better way of doing this.

credentials/static/sass/_views.scss
```diff
@@ -0,0 +1,103 @@
...
+// ------------------------------
+// #XSERIES PROGRAM CERT
+// ------------------------------
+.certificate-xseries {
+
+  .accomplishment-rendering {
+    background-color: palette(grayscale, white-t);
+    background-image: url("/static/images/xseries-seal-grayscale.png"); // LOCALDEV
+//    background-image: url("/media/certificate_template_assets/11/xseries-seal-grayscale.png"); // SANDBOX
+//    background-image: url("https://edxuploads.s3.amazonaws.com/certificate_template_assets/5/xseries-seal-grayscale.png"); // PROD
```
AlasdairSwan: There should be a better way of doing this.

credentials/static/sass/mitpe/_layouts-rendering.scss
```diff
@@ -0,0 +1,123 @@
...
+// ------------------------------
+// #ACCOMPLISHMENT
+// ------------------------------
+.accomplishment-rendering {
+  @extend %depth-2;
+  margin: 2%;
+  border: 1px solid palette(grayscale-cool, light);
+  padding: 5%;
+  background-color: palette(grayscale, white);
+  @include text-align(center);
```
AlasdairSwan: Please move this ```@include``` to the top of the selector.

credentials/static/sass/mitpe/_layouts-rendering.scss
```diff
@@ -0,0 +1,123 @@
...
+
+  @include susy-breakpoint($bp-screen-md, $susy) {
+    padding: 2% 5% 2% 5%;
```
AlasdairSwan: This is the same as ```padding: 2% 5%```

credentials/templates/base.html
AlasdairSwan: @awais786 as per @clintonb's comment I think you should move ```edx-icons.js``` into requireJS and let it be loaded that way (in turn ```afontgarde.js``` should be loaded via requireJS from ```edx-icons.js```).

credentials/static/sass/_views.scss
clintonb: Agreed. All of this commented-out stuff needs to go away. It is left over from the old repo. We should have variables to respective paths (e.g. `image_path`, `edx_pattern_library_path`) and set those accordingly for local usage. If we get local usage correct, Django Compressor can behave appropriately for sandboxes, stage, and production.

marcotuts:
@AlasdairSwan I should have checked this PR prior to merging the other styling work I made. You made a bunch of suggestions here that I should have included there. Would it make sense to capture these all in a separate story and then have me follow up with that PR tomorrow or early next week? Apologies for the churn in approach and review on this.

AlasdairSwan:
@marcotuts I'm fine having this PR be about bringing the UXPL into the repo and a follow up PR to clean up the Sass.

awais786:
screen shot 2016-01-15 at 6 56 44 pm
I am loading fonts files from base.html and getting this error.

awais786:
@AlasdairSwan I have implemented the UX pattern library using bower components. Please review this PR.

awais786:
@marcotuts Please create a PR to fix all comments related with sass.

credentials/static/sass/_base.scss
marcotuts: What is the suggestion here?

marcotuts:
@awais786, @AlasdairSwan I'll have a PR up later this afternoon with the suggested changes to the web certs styling. Thanks!

credentials/static/sass/_base.scss
clintonb: Do we need empty sections?

awais786:
@clintonb please review this PR.

clintonb:
Why are all of the image assets being added to the repository despite not being used? We should not have unused static assets.

clintonb:
@awais786 please respond to the first round of feedback before asking for more. There remains commented-out SASS that needs to go away. This was pointed out in the initial round of review.

credentials/static/sass/_build.scss
```diff
@@ -0,0 +1,33 @@
...
+// ------------------------------
+// #CONFIG
+// ------------------------------
+@import 'config';
+
+// ------------------------------
+// #LIB
+// ------------------------------
+@import 'lib';
+@import '../bower_components/edx-pattern-library/pattern-library/sass/edx-pattern-library';
```
clintonb: Create a variable for the pattern library path. It is used in at least four locations.

awais786:
@clintonb  Please check this comment.
https://github.com/edx/credentials/pull/32#issuecomment-171783457

@marcotuts will work on a followup story for sass related stuff.

clintonb:
@awais786 if your goal is to simply integrate the UX pattern library, you've done way too much work that is outside of the scope of the integration. The integration of the pattern library should not entail much more beyond updating `bower.json` and ensuring the correct SCSS and JS files are loaded.

credentials/static/js/apps/credentials_app.js
```diff
@@ -0,0 +1,9 @@
+(function() {
+    'use strict';
+
+    // Load files for credentials rendering
```
clintonb: This comment is incorrect, and not necessary.

credentials/static/js/apps/credentials_app.js
```diff
@@ -0,0 +1,9 @@
+(function() {
```
clintonb: Why is this file in an apps directory. There isn't really an app here. You'll be rendering a single page.

credentials/static/js/common.js
```diff
@@ -1,4 +1,4 @@
-require([
+define([
```
clintonb: Why was this change made? It's not necessary, and will probably break.

credentials/static/sass/main.scss
```diff
@@ -1 +1,2 @@
 // Styles for the credentials service
+
```
clintonb: Shouldn't you be including the pattern library here?

zubair-arbi:
There is lot of confusion about what this PR is doing so I am going to close it in favor of https://github.com/edx/credentials/pull/39

PR #33, opened 2016-01-12
provide type for code-block examples in docs

zubair-arbi:
@ahsan-ul-haq @tasawernawaz @awais786 

* For now the docs are not displaying code example and I believe that we need to provide the type to `code-block` parameter.
* Update input example for create user credential endpoint in docs.


awais786:
:+1:

tasawernawaz:
:+1:

PR #34, opened 2016-01-13
get organization and user data from lms

zubair-arbi:
ECOM-3282

* Create settings variables required for communication with the `organizations` API in `LMS`
* Add `get_organization` method to get an organization from edx rest api client with the provided `organization_key`
* Add tests for data retrieval from `organizations` api
* Create settings variables required for communication with the `user` API in `LMS`
* Add `get_user` method to get a user data from edx rest api client with the provided `username`
* Add tests for data retrieval from`users` api

For LMS checkout this PR to add JWT auth: https://github.com/edx/edx-platform/pull/11211

### DIFF
--- a/credentials/apps/credentials/apps.py
+++ b/credentials/apps/credentials/apps.py
@@ -1,0 +1,39 @@
+"""
+Basic configuration for the credentials app.
+"""
+import logging
+
+from django.apps import AppConfig
+from django.conf import settings
+
+
+log = logging.getLogger(__name__)
+
+
+class CredentialsConfig(AppConfig):
+    """
+    Credentials app configuration.
+    """
+    name = 'credentials.apps.credentials'
+    verbose_name = 'Credentials'
+
+    def ready(self):
+
+        required_config_settings = (
+            'CREDENTIALS_SERVICE_USER',
+            'PROGRAMS_API_URL', 'PROGRAMS_JWT_AUDIENCE', 'PROGRAMS_JWT_SECRET_KEY',
+            'ORGANIZATIONS_API_URL', 'ORGANIZATIONS_AUDIENCE', 'ORGANIZATIONS_SECRET_KEY',
+            'USER_API_URL', 'USER_JWT_AUDIENCE', 'USER_JWT_SECRET_KEY',
+        )
+
+        missing_config_settings = []
+        for setting in required_config_settings:
+            if not hasattr(settings, setting) or not getattr(settings, setting):
+                missing_config_settings.append(setting)
+
+        if missing_config_settings:
+            msg = 'The settings {} must be set in order to start the application!'.format(
+                ', '.join(missing_config_settings)
+            )
+            log.critical(msg)
+            raise AttributeError(msg)
--- a/credentials/apps/credentials/tests/mixins.py
+++ b/credentials/apps/credentials/tests/mixins.py
@@ -3,9 +3,11 @@ Mixins for use during testing.
 """
 import json
 
+from django.conf import settings
 import httpretty
 
-from django.conf import settings
+
+CONTENT_TYPE = 'application/json'
 
 
 class ProgramsDataMixin(object):
@@ -27,153 +29,168 @@ class ProgramsDataMixin(object):
     ]
 
     PROGRAMS_API_RESPONSE = {
-        'results': [
+        'id': 1,
+        'name': PROGRAM_NAMES[0],
+        'subtitle': 'A program used for testing purposes',
+        'category': 'xseries',
+        'status': 'unpublished',
+        'marketing_slug': '',
+        'organizations': [
             {
-                'id': 1,
-                'name': PROGRAM_NAMES[0],
-                'subtitle': 'A program used for testing purposes',
-                'category': 'xseries',
-                'status': 'unpublished',
-                'marketing_slug': '',
-                'organizations': [
+                'display_name': 'Test Organization A',
+                'key': 'organization-a'
+            }
+        ],
+        'course_codes': [
+            {
+                'display_name': 'Test Course A',
+                'key': 'course-a',
+                'organization': {
+                    'display_name': 'Test Organization A',
+                    'key': 'organization-a'
+                },
+                'run_modes': [
                     {
-                        'display_name': 'Test Organization A',
-                        'key': 'organization-a'
-                    }
-                ],
-                'course_codes': [
-                    {
-                        'display_name': 'Test Course A',
-                        'key': 'course-a',
-                        'organization': {
-                            'display_name': 'Test Organization A',
-                            'key': 'organization-a'
-                        },
-                        'run_modes': [
-                            {
-                                'course_key': COURSE_KEYS[0],
-                                'mode_slug': 'verified',
-                                'sku': '',
-                                'start_date': '2015-11-05T07:39:02.791741Z',
-                                'run_key': 'fall'
-                            },
-                            {
-                                'course_key': COURSE_KEYS[1],
-                                'mode_slug': 'verified',
-                                'sku': '',
-                                'start_date': '2015-11-05T07:39:02.791741Z',
-                                'run_key': 'winter'
-                            }
-                        ]
+                        'course_key': COURSE_KEYS[0],
+                        'mode_slug': 'verified',
+                        'sku': '',
+                        'start_date': '2015-11-05T07:39:02.791741Z',
+                        'run_key': 'fall'
                     },
                     {
-                        'display_name': 'Test Course B',
-                        'key': 'course-b',
-                        'organization': {
-                            'display_name': 'Test Organization A',
-                            'key': 'organization-a'
-                        },
-                        'run_modes': [
-                            {
-                                'course_key': COURSE_KEYS[2],
-                                'mode_slug': 'verified',
-                                'sku': '',
-                                'start_date': '2015-11-05T07:39:02.791741Z',
-                                'run_key': 'fall'
-                            },
-                            {
-                                'course_key': COURSE_KEYS[3],
-                                'mode_slug': 'verified',
-                                'sku': '',
-                                'start_date': '2015-11-05T07:39:02.791741Z',
-                                'run_key': 'winter'
-                            }
-                        ]
+                        'course_key': COURSE_KEYS[1],
+                        'mode_slug': 'verified',
+                        'sku': '',
+                        'start_date': '2015-11-05T07:39:02.791741Z',
+                        'run_key': 'winter'
                     }
-                ],
-                'created': '2015-10-26T17:52:32.861000Z',
-                'modified': '2015-11-18T22:21:30.826365Z'
+                ]
             },
             {
-                'id': 2,
-                'name': PROGRAM_NAMES[1],
-                'subtitle': 'Another program used for testing purposes',
-                'category': 'xseries',
-                'status': 'unpublished',
-                'marketing_slug': '',
-                'organizations': [
+                'display_name': 'Test Course B',
+                'key': 'course-b',
+                'organization': {
+                    'display_name': 'Test Organization A',
+                    'key': 'organization-a'
+                },
+                'run_modes': [
                     {
-                        'display_name': 'Test Organization B',
-                        'key': 'organization-b'
-                    }
-                ],
-                'course_codes': [
-                    {
-                        'display_name': 'Test Course C',
-                        'key': 'course-c',
-                        'organization': {
-                            'display_name': 'Test Organization B',
-                            'key': 'organization-b'
-                        },
-                        'run_modes': [
-                            {
-                                'course_key': COURSE_KEYS[4],
-                                'mode_slug': 'verified',
-                                'sku': '',
-                                'start_date': '2015-11-05T07:39:02.791741Z',
-                                'run_key': 'fall'
-                            },
-                            {
-                                'course_key': COURSE_KEYS[5],
-                                'mode_slug': 'verified',
-                                'sku': '',
-                                'start_date': '2015-11-05T07:39:02.791741Z',
-                                'run_key': 'winter'
-                            }
-                        ]
+                        'course_key': COURSE_KEYS[2],
+                        'mode_slug': 'verified',
+                        'sku': '',
+                        'start_date': '2015-11-05T07:39:02.791741Z',
+                        'run_key': 'fall'
                     },
                     {
-                        'display_name': 'Test Course D',
-                        'key': 'course-d',
-                        'organization': {
-                            'display_name': 'Test Organization B',
-                            'key': 'organization-b'
-                        },
-                        'run_modes': [
-                            {
-                                'course_key': COURSE_KEYS[6],
-                                'mode_slug': 'verified',
-                                'sku': '',
-                                'start_date': '2015-11-05T07:39:02.791741Z',
-                                'run_key': 'fall'
-                            },
-                            {
-                                'course_key': COURSE_KEYS[7],
-                                'mode_slug': 'verified',
-                                'sku': '',
-                                'start_date': '2015-11-05T07:39:02.791741Z',
-                                'run_key': 'winter'
-                            }
-                        ]
+                        'course_key': COURSE_KEYS[3],
+                        'mode_slug': 'verified',
+                        'sku': '',
+                        'start_date': '2015-11-05T07:39:02.791741Z',
+                        'run_key': 'winter'
                     }
-                ],
-                'created': '2015-10-26T19:59:03.064000Z',
-                'modified': '2015-10-26T19:59:18.536000Z'
+                ]
             }
-        ]
+        ],
+        'created': '2015-10-26T17:52:32.861000Z',
+        'modified': '2015-11-18T22:21:30.826365Z'
     }
 
-    def mock_programs_api(self, data=None, status_code=200):
+    def mock_programs_api(self, program_id):
+        """ Utility for mocking out Programs API URLs with status code 200."""
+        data = self.PROGRAMS_API_RESPONSE
+        self._mock_programs_api(program_id, data)
+
+    def mock_programs_api_404(self, program_id):
+        """ Utility for mocking out Programs API URLs with status code 404."""
+        data = None
+        self._mock_programs_api(program_id, data, 404)
+
+    def mock_programs_api_500(self, program_id):
+        """ Utility for mocking out Programs API URLs with status code 500."""
+        data = None
+        self._mock_programs_api(program_id, data, 500)
+
+    def _mock_programs_api(self, program_id, data, status_code=200):
         """ Utility for mocking out Programs API URLs."""
         self.assertTrue(httpretty.is_enabled(), msg='httpretty must be enabled to mock Programs API calls.')
 
         programs_api_url = settings.PROGRAMS_API_URL
-        url = programs_api_url.strip('/') + '/programs/'
-
-        if data is None:
-            data = self.PROGRAMS_API_RESPONSE
-
+        url = '{root}/programs/{program_id}/'.format(root=programs_api_url.strip('/'), program_id=program_id)
         body = json.dumps(data)
 
         httpretty.reset()
-        httpretty.register_uri(httpretty.GET, url, body=body, content_type='application/json', status=status_code)
+        httpretty.register_uri(httpretty.GET, url, body=body, content_type=CONTENT_TYPE, status=status_code)
+
+
+class OrganizationsDataMixin(object):
+    """ Mixin mocking Organizations API URLs and providing fake data for testing."""
+    ORGANIZATIONS_API_RESPONSE = {
+        'name': 'Test Organization',
+        'short_name': 'test-org',
+        'description': 'Organization for testing.',
+        'logo': 'http://testserver/media/organization_logos/test_org_logo.png',
+    }
+
+    def mock_organizations_api(self, organization_key):
+        """ Utility for mocking out Organizations API URLs with status code 200."""
+        data = self.ORGANIZATIONS_API_RESPONSE
+        self._mock_organizations_api(organization_key, data)
+
+    def mock_organizations_api_404(self, organization_key):
+        """ Utility for mocking out Organizations API URLs with status code 404."""
+        data = None
+        self._mock_organizations_api(organization_key, data, 404)
+
+    def mock_organizations_api_500(self, organization_key):
+        """ Utility for mocking out Organizations API URLs with status code 500."""
+        data = None
+        self._mock_organizations_api(organization_key, data, 500)
+
+    def _mock_organizations_api(self, organization_key, data, status_code=200):
+        """ Utility for mocking out Organizations API URLs."""
+        self.assertTrue(httpretty.is_enabled(), msg='httpretty must be enabled to mock Organizations API calls.')
+
+        organizations_api_url = settings.ORGANIZATIONS_API_URL
+        url = '{root}/organizations/{organization_key}/'.format(
+            root=organizations_api_url.strip('/'),
+            organization_key=organization_key
+        )
+        body = json.dumps(data)
+
+        httpretty.reset()
+        httpretty.register_uri(httpretty.GET, url, body=body, content_type=CONTENT_TYPE, status=status_code)
+
+
+class UserDataMixin(object):
+    """ Mixin mocking User API URLs and providing fake data for testing."""
+    USER_API_RESPONSE = {
+        "username": "test-user",
+        "name": "Test User",
+        "email": "test@example.org",
+    }
+
+    def mock_user_api(self, username):
+        """ Utility for mocking out User API URLs with status code 200."""
+        data = self.USER_API_RESPONSE
+        self._mock_user_api(username, data)
+
+    def mock_user_api_404(self, username):
+        """ Utility for mocking out User API URLs with status code 404."""
+        data = None
+        self._mock_user_api(username, data, 404)
+
+    def mock_user_api_500(self, username):
+        """ Utility for mocking out User API URLs with status code 500."""
+        data = None
+        self._mock_user_api(username, data, 500)
+
+    def _mock_user_api(self, username, data, status_code=200):
+        """ Helper method for mocking out User API URLs."""
+        self.assertTrue(httpretty.is_enabled(), msg='httpretty must be enabled to mock User API calls.')
+
+        user_api_url = settings.USER_API_URL
+        url = '{root}/accounts/{username}'.format(root=user_api_url.strip('/'), username=username)
+        body = json.dumps(data)
+
+        httpretty.reset()
+        httpretty.register_uri(httpretty.GET, url, body=body, content_type=CONTENT_TYPE, status=status_code)
--- a/credentials/apps/credentials/tests/test_apps.py
+++ b/credentials/apps/credentials/tests/test_apps.py
@@ -1,0 +1,47 @@
+"""
+Tests for credentials configuration.
+"""
+import ddt
+from django.apps import AppConfig
+from django.conf import settings
+from django.test import TestCase, override_settings
+from testfixtures import LogCapture
+
+
+LOGGER_NAME = 'credentials.apps.credentials.apps'
+
+
+@ddt.ddt
+class CredentialsConfigTests(TestCase):
+    """ Tests covering the configuration for programs api, organizations api
+    and user api.
+    """
+    def setUp(self):
+        super(CredentialsConfigTests, self).setUp()
+        self.app_config = AppConfig.create('credentials.apps.credentials')
+
+    @ddt.data(
+        'CREDENTIALS_SERVICE_USER',
+        'PROGRAMS_API_URL',
+        'PROGRAMS_JWT_AUDIENCE',
+        'PROGRAMS_JWT_SECRET_KEY',
+        'ORGANIZATIONS_API_URL',
+        'ORGANIZATIONS_AUDIENCE',
+        'ORGANIZATIONS_SECRET_KEY',
+        'USER_API_URL',
+        'USER_JWT_AUDIENCE',
+        'USER_JWT_SECRET_KEY',
+    )
+    @override_settings()
+    def test_credentials_config_settings_failure(self, setting_attribute):
+        """ Verify that the exception is raised when there are missing settings
+        for the credentials app.
+        """
+        delattr(settings, setting_attribute)
+        expected_error_msg = "The settings {} must be set in order to start the application!".format(setting_attribute)
+
+        with LogCapture(LOGGER_NAME) as log:
+            with self.assertRaises(AttributeError):
+                self.app_config.ready()
+
+            log.check((LOGGER_NAME, 'CRITICAL', expected_error_msg))
--- a/credentials/apps/credentials/tests/test_utils.py
+++ b/credentials/apps/credentials/tests/test_utils.py
@@ -1,22 +1,18 @@
 """
 Tests for Issuer class.
 """
-import copy
-
 import ddt
 import httpretty
-from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.core.cache import cache
-from django.test import TestCase
+from django.test import TestCase, override_settings
 from mock import patch
-from testfixtures import LogCapture
+from slumber import exceptions
 
-from credentials.apps.credentials.tests.mixins import ProgramsDataMixin
-from credentials.apps.credentials.utils import get_programs, validate_duplicate_attributes
+from credentials.apps.credentials.tests.mixins import OrganizationsDataMixin, ProgramsDataMixin, UserDataMixin
+from credentials.apps.credentials.utils import get_organization, get_program, get_user, validate_duplicate_attributes
 
 
-LOGGER_NAME = 'credentials.apps.credentials.utils'
 User = get_user_model()
 
 
@@ -25,7 +21,7 @@ class ValidateDuplicateAttributesTests(TestCase):
     """ Tests for Validate the attributes method """
 
     def test_with_non_duplicate_attributes(self):
-        """ Verify that the method will return True if no duplicated attributes found."""
+        """ Verify that the function will return True if no duplicated attributes found."""
         attributes = [
             {"name": "whitelist_reason", "value": "Reason for whitelisting."},
             {"name": "grade", "value": "0.85"}
@@ -33,7 +29,7 @@ class ValidateDuplicateAttributesTests(TestCase):
         self.assertTrue(validate_duplicate_attributes(attributes))
 
     def test_with_duplicate_attributes(self):
-        """ Verify that the method will return False if duplicated attributes found."""
+        """ Verify that the function will return False if duplicated attributes found."""
 
         attributes = [
             {"name": "whitelist_reason", "value": "Reason for whitelisting."},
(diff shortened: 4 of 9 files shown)


zubair-arbi:
@clintonb @jimabramson please review

credentials/settings/base.py
```diff
@@ -244,12 +245,21 @@
 PROGRAMS_JWT_AUDIENCE = 'programs-key'
 PROGRAMS_JWT_SECRET_KEY = 'programs-secret'
 
-# Absolute URL used to construct URLs pointing to the LMS
-LMS_URL_ROOT = 'http://127.0.0.1:8000'
```
tasawernawaz: Why we are removing this. You can use `settings.LMS_URL_ROOT`and `urljoin` for getting urls, instead of hard coding everywhere.

credentials/apps/credentials/tests/test_utils.py
```diff
@@ -137,3 +138,227 @@ def test_get_programs_with_no_data(self):
...
+        Verify the behavior when data can't be retrieved from User API.
+        """
+        self.mock_user_api(username=self.username, status_code=500)
+
+        actual_user_api_response = get_user(self.username)
+        self.assertEqual(actual_user_api_response, {})
+
+    @httpretty.activate
+    def test_get_user_with_no_data(self):
+        """ Verify the behavior when no user data is found from the
```
tasawernawaz: Please be consistent with format of docstrings.

credentials/apps/credentials/tests/mixins.py
```diff
@@ -177,3 +177,79 @@ def mock_programs_api(self, data=None, status_code=200):
...
+        "email": "test@example.org",
+        "date_joined": "2015-11-17T03:16:01Z"
+    }
+
+    def mock_user_api(self, username, status_code=200):
+        """ Utility for mocking out User API URLs."""
+        self.assertTrue(httpretty.is_enabled(), msg='httpretty must be enabled to mock User API calls.')
+
+        user_api_url = settings.USER_API_URL
+        url = user_api_url.strip('/') + '/accounts/{}'.format(username)
```
tasawernawaz: Instead of different techniques of joining, you can use `+` for appending `username` at the end for better readability.

credentials/settings/base.py
zubair-arbi: We are now using absolute urls of API's as they could be served from a separate IDA so we don't need this setting.
https://github.com/edx/credentials/pull/26#discussion_r49330623

credentials/apps/credentials/tests/test_utils.py
zubair-arbi: Done

credentials/apps/credentials/tests/test_utils.py
```diff
@@ -137,3 +138,226 @@ def test_get_programs_with_no_data(self):
...
+    def test_get_organizations_client_settings_failure(self, setting_attribute):
+        """
+        Verify the behavior when API client fails to initialize due to
+        missing settings.
+        """
+        self.mock_organizations_api(self.org_key)
+        patched_settings = copy.deepcopy(settings)
+        delattr(patched_settings, setting_attribute)
+        expected_error_msg = "Failed to get settings for communication with the Organizations API. Please make sure " \
+                             "that the settings for 'ORGANIZATIONS_API_URL', 'CREDENTIALS_SERVICE_USER', " \
```
jimabramson: string literals can be broken across multiple lines without backslashes

credentials/apps/credentials/tests/test_utils.py
```diff
@@ -137,3 +138,226 @@ def test_get_programs_with_no_data(self):
...
+    @httpretty.activate
+    def test_get_organization_with_no_data(self):
+        """ Verify the behavior when no organizations data is found from the
+        Organizations API.
+        """
+        org_key = 'invlaid-org-key'
+        self.mock_organizations_api(org_key)
+
+        actual_organizations_api_response = get_organization(org_key)
+        self.assertEqual(actual_organizations_api_response, {})
```
jimabramson: you don't expect a 404 in this case?

credentials/settings/production.py
```diff
@@ -15,6 +15,9 @@
 # AMAZON S3 STORAGE CONFIGURATION
 # See: https://django-storages.readthedocs.org/en/latest/backends/amazon-S3.html#settings
 
+SOCIAL_AUTH_EDX_OIDC_KEY = 'credentials-key'
```
clintonb: Remove.

jimabramson:
LGTM apart from nitpicks; expecting to see some configuration PRs (reviewed by ops) for all these auth configs at some point...

credentials/apps/credentials/utils.py
```diff
@@ -68,6 +71,88 @@ def get_programs():
...
+    if organizations_api is None:
+        return no_organization
+
+    try:
+        organization = organizations_api.organization(organization_key).get()
+    except Exception:  # pylint: disable=broad-except
+        log.exception('Failed to retrieve organization with key %s from the Organization API.', organization_key)
+        return no_organization
+
+    cache.set(cache_key, organization, settings.ORGANIZATIONS_CACHE_TTL)
```
clintonb: The value should only be cached if the API returned data. Move this inside the exception handler.

credentials/apps/credentials/utils.py
```diff
@@ -87,7 +172,46 @@ def get_program_api_client():
...
+    """
+    Return api client to communicate with the organizations api by using the
+    credentials service user in the LMS.
+    """
+    try:
+        organizations_api_url = settings.ORGANIZATIONS_API_URL
+        service_username = settings.CREDENTIALS_SERVICE_USER
+        jwt_audience = settings.ORGANIZATIONS_AUDIENCE
+        jwt_secret_key = settings.ORGANIZATIONS_SECRET_KEY
+    except AttributeError:
```
clintonb: Remove the exception handler. These are pretty significant errors that we want to know about ASAP. Ideally, the presence of these values should be checked before the application starts. The `ready()` method can be used for this purpose: https://github.com/edx/course-discovery/blob/4cdecab5b4ccbdfcbcd7eaa8af3f26a40e1e6d69/course_discovery/apps/courses/apps.py.

credentials/apps/credentials/utils.py
zubair-arbi: This is the case here. If there is no organization this code won't run.

credentials/apps/credentials/utils.py
```diff
@@ -87,7 +172,46 @@ def get_program_api_client():
...
+    """
+    Return api client to communicate with the user api by using the credentials
+    service user in the LMS.
+    """
+    try:
+        user_api_url = settings.USER_API_URL
+        service_username = settings.CREDENTIALS_SERVICE_USER
+        jwt_audience = settings.USER_JWT_AUDIENCE
+        jwt_secret_key = settings.USER_JWT_SECRET_KEY
+    except AttributeError:
```
clintonb: Same as above. If you combine these exception handlers with the fact that you are currently caching an empty dict when any of this API call fails, you will only log the exception once. That's not enough to be noticed by New Relic or Splunk. Catching this exception is actually hiding it operationally, which is not good.

credentials/apps/credentials/utils.py
clintonb: Move it all inside the exception handler to make it explicit that the cache is only set when the call succeeds. As we've seen, I just glossed over the fact that you're returning from within your exception handler. You don't want to run the risk of that happening and someone doing something stranger than making a PR comment.

credentials/apps/credentials/tests/test_utils.py
zubair-arbi: Nope. If there is no organization then the `get_organization` will return empty dict

zubair-arbi:
@jimabramson I have added another PR to add basic settings to `credentials.yml`: https://github.com/edx/configuration/pull/2704

credentials/apps/credentials/utils.py
zubair-arbi: This is not the case. If there is any attribute missing from settings than this exception will be logged each time when a user is not found in cache and need to get user api client.

credentials/apps/credentials/utils.py
zubair-arbi: I believe we need this to exception handling to make sure the exception is logged with proper message.

zubair-arbi:
@clintonb @jimabramson please review

credentials/apps/credentials/utils.py
clintonb: These settings are crucial to your service operating correctly. Realistically, the credentials service should NOT start if these settings are not available. If you simply catch/log the exception, you need to have monitoring in place for this log message. This means a few dozen users will see an error message before an alert is fired, someone notices the alert, the issue is investigated, and a solution finally put in place. That's a long time.

Let's say you added a check at startup. When we deploy to stage/prod the service doesn't start. If the service doesn't start we don't even switch the ELB. No users are affected. No time is wasted debugging.

You want to fail early. In some cases, catching/logging the exception would be fine. In this case, when the error is clearly associated with a person having the invalid settings, an error from which the software cannot recover without external intervention, you should not even start the program.

credentials/apps/credentials/utils.py
clintonb: See my note above. You can save yourself a lot of headache in terms of testing and later debugging if you validate these settings at startup.

credentials/apps/credentials/tests/mixins.py
```diff
@@ -177,3 +177,79 @@ def mock_programs_api(self, data=None, status_code=200):
...
+        if username == 'test-user':
+            data = self.USER_API_RESPONSE
+        else:
+            data = None
+            status_code = 404
+
+        body = json.dumps(data)
+
+        httpretty.reset()
+        httpretty.register_uri(httpretty.GET, url, body=body, content_type='application/json', status=status_code)
```
clintonb: Move `'application/json'` to a constant and update all uses in this file.

credentials/apps/credentials/tests/mixins.py
```diff
@@ -177,3 +177,79 @@ def mock_programs_api(self, data=None, status_code=200):
...
+
+
+class UserDataMixin(object):
+    """ Mixin mocking User API URLs and providing fake data for testing."""
+    USER_API_RESPONSE = {
+        "username": "test-user",
+        "bio": "A test user.",
```
clintonb: Unless you need specific fields for testing, you don't need all of this data. You could get by with just a username.

credentials/apps/credentials/tests/mixins.py
```diff
@@ -177,3 +177,79 @@ def mock_programs_api(self, data=None, status_code=200):
...
         httpretty.reset()
         httpretty.register_uri(httpretty.GET, url, body=body, content_type='application/json', status=status_code)
+
+
+class OrganizationsDataMixin(object):
+    """ Mixin mocking Organizations API URLs and providing fake data for testing."""
+    ORGANIZATIONS_API_RESPONSE = {
+        'name': 'Test Organization',
+        'short_name': 'test-org',
+        'description': 'Oraganization for testing.',
```
clintonb: Organization.

credentials/apps/credentials/tests/test_utils.py
```diff
@@ -137,3 +138,226 @@ def test_get_programs_with_no_data(self):
...
+
+
+@ddt.ddt
+class TestOrganizationRetrieval(OrganizationsDataMixin, TestCase):
+    """
+    Tests covering the retrieval of organization from LMS.
+    """
+    def setUp(self):
+        super(TestOrganizationRetrieval, self).setUp()
+        # organization key of an existing organization
```
clintonb: Remove this comment, and change the variable name to `organization_key`.

credentials/apps/credentials/tests/mixins.py
```diff
@@ -177,3 +177,79 @@ def mock_programs_api(self, data=None, status_code=200):
...
+
+    def mock_user_api(self, username, status_code=200):
+        """ Utility for mocking out User API URLs."""
+        self.assertTrue(httpretty.is_enabled(), msg='httpretty must be enabled to mock User API calls.')
+
+        user_api_url = settings.USER_API_URL
+        url = user_api_url.strip('/') + '/accounts/' + username
+
+        # only return data for test user
+        if username == 'test-user':
```
clintonb: Split this into two methods: `mock_user_api(username)` and `mock_user_api_404()`. Both can call the same private method, `_mock_user_api(data, status=200)`

credentials/apps/credentials/tests/mixins.py
```diff
@@ -177,3 +177,79 @@ def mock_programs_api(self, data=None, status_code=200):
...
+
+    def mock_organizations_api(self, organization_key, status_code=200):
+        """ Utility for mocking out Organizations API URLs."""
+        self.assertTrue(httpretty.is_enabled(), msg='httpretty must be enabled to mock Organizations API calls.')
+
+        organizations_api_url = settings.ORGANIZATIONS_API_URL
+        url = organizations_api_url.strip('/') + '/organization/' + organization_key + '/'
+
+        # only return data for test org
+        if organization_key == 'test-org':
```
clintonb: Same as below. Split this method into two.

credentials/apps/credentials/tests/test_utils.py
```diff
@@ -137,3 +138,226 @@ def test_get_programs_with_no_data(self):
...
+        self.assertEqual(
+            actual_organizations_api_response,
+            self.ORGANIZATIONS_API_RESPONSE
+        )
+
+        # verify the API was actually hit (not the cache)
+        self.assertEqual(len(httpretty.httpretty.latest_requests), 1)
+
+    @httpretty.activate
+    @patch('django.conf.settings.ORGANIZATIONS_CACHE_TTL', 1)
```
clintonb: Use [`override_settings`](https://docs.djangoproject.com/en/1.8/topics/testing/tools/#django.test.override_settings).

credentials/apps/credentials/tests/test_utils.py
```diff
@@ -137,3 +138,226 @@ def test_get_programs_with_no_data(self):
...
+        for _ in range(2):
+            get_organization(self.org_key)
+
+        # verify that only one request has been made
+        self.assertEqual(len(httpretty.httpretty.latest_requests), 1)
+
+    @patch('edx_rest_api_client.client.EdxRestApiClient.__init__')
+    def test_get_organizations_client_initialization_failure(self, mock_init):
+        """
+        Verify the behavior when API client fails to initialize.
```
clintonb: Verify what behavior?

credentials/apps/credentials/tests/test_utils.py
```diff
@@ -137,3 +138,226 @@ def test_get_programs_with_no_data(self):
...
+    @httpretty.activate
+    @ddt.data(
+        'ORGANIZATIONS_API_URL',
+        'CREDENTIALS_SERVICE_USER',
+        'ORGANIZATIONS_AUDIENCE',
+        'ORGANIZATIONS_SECRET_KEY',
+    )
+    def test_get_organizations_client_settings_failure(self, setting_attribute):
+        """
+        Verify the behavior when API client fails to initialize due to
```
clintonb: What behavior? Your test docstrings should be explicit about expected behavior.

credentials/apps/credentials/tests/test_utils.py
```diff
@@ -137,3 +138,226 @@ def test_get_programs_with_no_data(self):
...
+
+
+@ddt.ddt
+class TestUserRetrieval(UserDataMixin, TestCase):
+    """
+    Tests covering the retrieval of user from LMS.
+    """
+    def setUp(self):
+        super(TestUserRetrieval, self).setUp()
+        # username of an existing user
```
clintonb: This comment is not necessary.

credentials/apps/credentials/tests/test_utils.py
```diff
@@ -137,3 +138,226 @@ def test_get_programs_with_no_data(self):
...
+        self.assertEqual(
+            actual_user_api_response,
+            self.USER_API_RESPONSE
+        )
+
+        # verify the API was actually hit (not the cache)
+        self.assertEqual(len(httpretty.httpretty.latest_requests), 1)
+
+    @httpretty.activate
+    @patch('django.conf.settings.ORGANIZATIONS_CACHE_TTL', 1)
```
clintonb: `override_settings`

credentials/apps/credentials/tests/test_utils.py
```diff
@@ -137,3 +138,226 @@ def test_get_programs_with_no_data(self):
...
+
+        with patch('credentials.apps.credentials.utils.settings', patched_settings):
+            with LogCapture(LOGGER_NAME) as log:
+                actual_user_api_response = get_user(self.username)
+                self.assertEqual(actual_user_api_response, {})
+                log.check((LOGGER_NAME, 'ERROR', expected_error_msg))
+
+    def test_get_user_with_unicode_key(self):
+        """
+        Verify the behavior when the method 'get_user' fails to fetch the
```
clintonb: What behavior?

credentials/apps/credentials/utils.py
```diff
@@ -58,16 +61,95 @@ def get_programs():
...
+        organization_key (str): Unique key of the organization for retrieval
+
+    Returns:
+        dict, representing organization data returned by the LMS.
+    """
+    no_organization = {}
+    # ensure that the 'organization_key' contains only ASCII characters
+    try:
+        organization_key.decode('ascii')
+    except UnicodeDecodeError:
```
clintonb: Why would this ever be an issue? If the API supports non-ASCII characters, your code should as well. If the API does not support non-ASCII characters, how would be ever get a non-ASCII key?

credentials/apps/credentials/utils.py
```diff
@@ -58,16 +61,95 @@ def get_programs():
...
+        log.exception('Invalid organization key %s.', organization_key)
+        return no_organization
+
+    cache_key = ORGANIZATIONS_CACHE_KEY + '.{}'.format(_make_hash(organization_key))
```
clintonb: Get rid of the string concatenation. Use `format()` only.

credentials/apps/credentials/utils.py
```diff
@@ -58,16 +61,95 @@ def get_programs():
...
+    organization = cache.get(cache_key)
+
+    if organization is not None:
```
clintonb: Use `if organization`. If we, for some reason, cache an empty dict, we should check the API again, because an empty dict is not valid.

credentials/apps/credentials/utils.py
```diff
@@ -58,16 +61,95 @@ def get_programs():
...
+        return organization
+
+    organizations_api = get_organizations_api_client()
+    if organizations_api is None:
+        return no_organization
```
clintonb: How will the application respond to an empty dictionary instead of organization data? I would raise an exception and let the consuming code decide how to deal with this issue. The more you try to squash the exceptions, the more brittle the consuming code will become as it tries to work around the empty dictionary return value.

credentials/apps/credentials/utils.py
```diff
@@ -58,16 +61,95 @@ def get_programs():
...
+
+    try:
+        organization = organizations_api.organization(organization_key).get()
+        cache.set(cache_key, organization, settings.ORGANIZATIONS_CACHE_TTL)
+    except Exception:  # pylint: disable=broad-except
+        log.exception('Failed to retrieve organization with key %s from the Organization API.', organization_key)
+        return no_organization
```
clintonb: Same as above. Just let the exception bubble up, or catch it and raise a custom exception.

credentials/apps/credentials/utils.py
```diff
@@ -58,16 +61,95 @@ def get_programs():
...
+        username (str): Unique identifier of the user for retrieval
+
+    Returns:
+        dict, representing user data returned by the LMS.
+    """
+    no_user = {}
+    # ensure that the 'username' contains only ASCII characters
+    try:
+        username.decode('ascii')
+    except UnicodeDecodeError:
```
clintonb: As with the Organizations API, why are we accounting for this?

credentials/apps/credentials/utils.py
```diff
@@ -58,16 +61,95 @@ def get_programs():
...
+    """ Get the user data from the edx-platform on behalf of the
+    credentials service user present in the edx-platform system.
+    Returned value is cached to avoid calling LMS each time a certificate is
+    viewed.
+
+    Arguments:
+        username (str): Unique identifier of the user for retrieval
+
+    Returns:
+        dict, representing user data returned by the LMS.
```
clintonb: s/LMS/User API

credentials/apps/credentials/utils.py
```diff
@@ -58,16 +61,95 @@ def get_programs():
...
+        cache.set(cache_key, organization, settings.ORGANIZATIONS_CACHE_TTL)
+    except Exception:  # pylint: disable=broad-except
+        log.exception('Failed to retrieve organization with key %s from the Organization API.', organization_key)
+        return no_organization
+
+    return organization
+
+
+def get_user(username):
+    """ Get the user data from the edx-platform on behalf of the
```
clintonb: Retrieve the user details from the User API.

If the API call is successful, the returned data will be cached for the duration of USER_CACHE_TTL (in seconds). Failed API responses will NOT be cached.

credentials/apps/credentials/utils.py
```diff
@@ -58,16 +61,95 @@ def get_programs():
...
+
+    cache_key = USER_CACHE_KEY + '.{}'.format(_make_hash(username))
+    user = cache.get(cache_key)
+
+    if user is not None:
+        return user
+
+    user_api = get_user_api_client()
+    if user_api is None:
+        return no_user
```
clintonb: As with organizations, don't return empty data. Raise an exception and let the downstream code deal with this.

credentials/apps/credentials/utils.py
```diff
@@ -87,7 +169,46 @@ def get_program_api_client():
...
+        service_username = settings.CREDENTIALS_SERVICE_USER
+        jwt_audience = settings.USER_JWT_AUDIENCE
+        jwt_secret_key = settings.USER_JWT_SECRET_KEY
+    except AttributeError:
+        log.exception("Failed to get settings for communication with the User API. Please make sure that the "
+                      "settings for 'USER_API_URL', 'CREDENTIALS_SERVICE_USER', 'USER_JWT_AUDIENCE', "
+                      "'USER_JWT_SECRET_KEY' are provided.")
+        return None
+
+    # user api don't accept url with trailing slash so make `append_slash` False
```
clintonb: If that is the case, you should explicitly set `append_slash=False` in this call rather than rely on (a) this comment, and (b) the default values always remaining the same.

credentials/apps/credentials/utils.py
```diff
@@ -87,7 +169,46 @@ def get_program_api_client():
...
+    return _get_service_user_api_client(user_api_url, service_username, jwt_audience, jwt_secret_key, False)
+
+
+def _get_service_user_api_client(api_url, service_username, jwt_audience, jwt_secret_key, append_slash=True):
```
clintonb: Replace `append_slash=True` with `**kwargs`.

credentials/apps/credentials/utils.py
```diff
@@ -87,7 +169,46 @@ def get_program_api_client():
...
     """
     Helper method to get edx rest api client for the provided service user
```
clintonb: Returns an API client authenticated as the service user.

credentials/apps/credentials/utils.py
zubair-arbi: We are not hiding any errors here. The exception will be logged, also we don't want the code to crash for exception like `HttpNotFoundError`.

zubair-arbi:
@clintonb I have addressed your feedback please review

credentials/apps/credentials/utils.py
clintonb: Catch the exceptions and raise your own. The issue here is that you are returning an empty dict to code expecting information about an organization. What is the code supposed to do with the empty dict? If the code needs to check if the return value is an empty dict, you are essentially re-writing exception handling, which is not necessary or good practice.

```python
key = 'edx'

# This is most likely how a consumer would use this function as it is currently written.
organization = get_organization(key)

if organization == {}:
  # Handle missing organization: render error?
else:
  # Handle found organization: render certificate?
```

```python
# This is how the function would be used with exception handling
key = 'edx'

try:
  organization = get_organization(key)
except OrganizationError:
  # Handle missing organization: render error, halt execution

# Handle ideal case with organization.
```

Compare the two blocks above. In the first block, your function caller needs to know about a special return value. If the function caller fails to account for the special value, the user may end up seeing something unexpected. However, if the function instead raises an exception, there is no room for error regarding the return value because there isn't one. If the caller fails to catch the exception, Django will ultimately render a 500 page. If the caller does catch the exception, it can be handled appropriately.

My point is that you should not use a special return value to handle exceptional cases. That's the job of exceptions.

credentials/apps/credentials/tests/test_utils.py
```diff
@@ -130,10 +100,155 @@ def test_get_programs_data_retrieval_failure(self):
...
+        self.mock_organizations_api_404(organization_key)
+
+        actual_organizations_api_response = get_organization(organization_key)
+        self.assertEqual(actual_organizations_api_response, {})
+
+
+@ddt.ddt
+class TestUserRetrieval(UserDataMixin, TestCase):
+    """
+    Tests covering the retrieval of user from LMS.
```
clintonb: Tests for get_user.

credentials/apps/credentials/tests/test_utils.py
```diff
@@ -130,10 +100,155 @@ def test_get_programs_data_retrieval_failure(self):
...
         self.mock_programs_api(data={'results': []})
 
         actual_programs_api_response = get_programs()
         self.assertEqual(actual_programs_api_response, [])
+
+
+@ddt.ddt
+class TestOrganizationRetrieval(OrganizationsDataMixin, TestCase):
+    """
+    Tests covering the retrieval of organization from LMS.
```
clintonb: Tests for get_organization.

credentials/apps/credentials/tests/test_utils.py
```diff
@@ -130,10 +100,155 @@ def test_get_programs_data_retrieval_failure(self):
 
     @httpretty.activate
     def test_get_programs_with_no_data(self):
-        """ Verify the behavior when no programs data is found from the
-        Programs service.
+        """ Verify that the method 'get_programs' returns empty dict if
```
clintonb: s/method/function

credentials/apps/credentials/apps.py
```diff
@@ -0,0 +1,40 @@
...
+from django.apps import AppConfig
+from django.conf import settings
+
+
+log = logging.getLogger(__name__)
+
+
+class CredentialsConfig(AppConfig):
+    """
+    Basic configuration validation.
```
clintonb: Credentials app configuration.

credentials/apps/credentials/apps.py
```diff
@@ -0,0 +1,40 @@
...
+    """
+    name = 'credentials.apps.credentials'
+    verbose_name = 'Credentials'
+
+    def ready(self):
+
+        try:
+            settings.PROGRAMS_API_URL and settings.PROGRAMS_JWT_AUDIENCE and settings.PROGRAMS_JWT_SECRET_KEY
+        except AttributeError:
```
clintonb: The point of this validation is to prevent the application from starting if any of these settings are missing. You are catching the exceptions, logging them, and continuing with program execution. Don't catch the exception. Let the app fail.

credentials/apps/credentials/apps.py
```diff
@@ -0,0 +1,40 @@
...
+class CredentialsConfig(AppConfig):
+    """
+    Basic configuration validation.
+    """
+    name = 'credentials.apps.credentials'
+    verbose_name = 'Credentials'
+
+    def ready(self):
+
+        try:
```
clintonb: Your code should look like this:

```python
required_settings = ('PROGRAMS_API_URL', 'PROGRAMS_JWT_AUDIENCE',)

for setting in required_settings:
  if not hasattr(settings, setting):
    msg = 'The setting {} must be set in order to start the application!'.format(setting)
    log.critical(msg)
    raise AttributeError(msg)
```

jimabramson: As long as we're implementing this check (which I agree with), I'd suggest a minor change which is to first check all the config attributes and then raise the exception including all of them.  (This would prevent a whack-a-mole situation where there is a larger misconfiguration problem, but the deployer has to discover each missing attribute and resolve one at a time.)

credentials/apps/credentials/tests/mixins.py
```diff
@@ -176,4 +178,75 @@ def mock_programs_api(self, data=None, status_code=200):
...
+        self._mock_organizations_api(organization_key, data)
+
+    def mock_organizations_api_404(self, organization_key):
+        """ Utility for mocking out Organizations API URLs with status code 200."""
+        data = self.ORGANIZATIONS_API_RESPONSE
+        self._mock_organizations_api(organization_key, data, 404)
+
+    def mock_organizations_api_500(self, organization_key):
+        """ Utility for mocking out Organizations API URLs with status code 200."""
+        data = self.ORGANIZATIONS_API_RESPONSE
```
jimabramson: nit: the `data` argument is always the same so why not move it inside `_mock_organizations_api`

credentials/apps/credentials/tests/test_apps.py
```diff
@@ -0,0 +1,84 @@
...
+        patched_settings = copy.deepcopy(settings)
+        delattr(patched_settings, setting_attribute)
+        expected_error_msg = ("Failed to get settings for communication with the User API. Please make sure "
+                              "that the settings for 'USER_API_URL', 'USER_JWT_AUDIENCE', "
+                              "'USER_JWT_SECRET_KEY' are provided.")
+
+        with mock.patch('credentials.apps.credentials.apps.settings', patched_settings):
+            with LogCapture(LOGGER_NAME) as log:
+                self.app_config.ready()
+                log.check((LOGGER_NAME, 'ERROR', expected_error_msg))
```
jimabramson: it might be worth rolling all the settings check tests - and the settings check itself - into a single method instead of one each per API.  Especially if the app will not run when any of them is missing.

credentials/apps/credentials/tests/test_utils.py
```diff
@@ -94,34 +89,9 @@ def test_get_programs_client_initialization_failure(self, mock_init):
...
-            with LogCapture(LOGGER_NAME) as log:
-                actual_programs_api_response = get_programs()
-                self.assertEqual(actual_programs_api_response, [])
-                log.check((LOGGER_NAME, 'ERROR', expected_error_msg))
-
-    @httpretty.activate
     def test_get_programs_data_retrieval_failure(self):
-        """
-        Verify the behavior when data can't be retrieved from Programs.
+        """ Verify that the data can't be retrieved from Programs API if
```
jimabramson: Verify that an empty list will be returned if the data can't be retrieved from the Programs API due to a server error

credentials/apps/credentials/utils.py
jimabramson: I agree that surfacing specific exceptions lead to cleaner and healthier code (not to mention more intelligible errors/logs when something goes wrong).  It's somewhat of a departure from the way we've recently written code in other places (e.g. the xseries messages in the student dashboard) but I think this is a better way to go.

credentials/apps/credentials/utils.py
```diff
@@ -53,41 +56,125 @@ def get_programs():
...
+        return user
+
+    user_api = get_user_api_client()
+
+    try:
+        user = user_api.accounts(username).get()
+        cache.set(cache_key, user, settings.USER_CACHE_TTL)
+    except Exception:  # pylint: disable=broad-except
+        log.exception('Failed to retrieve user with username %s from the User API.', username)
+        return no_user
```
jimabramson: lets not forget to apply the same pattern being discussed above, here as well

zubair-arbi:
@clintonb @jimabramson please review.

zubair-arbi:
@clintonb @jimabramson This PR is ready for final review.

credentials/apps/credentials/tests/test_apps.py
```diff
@@ -0,0 +1,51 @@
...
+        'ORGANIZATIONS_SECRET_KEY',
+        'USER_API_URL',
+        'USER_JWT_AUDIENCE',
+        'USER_JWT_SECRET_KEY',
+    )
+    def test_credentials_config_settings_failure(self, setting_attribute):
+        """ Verify that the exception is raised when there are missing settings
+        for the credentials app.
+        """
+        patched_settings = copy.deepcopy(settings)
```
clintonb: No need for the copying/patching. Add the `@override_settings()` decorator and call `delattr(settings, setting_attribute)`.

credentials/apps/credentials/utils.py
```diff
@@ -53,41 +56,109 @@ def get_programs():
...
-        return no_programs
-
-    try:
-        response = programs_api.programs.get()
-    except Exception:  # pylint: disable=broad-except
-        log.exception('Failed to retrieve programs from the Programs API.')
-        return no_programs
-
+    response = programs_api.programs.get()
     results = response.get('results', no_programs)
```
clintonb: If the response is missing the results key, there is probably a bigger issue at hand. I'm inclined to let this bubble up and render an error, rather than hide the fact that the API response is invalid.

clintonb: Actually...what is this API call supposed to be doing? Why are we retrieving all programs? Note that this will only get the first page of results.

zubair-arbi: Thanks for catching this. We are getting all programs at a time since there are not much programs and yes you are right that this will only get first page. I will change the code to get a single program to get it consistent with how we are retrieving data from other APIs.

zubair-arbi:
@clintonb I have addressed your feedback and updated the programs fetching functionality. Please review.

credentials/apps/credentials/tests/mixins.py
```diff
@@ -27,153 +29,165 @@ class ProgramsDataMixin(object):
...
     }
 
-    def mock_programs_api(self, data=None, status_code=200):
+    def mock_programs_api(self, program_id):
+        """ Utility for mocking out Programs API URLs with status code 200."""
+        data = self.PROGRAMS_API_RESPONSE
+        self._mock_programs_api(program_id, data)
+
+    def mock_programs_api_404(self, program_id):
+        """ Utility for mocking out Programs API URLs with status code 200."""
```
clintonb: 404

credentials/apps/credentials/tests/mixins.py
```diff
@@ -27,153 +29,165 @@ class ProgramsDataMixin(object):
...
+        data = None
+        self._mock_programs_api(program_id, data, 404)
+
+    def mock_programs_api_500(self, program_id):
+        """ Utility for mocking out Programs API URLs with status code 200."""
```
clintonb: 500

credentials/apps/credentials/tests/mixins.py
```diff
@@ -27,153 +29,165 @@ class ProgramsDataMixin(object):
...
+        data = None
+        self._mock_programs_api(program_id, data, 500)
+
+    def _mock_programs_api(self, program_id, data, status_code=200):
         """ Utility for mocking out Programs API URLs."""
         self.assertTrue(httpretty.is_enabled(), msg='httpretty must be enabled to mock Programs API calls.')
 
         programs_api_url = settings.PROGRAMS_API_URL
-        url = programs_api_url.strip('/') + '/programs/'
+        url = programs_api_url.strip('/') + '/programs/' + str(program_id) + '/'
```
clintonb:
```python
url = '{root}/programs/{program_id}/'.format(root=programs_api_url.strip('/'), program_id=program_id)
```

credentials/apps/credentials/utils.py
```diff
@@ -36,36 +39,82 @@ def keyfunc(attribute):  # pylint: disable=missing-docstring
     return True
 
 
-def get_programs():
-    """ Get active programs from the Programs service on behalf of the
+def get_program(program_id):
+    """ Get active program from the Programs service on behalf of the
```
clintonb: Retrieve program detail from the Programs API.

credentials/apps/credentials/utils.py
```diff
@@ -36,36 +39,82 @@ def keyfunc(attribute):  # pylint: disable=missing-docstring
...
     credentials service user present in the Programs service.
 
     Returned value is cached to avoid calling programs service each time a
     certificate is viewed.
 
+    Arguments:
+        program_id (str): Unique id of the program for retrieval
```
clintonb: int

clintonb:
:+1: (after remaining comments addressed)

PR #41, opened 2016-01-25
Increased verbosity of static files commands

clintonb:
Limited output may be nice for local development, but is frustrating for failed runs non-development environments. Developers can live with extra output.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,8 +59,8 @@ quality:
 
 static:
 	$(NODE_BIN)/r.js -o build.js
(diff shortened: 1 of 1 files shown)


clintonb:
FYI @edx/ecommerce no more decreased verbosity for the sake of local development. If a command is being run outside of development, it needs to be verbose.

rlucioni:
:+1:

jimabramson:
:+1: :+1:

peter-fogg:
:+1:

PR #45, opened 2016-01-27
add default css filter in base settings

zubair-arbi:
Add default css filter `CssAbsoluteFilter` from django compressor to base settings to resolve the issue of bower static files paths in css on production.


ahsan-ul-haq:
LGTM :+1:

awais786:
:+1:

PR #49, opened 2016-01-28
show title from certificate confiuration if set

zubair-arbi:
ECOM-3557
* Show `title` from the program certificate configurations if set, instead of the program name while rendering the certificate
* Add the required field `logo_url` for model `SiteConfiguration` and make the field `theme_scss_path` not required since we are using default stylng.
* Move `SiteConfiguration` model from app `credentials` to `core`
* Add migration for changes in model `SiteConfiguration`
* User site configuration for showing site logo and linking to its lms url
* Add/Update test for above changes


credentials/apps/credentials/tests/test_views.py
```diff
@@ -20,7 +20,7 @@ class RenderCredentialPageTests(TestCase):
 
     def setUp(self):
         super(RenderCredentialPageTests, self).setUp()
-        self.program_certificate = factories.ProgramCertificateFactory.create(template=None)
+        self.program_certificate = factories.ProgramCertificateFactory.create(template=None, title=None)
```
awais786: No need to pass None.

zubair-arbi:
@awais786 I have updated the PR. Please review again.

awais786:
:+1:

zubair-arbi:
@clintonb @peter-fogg can anyone of you please review?

credentials/templates/credentials/program_certificate.html
```diff
@@ -9,9 +9,9 @@
             <header class="header-app" role="banner">
                 <h1 class="header-app-title">
                     <a class="logo" href="http://edx.org">
```
clintonb: The edX URL should not be hard-coded, nor should other mentions of edX.

credentials/templates/credentials/program_certificate.html
```diff
@@ -88,7 +88,13 @@ <h4 class="organization-name">{{ certificate_context.organization_data.name }}</
                                     <span class="accomplishment-summary copy">successfully completed all courses in the XSeries Program</span>
                                 {% endblocktrans %}
                                 <span class="accomplishment-program">
-                                    <span class="accomplishment-program-name">{{ certificate_context.programs_data.name }}</span>
+                                    <span class="accomplishment-program-name">
+                                        {% if certificate_context.credential_title.strip %}
```
clintonb: Why the need for `strip`? We should never store invalid data in the database. Stripping should occur at the point the record is created so that we can avoid littering the code with checks like these.

clintonb: Once the `strip` call is removed, you may be able to use [`firstof`](https://docs.djangoproject.com/en/1.9/ref/templates/builtins/#firstof).

peter-fogg:
Agreed with @clintonb on the template changes. Otherwise this looks good, so :+1: when that's fixed.

zubair-arbi:
@clintonb I have now modified PR to use the site configuration for displaying the logo and and linking to its lms url. Please review

credentials/apps/credentials/models.py
```diff
@@ -82,12 +82,20 @@ class SiteConfiguration(models.Model):
...
+        null=False,
+        blank=False
+    )
     theme_scss_path = models.CharField(
         verbose_name=_('Path to custom site theme'),
         help_text=_('Path to scss files of the custom site theme'),
         max_length=255,
         null=False,
-        blank=False
+        blank=True
```
clintonb: This should never be blank. Every site needs a theme.

credentials/apps/core/context_processors.py
```diff
@@ -1,10 +1,12 @@
 """ Core context processors. """
 from django.conf import settings
+from django.contrib.sites.models import Site
 
 
 def core(request):
     """ Site-wide context processor. """
     return {
+        'site': Site.objects.get_current(),
```
clintonb: request.site, assuming you have the middleware enabled.

credentials/apps/core/tests/test_context_processors.py
```diff
@@ -17,5 +18,6 @@ def test_core(self):
         request = RequestFactory().get('/')
         request.LANGUAGE_CODE = LANGUAGE_CODE
         self.assertDictEqual(
-            core(request), {'platform_name': PLATFORM_NAME, 'language_code': LANGUAGE_CODE}
+            core(request),
+            {'site': Site.objects.get_current(), 'platform_name': PLATFORM_NAME, 'language_code': LANGUAGE_CODE}
```
clintonb: Move this to a variable, with each key on its own line. Otherwise, every change will require changing the entire line, making doffs more difficult.

credentials/apps/credentials/models.py
zubair-arbi: @clintonb You are right but since we are using the default theme so it make sense to skip this value until you want to change the theme for a specific site.

clintonb: Nope. Every site should require this to be explicitly set. Otherwise, it is too easy for someone to forget it and (even if for a short time) have a site that is misconfigured. You want to make it difficult for people to do the wrong thing. That means, this field needs to be required. The `logo_url` field should also be required. In fact, let's set the rule now: every site configuration field is required. No defaults.

credentials/apps/credentials/models.py
```diff
@@ -82,12 +82,20 @@ class SiteConfiguration(models.Model):
         null=False,
         blank=False
     )
+    logo_url = models.URLField(
+        verbose_name=_('Logo url for custom site/microsite'),
```
clintonb: We do not use microsites. That term should be removed from edX vocabulary.

`Logo URL for site`

credentials/apps/credentials/models.py
zubair-arbi: @clintonb Ok that makes sense. So what should be the path for default styling?

clintonb: The path for edX styling is the path to the edX SASS.

credentials/apps/credentials/models.py
```diff
@@ -82,12 +82,20 @@ class SiteConfiguration(models.Model):
         null=False,
```
clintonb: I just realized this model is in the `credentials` app. It belongs in `core`.

credentials/templates/credentials/program_certificate.html
```diff
@@ -8,10 +8,10 @@
         <div class="wrapper-header">
             <header class="header-app" role="banner">
                 <h1 class="header-app-title">
-                    <a class="logo" href="http://edx.org">
-                        <img class="logo-img" src="{% static 'images/logo-edX.png' %}" alt="{{ platform_name }} {% trans 'Home' %}" />
+                    <a class="logo" href="{{ site.siteconfiguration.lms_url_root }}">
+                        <img class="logo-img" src="{{ site.siteconfiguration.logo_url }}" alt="{% firstof site.name platform_name %}" />
```
clintonb: If you set the logo via CSS, you don't need the `logo_url` to be configured for the `SiteConfiguration`. See https://github.com/edx/ecommerce/blob/master/ecommerce/static/sass/themes/default.scss and https://github.com/edx/ecommerce/blob/master/ecommerce/static/sass/themes/edx.scss.

zubair-arbi: @clintonb this is not the scope of current story.
I have created a separate story [ECOM-3595](https://openedx.atlassian.net/browse/ECOM-3595) to add site theming support.

zubair-arbi:
@clintonb Please review

.travis.yml
```diff
@@ -17,6 +17,7 @@ install:
     - make requirements
 script:
     # Compile assets and run validation
+    - make clean clean_static
```
clintonb: This should not be necessary since we are starting with a fresh server.

credentials/apps/core/models.py
```diff
@@ -1,10 +1,39 @@
 """ Core models. """
-
+# pylint: disable=model-missing-unicode
```
clintonb: Add the `__unicode__` method.

clintonb:
Moving the SiteConfiguration model must be a separate PR. That is too big of a change to combine with another PR.

credentials/templates/credentials/program_certificate.html
clintonb: Given that you are adding `logo_url` in this PR, it is in scope. Why add an unnecessary database column? If you handle this in CSS you (a) avoid the need for a migration and (b) ensure the logo assets are being served from S3 alongside other static assets. This isn't that difficult. I've given you two examples. If you want to wait until you start work on theming, do so; but, don't add the `logo_url` field, given that this PR is supposed to be about displaying a title.

zubair-arbi:
Closing this PR in favor of https://github.com/edx/credentials/pull/53

PR #54, opened 2016-02-03
static files s3 settings

zubair-arbi:
ECOM-3605
* Remove settings for S3 as these settings will come from configuration https://github.com/edx/configuration/pull/2758/
* Fix the images path in file `_config.scss` to work with S3

Sample S3 config settings:
```yml
AWS_STORAGE_BUCKET_NAME: mybucket
AWS_CUSTOM_DOMAIN: mybucket.s3.amazonaws.com
AWS_ACCESS_KEY_ID: XXXAWS_ACCESS_KEYXXX
AWS_SECRET_ACCESS_KEY: XXXAWS_SECRETY_KEYXXX
AWS_QUERYSTRING_AUTH: False
AWS_QUERYSTRING_EXPIRE: False
AWS_DEFAULT_ACL: ''
AWS_HEADERS:
  Cache-Control: max-age-31536000
  Access-Control-Allow-Origin: PUT YOUR HOSTNAME HERE

COMPRESS_URL: 'https://mybucket.s3.amazonaws.com/'
STATIC_URL: 'https://mybucket.s3.amazonaws.com/'
COMPRESS_ROOT: {{ CREDENTIALS_STATIC_ROOT }}
COMPRESS_STORAGE: 'storages.backends.s3boto.S3BotoStorage'
STATICFILES_STORAGE: 'storages.backends.s3boto.S3BotoStorage'
DEFAULT_FILE_STORAGE: 'storages.backends.s3boto.S3BotoStorage'
```
These settings will come from configuration: https://github.com/edx/configuration/pull/2758/

### DIFF
--- a/credentials/settings/production.py
+++ b/credentials/settings/production.py
@@ -20,6 +20,9 @@ with open(CONFIG_FILE) as f:
     config_from_yaml = yaml.load(f)
     vars().update(config_from_yaml)
 
+    # Load the files storage backend settings for django storages
+    vars().update(FILE_STORAGE_BACKEND)
+
 DB_OVERRIDES = dict(
     PASSWORD=environ.get('DB_MIGRATION_PASS', DATABASES['default']['PASSWORD']),
     ENGINE=environ.get('DB_MIGRATION_ENGINE', DATABASES['default']['ENGINE']),
@@ -31,9 +34,6 @@ DB_OVERRIDES = dict(
 
 for override, value in DB_OVERRIDES.iteritems():
     DATABASES['default'][override] = value
-
-if AWS_STORAGE_BUCKET_NAME and AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY:
-    DEFAULT_FILE_STORAGE = 'storages.backends.s3boto.S3BotoStorage'
 
 JWT_AUTH.update({
     'JWT_SECRET_KEY': SOCIAL_AUTH_EDX_OIDC_SECRET,
--- a/credentials/static/sass/_config.scss
+++ b/credentials/static/sass/_config.scss
@@ -6,10 +6,10 @@
 // ------------------------------
 // #VARIABLES
 // ------------------------------
- $font-path: '../bower_components/edx-pattern-library/pattern-library/fonts';
+$font-path: '../bower_components/edx-pattern-library/pattern-library/fonts';
 
-$accomplishment-verified-path: '/static/images';
+$accomplishment-verified-path: '../images';
 
-$accomplishment-professional-path: '/static/images';
+$accomplishment-professional-path: '../images';
 
-$accomplishment-xseries-path: '/static/images';
+$accomplishment-xseries-path: '../images';


zubair-arbi:
@maxrothman this is the PR for changes in credentials service for S3 support. Please review

maxrothman:
Please use the pattern used with profile images. You can see how the settings get ingested [here](https://github.com/edx/edx-platform/blob/master/lms/envs/aws.py#L705) and how they get used [here](https://github.com/edx/edx-platform/blob/8ed051e3cfcf82bfa0a3a4f99c89ff1d21b5d603/openedx/core/djangoapps/user_api/accounts/image_helpers.py#L27).

credentials/settings/production.py
```diff
@@ -32,8 +32,25 @@
 for override, value in DB_OVERRIDES.iteritems():
     DATABASES['default'][override] = value
 
-if AWS_STORAGE_BUCKET_NAME and AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY:
-    DEFAULT_FILE_STORAGE = 'storages.backends.s3boto.S3BotoStorage'
+if FILE_STORAGE_BACKEND and FILE_STORAGE_BACKEND['class'] == 'storages.backends.s3boto.S3BotoStorage':
+    storage_class = FILE_STORAGE_BACKEND['class']
+    storage_options = FILE_STORAGE_BACKEND['options']
+
+    AWS_STORAGE_BUCKET_NAME = storage_options['bucket']
```
zubair-arbi: @maxrothman @cpennington Since we are using the `django-storages` and `django-compressor` so I have disassembled the dict `FILE_STORAGE_BACKEND` here to set the variables according to their docs.
@maxrothman I wasn't able to use the storage dict directly as for [uploading the profile images](https://github.com/edx/edx-platform/blob/8ed051e3cfcf82bfa0a3a4f99c89ff1d21b5d603/openedx/core/djangoapps/user_api/accounts/image_helpers.py#L27) since we are using S3 for both static and media files.

zubair-arbi:
@cpennington @maxrothman please review

credentials/settings/production.py
cpennington: You should be able to set all of the variables via configuration. Line 21 sets all configuration values from the yaml file as python variables. So if you set all of these in the yaml, then you don't need to set them explicitly in this file.

zubair-arbi: @cpennington I used that approach previously but @maxrothman told to pass the dict `FILE_STORAGE_BACKEND` directly and django should be responsible to enabling that setting.

cpennington: Sorry, I think you misunderstand. I'm suggesting that you delete all of lines 35 - 53, and add them to the block of service configuration in ansible here: https://github.com/edx/configuration/blob/master/playbooks/roles/credentials/defaults/main.yml#L86-L126, since it sounds like you're saying that for boto, it expects a bunch of separate django settings variables (on top of the options set in `FILE_STORAGE_BACKEND['options']`).

credentials/settings/production.py
```diff
@@ -32,8 +32,25 @@
...
+    AWS_QUERYSTRING_AUTH = storage_options['querystring_auth']
+    AWS_QUERYSTRING_EXPIRE = storage_options['querystring_expire']
+    AWS_DEFAULT_ACL = storage_options['bucket_acl']
+
+    COMPRESS_URL = 'https://{aws_s3_custom_domain}/'.format(aws_s3_custom_domain=AWS_CUSTOM_DOMAIN)
+    STATIC_URL = COMPRESS_URL
+    COMPRESS_ROOT = STATIC_ROOT
+    COMPRESS_STORAGE = storage_class
+    STATICFILES_STORAGE = storage_class
+    DEFAULT_FILE_STORAGE = storage_class
```
cpennington: Please remove all of these lines in favor of reading from the yaml file.

zubair-arbi:
@cpennington @maxrothman I have removed the S3 settings from django as now these settings will come from credentials configuration `credentials.yml`. Please review.

maxrothman:
:+1:

PR #57, opened 2016-02-09
Remove unnecessary padding above print cert icon

rlucioni:
@marcotuts @ahsan-ul-haq 

This change takes us from here:

To here:


marcotuts:
:+1:

PR #60, opened 2016-02-16
add S3BotoStorage for static and media files

zubair-arbi:
@awais786 @ahsan-ul-haq @tasawernawaz @waheedahmed 

Update S3BotoStorage to use 'media' folder for media files and 'staticfiles' for static files.

Related configuration changes: https://github.com/edx/configuration/pull/2781

http://stackoverflow.com/questions/10390244/how-to-set-up-a-django-project-with-django-storages-and-amazon-s3-but-with-diff/10825691#10825691
http://martinbrochhaus.com/s3.html

### DIFF
--- a/credentials/apps/core/s3utils.py
+++ b/credentials/apps/core/s3utils.py
@@ -1,0 +1,18 @@
+"""
+Custom S3 storage backends to store files in sub folders.
+"""
+from functools import partial
+
+from django.conf import settings
+from storages.backends.s3boto import S3BotoStorage
+
+
+MediaS3BotoStorage = partial(
+    S3BotoStorage,
+    location=settings.MEDIA_ROOT.strip('/')
+)
+
+StaticS3BotoStorage = partial(
+    S3BotoStorage,
+    location=settings.STATIC_ROOT.strip('/')
+)
--- a/.coveragerc
+++ b/.coveragerc
@@ -5,6 +5,7 @@ source=credentials
 omit =
     credentials/settings*
     credentials/conf*
+    credentials/apps/core/s3utils.py
     *wsgi.py
     *migrations*
     *admin.py


zubair-arbi:
@cpennington @maxrothman please review

credentials/apps/core/s3utils.py
```diff
@@ -0,0 +1,8 @@
+"""
+Custom S3 storage backends to store files in sub folders.
+"""
+from storages.backends.s3boto import S3BotoStorage
+
+
+MediaRootS3BotoStorage = lambda: S3BotoStorage(location='media')
+StaticS3BotoStorage = lambda: S3BotoStorage(location='staticfiles')
```
cpennington: Nit: you could use `functools.partial` to allow other arguments to be passed through as well.

cpennington: Actually, I think I'd upgrade this from a Nit to a Blocker.

cpennington:
:+1: after fixing `partial`.

maxrothman:
@cpennington If we're switching to using custom storage objects, would it be possible to pass our options directly into the instantiation rather than setting a bunch of global config vars?

cpennington:
@maxrothman yes, but I'm not sure it would be a benefit, since it would make us not match the package conventions/documentation.

cpennington:
Plus, we'd need to make custom classes for *any* backend we wanted to change, rather than just for boto (which is all we're doing right now).

maxrothman:
It'd allow us to have a consistent interface from the config side though. I guess I don't understand what the drawbacks would be.

zubair-arbi:
@cpennington I have updated the PR. Please let me know if this is what you have suggested.

cpennington:
We already have a consistent interface from the config side: set the CREDENTIALS_FILE_STORAGE_DICT.

cpennington:
@zubair-arbi Yeah, that's exactly it

zubair-arbi:
@jimabramson this is the PR which holds the custom AWS S3 backends for making subfolders for media and static files.

rlucioni:
@jimabramson I've added a commit here that infers S3BotoStorage locations from settings. Let me know if it's what we're after.

jimabramson:
@rlucioni thanks for this.  One question: I believe these storage classes are referenced from django settings, but they in turn reference django settings themselves to initialize; will this work?

rlucioni:
@jimabramson from what I've seen, these storage classes are referenced from Django settings using dotted import paths (strings), so they're not evaluated at the same time as the settings. Given that, I'd expect this to work.

jimabramson:
@rlucioni LGTM assuming it makes sense to @maxrothman and @zubair-arbi

zubair-arbi:
lgtm

credentials/apps/core/s3utils.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+MediaS3BotoStorage = partial(
+    S3BotoStorage,
+    location=settings.MEDIA_ROOT.strip('/').split('/')[-1]
+)
+
+StaticS3BotoStorage = partial(
+    S3BotoStorage,
+    location=settings.STATIC_ROOT.strip('/').split('/')[-1]
```
maxrothman: What's the string manipulation intended to accomplish?

rlucioni: The intent is to take a path from settings like `/edx/var/credentials/staticfiles` and reduce it to `staticfiles`. I was under the impression this is what we were after here.

maxrothman: I think there's been some confusion. There are vars for `MEDIA_URL` and `STATIC_URL` that'll be `/media/` and `/static/` for file storage and use S3 URLs for S3, respectively.

Also, for future reference, the proper way to do that is with [os.path.basename](https://docs.python.org/2/library/os.path.html#os.path.basename).

zubair-arbi:
@maxrothman @jimabramson @rlucioni I have updated the file config settings and there usage here. Please review

credentials/apps/core/s3utils.py
zubair-arbi: @maxrothman I have updated the config settings so that we can use the media and static root values directly.

rlucioni:
Looks fine to me assuming `MEDIA_ROOT` and `STATIC_ROOT` are overridden as proposed in https://github.com/edx/configuration/pull/2781. :+1: once commits are squashed.

PR #63, opened 2016-02-22
Acceptance test to view credential

awais786:
For theses tests data should be exists in lms and programs. 

1. In LMS user should have completed x-series program.

2. In test case it will try to try to credential. If test case already exist then will change in db. Then test in a test user will login on lms and from dashboard it will click the credential link.  

3. In other test case first credentials will be revoked and then again user will login and in this case there will be no credential.

### DIFF
--- a/acceptance_tests/mixins.py
+++ b/acceptance_tests/mixins.py
@@ -1,6 +1,10 @@
+import datetime
 import logging
 
-from acceptance_tests.config import LMS_EMAIL, LMS_PASSWORD
+from edx_rest_api_client.client import EdxRestApiClient
+import jwt
+
+from acceptance_tests import config
 from acceptance_tests.pages import LMSLoginPage
 
 
@@ -16,8 +20,48 @@ class LoginMixin(object):
 
     def login_with_lms(self):
         """ Visit LMS and login."""
-        email = LMS_EMAIL
-        password = LMS_PASSWORD
+        email = config.LMS_EMAIL
+        password = config.LMS_PASSWORD
 
         self.lms_login_page.browser.get(self.lms_login_page.url())  # pylint: disable=not-callable
         self.lms_login_page.login(email, password)
+
+
+class CredentialsApiMixin(object):
+    """ Mixin used for login on credentials."""
+    def setUp(self):
+        super(CredentialsApiMixin, self).setUp()
+        self.data = None
+
+    @property
+    def credential_api_client(self):
+        now = datetime.datetime.utcnow()
+        expires_in = 60
+        payload = {
+            "iss": config.OAUTH_URL,
+            "aud": config.USER_JWT_AUDIENCE,
+            "exp": now + datetime.timedelta(seconds=expires_in),
+            "iat": now,
+            "preferred_username": config.LMS_USERNAME,
+            "administrator": True,
+        }
+        try:
+            jwt_data = jwt.encode(payload, config.JWT_SECRET_KEY)
+            api_client = EdxRestApiClient(config.CREDENTIALS_API_URL, jwt=jwt_data)
+        except Exception:  # pylint: disable=broad-except
+            log.exception("Failed to initialize the API client with url '%s'.", config.CREDENTIALS_API_URL)
+            return
+        return api_client
+
+    def create_credential(self):
+        """Create user credential for a program."""
+        self.data = self.credential_api_client.user_credentials.post({
+            'username': config.LMS_USERNAME,
+            'credential': {'program_id': config.PROGRAM_ID},
+            'attributes': []
+        })
+
+    def change_credential_status(self, status):
+        """Update the credential status to awarded or revoked."""
+        self.data['status'] = status
+        self.credential_api_client.user_credentials(self.data['id']).patch(self.data)
--- a/acceptance_tests/pages.py
+++ b/acceptance_tests/pages.py
@@ -54,9 +54,22 @@ class LMSLoginPage(LMSPage):
 
 
 class LMSDashboardPage(LMSPage):
+    def __init__(self, browser):
+        super(LMSDashboardPage, self).__init__(browser)
+        self.credential_css_selector = '.wrapper-xseries-certificates'
+
     @property
     def url(self):
         return self._build_url('dashboard')
 
     def is_browser_on_page(self):
         return self.browser.title.startswith('Dashboard')
+
+    def are_credential_links_present(self):
+        return self.q(css=self.credential_css_selector).present
+
+    def click_credential_link(self):
+        self.q(css=self.credential_css_selector + ' ul li a').click()
+
+    def get_credential_link(self):
+        return self.q(css=self.credential_css_selector + ' ul li a').attrs('href')[0]
--- a/acceptance_tests/test_rendering.py
+++ b/acceptance_tests/test_rendering.py
@@ -1,0 +1,44 @@
+from bok_choy.web_app_test import WebAppTest
+from selenium.webdriver.common.by import By
+from selenium.webdriver.support import expected_conditions as EC
+from selenium.webdriver.support.ui import WebDriverWait
+
+from acceptance_tests.mixins import LoginMixin, CredentialsApiMixin
+from acceptance_tests.pages import LMSDashboardPage
+
+
+class RenderCredentialTests(LoginMixin, WebAppTest, CredentialsApiMixin):
+
+    def setUp(self):
+        super(RenderCredentialTests, self).setUp()
+        self.create_credential()
+        self.change_credential_status('awarded')
+
+    def test_student_dashboard_with_certificate(self):
+        """Verify that links to XSeries credentials are displayed on the student dashboard."""
+        self.login_with_lms()
+        student_dashboard = LMSDashboardPage(self.browser).wait_for_page()
+        student_dashboard.is_browser_on_page()
+
+        self.assertTrue(student_dashboard.are_credential_links_present())
+
+        credential_link = student_dashboard.get_credential_link()
+
+        student_dashboard.click_credential_link()
+        WebDriverWait(self.browser, 10).until(EC.presence_of_element_located((By.ID, 'action-print-view')))
+
+        # check the credential hash-id matches with rendered credential page.
+        self.assertIn(
+            self.browser.find_elements_by_css_selector('.accomplishment-stamp-validity span:nth-of-type(2)')[0].text,
+            credential_link
+        )
+
+    def test_student_dashboard_with_revoked_certificate(self):
+        """Verify that no links to XSeries credentials are displayed on the student dashboard if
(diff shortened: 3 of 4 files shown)


awais786:

ahsan-ul-haq:
:+1:

awais786:
@rlucioni Please review this PR.

acceptance_tests/mixins.py
```diff
@@ -21,3 +25,44 @@ def login_with_lms(self):
 
         self.lms_login_page.browser.get(self.lms_login_page.url())  # pylint: disable=not-callable
         self.lms_login_page.login(email, password)
+
+
+class CredentialsApiMixin(object):
+    """ Mixin used for login on credentials."""
+    def setUp(self):
+        super(CredentialsApiMixin, self).setUp()
+        self.data = ''
```
rlucioni: I'd initialize this to `None` instead of the empty string.

acceptance_tests/mixins.py
```diff
@@ -21,3 +25,44 @@ def login_with_lms(self):
...
+    def _create_credential(self):
+        """Create user credential for a program."""
+
+        return self.credential_api_client.user_credentials.post({
+            'username': LMS_USERNAME,
+            'credential': {'program_id': PROGRAM_ID},
+            'attributes': []
+        })
+
+    def _change_credential_status(self, data, status):
```
rlucioni: Use `self.data` instead of taking it as an argument.

acceptance_tests/mixins.py
```diff
@@ -21,3 +25,44 @@ def login_with_lms(self):
...
+
+
+class CredentialsApiMixin(object):
+    """ Mixin used for login on credentials."""
+    def setUp(self):
+        super(CredentialsApiMixin, self).setUp()
+        self.data = ''
+
+    @property
+    def credential_api_client(self):
```
rlucioni: I'd like to see this client use an OAuth 2.0 access token instead of a JWT. It'll be cleaner. For one, you won't need a signing key nor an OAuth client ID to run the tests.

In order to do this, you'll need to implement and use an authentication class similar to [BearerAuthentication](https://github.com/edx/ecommerce/blob/master/ecommerce/extensions/api/authentication.py#L81) from ecommerce.

acceptance_tests/mixins.py
```diff
@@ -21,3 +25,44 @@ def login_with_lms(self):
...
+        payload = {
+            "iss": OAUTH_URL,
+            "aud": 'credentials-key',
+            "exp": now + datetime.timedelta(seconds=expires_in),
+            "iat": now,
+            "preferred_username": LMS_USERNAME,
+            "administrator": True,
+        }
+        try:
+            jwt_data = jwt.encode(payload, 'credentials-secret')
```
rlucioni: Signing keys and other configuration should not be hardcoded here. This should go away when you switch to using an access token.

acceptance_tests/test_rendering.py
```diff
@@ -0,0 +1,55 @@
...
+from acceptance_tests.mixins import LoginMixin, CredentialsApiMixin
+from acceptance_tests.pages import LMSDashboardPage
+
+
+@skipUnless(ENABLE_OAUTH2_TESTS, 'OAuth2 tests are not enabled.')
+class RenderCredentialTests(LoginMixin, WebAppTest, CredentialsApiMixin):
+
+    def setUp(self):
+        super(RenderCredentialTests, self).setUp()
+        self.browser.maximize_window()
```
rlucioni: Why is this necessary?

acceptance_tests/test_rendering.py
```diff
@@ -0,0 +1,55 @@
...
+        """
+        self.login_with_lms()
+        lms = LMSDashboardPage(self.browser).wait_for_page()
+        lms.is_browser_on_page()
+
+        self.assertTrue(lms.credential_available())
+
+        credential_link = lms.get_credential_link()
+
+        # click the credential link
```
rlucioni: This comment is unnecessary.

acceptance_tests/test_rendering.py
```diff
@@ -0,0 +1,55 @@
...
+from selenium.webdriver.common.by import By
+from selenium.webdriver.support import expected_conditions as EC
+from selenium.webdriver.support.ui import WebDriverWait
+
+from acceptance_tests.config import ENABLE_OAUTH2_TESTS
+from acceptance_tests.mixins import LoginMixin, CredentialsApiMixin
+from acceptance_tests.pages import LMSDashboardPage
+
+
+@skipUnless(ENABLE_OAUTH2_TESTS, 'OAuth2 tests are not enabled.')
```
rlucioni: Don't do this. The acceptance tests which already exist in this repo are decorated like this, but they shouldn't be. It doesn't make sense if you look at what's going on in the config module. If an access token isn't provided, the tests won't run.

acceptance_tests/test_rendering.py
```diff
@@ -0,0 +1,55 @@
...
+        self.data = self._create_credential()
+        self._change_credential_status(self.data, 'awarded')
+
+    def test_student_dashboard_with_certificate(self):
+        """ Render the credentials link from student dashboard.
+        After successful login user will redirect towards the dashboard page. and
+        click on the credential link.
+        """
+        self.login_with_lms()
+        lms = LMSDashboardPage(self.browser).wait_for_page()
```
rlucioni: This would be much easier to read if the variable was named `dashboard` or `student_dashboard`.

acceptance_tests/mixins.py
awais786: @rlucioni can we implement this into another story ?. 
Because It will increase the scope. I have created another PR for this https://github.com/edx/credentials/pull/67

rlucioni: Implementing this in a separate PR is great. Are you planning to finish up #67, merge it, update these tests, then merge this PR?

awais786: I have a plan to merge tests and then merge other PR and then fix tests again. What do u think ?

awais786: @rlucioni I think first merging these test will be a good plan.

awais786:
@rlucioni Please review this PR again. BearerAuth will be implemented in other story. We can merged this PR without BearerAuth.

acceptance_tests/mixins.py
```diff
@@ -21,3 +27,44 @@ def login_with_lms(self):
...
+        }
+        try:
+            jwt_data = jwt.encode(payload, JWT_SECRET_KEY)
+            api_client = EdxRestApiClient(CREDENTIALS_API_URL, jwt=jwt_data)
+        except Exception:  # pylint: disable=broad-except
+            log.exception("Failed to initialize the API client with url '%s'.", CREDENTIALS_API_URL)
+            return
+        return api_client
+
+    def _create_credential(self):
```
rlucioni: This method and the one below aren't really private, are they? I would drop the underscores.

acceptance_tests/mixins.py
```diff
@@ -21,3 +27,44 @@ def login_with_lms(self):
...
+        """Create user credential for a program."""
+
+        return self.credential_api_client.user_credentials.post({
```
rlucioni: Instead of returning data only to update the `data` class attribute later, just update `self.data` here. It'll make your code a little cleaner.

acceptance_tests/test_rendering.py
```diff
@@ -0,0 +1,51 @@
...
+    def setUp(self):
+        super(RenderCredentialTests, self).setUp()
+        self.data = self._create_credential()
+        self._change_credential_status('awarded')
+
+    def test_student_dashboard_with_certificate(self):
+        """ Render the credentials link from student dashboard.
+        After successful login user will redirect towards the dashboard page. and
+        click on the credential link.
+        """
```
rlucioni: I'd clean this up a bit:
```python
"""Verify that links to XSeries credentials are displayed on the student dashboard."""
```

acceptance_tests/pages.py
```diff
@@ -54,9 +54,22 @@ def login(self, username, password):
...
+        self.credential_css_selector = '.wrapper-xseries-certificates'
+
     @property
     def url(self):
         return self._build_url('dashboard')
 
     def is_browser_on_page(self):
         return self.browser.title.startswith('Dashboard')
+
+    def credential_available(self):
```
rlucioni: Give this method a name more consistent with what it does. For example, `are_credential_links_present`.

acceptance_tests/pages.py
```diff
@@ -54,9 +54,22 @@ def login(self, username, password):
...
+        return self.q(css=self.credential_css_selector).present
+
+    def credential_link(self):
```
rlucioni: Same as above. I'd use `click_credential_link`.

acceptance_tests/test_rendering.py
```diff
@@ -0,0 +1,51 @@
...
+        student_dashboard.credential_link()
+        WebDriverWait(self.browser, 10).until(EC.presence_of_element_located((By.ID, 'action-print-view')))
+
+        # check the credential hash-id matches with rendered credential page.
+        self.assertIn(
+            self.browser.find_elements_by_css_selector('.accomplishment-stamp-validity span:nth-of-type(2)')[0].text,
+            credential_link
+        )
+
+
```
rlucioni: Remove this line. Within classes, only one line should separate methods.

acceptance_tests/test_rendering.py
```diff
@@ -0,0 +1,51 @@
...
+    def test_student_dashboard_without_certificate(self):
+        """ If student has no credentials or with revoked status then no link
```
rlucioni: This test doesn't cover the case where the student has no credentials, only the case in which credentials are revoked. Either update the docstring or update the test so that the credential is deleted and the dashboard is re-checked.

awais786:
@rlucioni it's ready for final review.

acceptance_tests/mixins.py
```diff
@@ -1,6 +1,13 @@
...
 import logging
 
-from acceptance_tests.config import LMS_EMAIL, LMS_PASSWORD
+from edx_rest_api_client.client import EdxRestApiClient
+import jwt
+
+from acceptance_tests.config import (
+    LMS_EMAIL, LMS_PASSWORD, CREDENTIALS_API_URL, LMS_USERNAME, PROGRAM_ID, OAUTH_URL,
+    JWT_SECRET_KEY, USER_JWT_AUDIENCE
+)
```
rlucioni: Instead of importing all these constants, just import `config` and access these constants as you need them. It makes it clearer where these constants are coming from.

rlucioni:
:+1: once final comment addressed and commits squashed.

awais786:
Thanks @rlucioni I will do this change tomorrow and than merge it.

PR #69, opened 2016-02-25
Remove prod-requirements Make target

rlucioni:
@zubair-arbi @tasawernawaz @jimabramson FYI.


zubair-arbi:
Change looks good, after rebasing to master :+1:

PR #70, opened 2016-02-26
add optional prod requirement to make prod-requirements command

zubair-arbi:
@awais786 @tasawernawaz @waheedahmed 
Fix the issue where the command `make prod-requirements` was not installing optional requirements.

@rlucioni @clintonb @jimabramson @maxrothman Please let me know your thoughts on this approach.
I believe this way we won't have to merge the changes for [removing make target for prod requirements](https://github.com/edx/credentials/pull/69) and [updating the credentials configuration to install application requirements](https://github.com/edx/configuration/pull/2812/).

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
 # This file is here because many Platforms as a Service look for
 #	requirements.txt in the root directory of a project.
 -r requirements/production.txt
+-r requirements/optional.txt 


clintonb:
Optional requirements should be optional. Today, that option is exercised in Ansible, not in the repos.

zubair-arbi:
@clintonb So Its ok If we don't use a make command in ansible role task to install all prod related requirements?

clintonb:
Yes.

zubair-arbi:
@clintonb Thanks.
I am going to close this PR in favor of https://github.com/edx/credentials/pull/69

PR #72, opened 2016-02-27
Allow app to start when STATICFILES_STORAGE is undefined

rlucioni:
@e0d @maxrothman

FYI @zubair-arbi

### DIFF
--- a/credentials/settings/production.py
+++ b/credentials/settings/production.py
@@ -1,4 +1,5 @@
 from os import environ
+import warnings
 import yaml
 
 from credentials.settings.base import *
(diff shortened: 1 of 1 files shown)


e0d:
Log that it's unavailable?

rlucioni:
@e0d updated. Logging handlers aren't defined at this stage, so I'm using the `warnings` module instead of the root logger.

e0d:
:+1:

maxrothman:
FWIW `STATICFILES_STORAGE` should be defined when using file storage as well.

PR #73, opened 2016-03-01
Workaround for https://github.com/boto/boto/issues/1477

jimabramson:
@zubair-arbi @tasawernawaz @maxrothman PTAL

This could be reverted in the future if we come up with a better workaround.


awais786:
:+1:

PR #74, opened 2016-03-01
Use default JWT_LEEWAY of 10s

jimabramson:
@maxrothman @zubair-arbi @tasawernawaz @rlucioni FYI.

see also https://github.com/edx/edx-platform/pull/11724


jimabramson:
This PR was not necessary.

PR #78, opened 2016-03-07
Adding check only active programs can generate credentials.

awais786:
Adding checks so that only active ProgramCertificate configurations can generate credentials. I have added check for CourseCertificates too.
https://openedx.atlassian.net/browse/ECOM-3858

### DIFF
--- a/credentials/apps/api/tests/test_views.py
+++ b/credentials/apps/api/tests/test_views.py
@@ -313,7 +313,7 @@ class UserCredentialViewSetTests(APITestCase):
             ]
         }
 
-        msg = "Credential ID [{cred_id}] for [ProgramCertificate matching query does not exist.]".format(
+        msg = "Credential ID [{cred_id}] for ProgramCertificate matching query does not exist.".format(
             cred_id=cred_id
         )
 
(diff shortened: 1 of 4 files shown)


ahsan-ul-haq:
Please write test to verify the behavior for inactive course credentials as well.

awais786:
@ahsan-ul-haq courses credentials not implemented yet. So we don't need to write any tests cases for that.

ahsan-ul-haq:
:+1:

awais786:
@clintonb Please review this PR.

credentials/apps/api/tests/test_views.py
```diff
@@ -406,6 +406,33 @@ def _assert_usercredential_fields(self, response, username, expected_attrs):
...
+            },
+            "attributes": [
+                {
+                    "name": self.user_credential_attribute.name,
+                    "value": self.user_credential_attribute.value
+                },
+            ]
+        }
+
+        msg = "Credential ID [{cred_id}] for [ProgramCertificate matching query does not exist.]".format(
```
clintonb: No need for the second set of brackets.

clintonb:
:+1:

PR #80, opened 2016-03-07
Use django template tag for removing the querystrings from media images url

awais786:
@awais786 @tasawernawaz 
Custom template tag `strip_querystrings` is added to strip the querystrings from the media images url.
To enable this functionality add the waffle switch **`strip_image_querystrings`** and mark it as active from django admin panel.

_**Note:** For some unknown reasons `boto` is adding querystring `x-amz-security-token` with the s3 images on stage. At the moment team is trying different solutions to fix that. But for the last resort this PR is created._

Related PR: https://github.com/edx/credentials/pull/81

### DIFF
--- a/credentials/apps/credentials/tests/test_views.py
+++ b/credentials/apps/credentials/tests/test_views.py
@@ -2,12 +2,14 @@
 Tests for credentials rendering views.
 """
 from __future__ import unicode_literals
+import urllib
 import uuid
 
 from django.conf import settings
 from django.core.urlresolvers import reverse
 from django.test import TestCase
 from mock import patch
+from waffle.models import Switch
 
 from credentials.apps.api.tests import factories
 from credentials.apps.credentials.models import Signatory, UserCredential
@@ -23,7 +25,9 @@ class RenderCredentialPageTests(TestCase):
         self.program_certificate = factories.ProgramCertificateFactory.create(template=None)
         self.site = self.program_certificate.site
         self.signatory_1 = Signatory.objects.create(name='Signatory 1', title='Manager', image='images/signatory_1.png')
-        self.signatory_2 = Signatory.objects.create(name='Signatory 1', title='Manager', image='images/signatory_1.png')
+        self.signatory_2 = Signatory.objects.create(
+            name='Signatory 2', title='director', image='images/signatory_2.png?x-amz-security'
+        )
         self.program_certificate.signatories.add(self.signatory_1, self.signatory_2)
         self.user_credential = factories.UserCredentialFactory.create(
             credential=self.program_certificate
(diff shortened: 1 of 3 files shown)


awais786:
@jimabramson , @zubair-arbi I have added the experiment code and added the waffle switch. It is working fine.
[credential](https://credentials-newcerts.sandbox.edx.org/credentials/a1751db2b5ff4508a50b3f0e9651eceb/)

You can search following line in html source.


For enable/disable you can use the admin.
[AdminPanel](https://credentials-newcerts.sandbox.edx.org/admin/)

credentials/apps/credentials/tests/test_views.py
```diff
@@ -178,3 +179,18 @@ def test_get_cert_with_awarded_status_without_signatory(self):
...
+        """
+        __, __ = Switch.objects.get_or_create(name='ENABLE_MANUAL_SIGNATORIES_IMAGES', defaults={'active': True})
+
+        response = self._render_user_credential()
+
+        self.assertEqual(response.status_code, 200)
+        self._assert_user_credential_template_data(response, self.user_credential, certificate_title='Test Program A')
+        self._assert_signatory_data(response, self.signatory_1)
+        self._assert_signatory_data(response, self.signatory_2)
+        self.assertContains(response, '<!--waffle switch is active-->')
```
jimabramson: I think this test should be checking that querystrings are actually being removed.

credentials/apps/credentials/views.py
```diff
@@ -32,9 +33,23 @@ def get_context_data(self, **kwargs):
         if user_credential.credential_content_type.model_class() != ProgramCertificate:
             raise Http404
 
+        signatories = []
+        if waffle.switch_is_active('ENABLE_MANUAL_SIGNATORIES_IMAGES'):
```
jimabramson: why not name the switch something more obvious, like 'strip image querystrings' ?

credentials/templates/credentials/program_certificate.html
```diff
@@ -106,19 +106,36 @@ <h3 class="accomplishment-signatories-title sr-only">{% trans "Noted by" %}</h3>
 
                             <div class="wrapper-signatories">
                                 <div class="list-signatories">
+                                 {% if enable_manual_images %}
```
jimabramson: The duplication of template markup doesn't seem good.  Everything is the same except for the specific attributes you're accessing in a handful of places.  Seems like it would be better to always convert the signatories to the custom dictionary, but only strip querystrings when the switch is active, then you don't need to repeat all the template code.

credentials/apps/credentials/views.py
```diff
@@ -35,6 +36,7 @@ def get_context_data(self, **kwargs):
         context.update({
             'user_credential': user_credential,
             'certificate_context': self.get_program_certificate_context(user_credential),
+            'enable_manual_images': waffle.switch_is_active('strip_image_querystrings')
```
zubair-arbi: no need for passing it from view, you can do this directly in the template: https://waffle.readthedocs.org/en/v0.8.1/usage.html#using-waffle-in-django-templates

credentials/templates/credentials/program_certificate.html
```diff
@@ -109,7 +110,12 @@ <h3 class="accomplishment-signatories-title sr-only">{% trans "Noted by" %}</h3>
                                     {% for signatory in user_credential.credential.signatories.all %}
                                     <div class="signatory">
                                         {% if signatory.image %}
-                                            <img class="signatory-signature" src="{{signatory.image.url}}" alt="{{signatory.name}}">
+                                            {% if enable_manual_images %}
+                                                <!--waffle switch is active-->
```
zubair-arbi: remove this comment

zubair-arbi:
changes looks good. :+1: 
But lets discuss with @jimabramson and @maxrothman first before merging this.

PR #82, opened 2016-03-08
remove s3 querystrings in the storage adapter

jimabramson:
@zubair-arbi @clintonb please have a look


clintonb:
Tests?

clintonb:
:+1: pending:

- [ ] tests (unless you are confident they aren't needed)
- [ ] squashed commits

jimabramson:
@maxrothman and I figured out the root cause, so this won't be necessary.

PR #81, opened 2016-03-08
re-revert boto fix by jim

zubair-arbi:
@awais786 @tasawernawaz @waheedahmed 
I have verified on my sandbox that @jimabramson fix is working fine for more than 1 hour so my understanding is that this fix is working and we will check for how long it works for future reference if it stops working again.

So going forward I say we should merge this PR and for workaround (just in case) we will also merge the @awais786 [PR for disabling aws querystrings for media images](https://github.com/edx/credentials/pull/80) with a waffle switch 'strip_image_querystrings'.

Related PR's:
- https://github.com/edx/credentials/pull/73 (override boto class `S3Connection` for empty `security_token`)
- https://github.com/edx/credentials/pull/76 (Use leonsim's fork for boto for fixing aws querysting issue)
- https://github.com/edx/credentials/pull/79 (reword pip requirement for leonsim's boto fork)
- https://github.com/edx/credentials/pull/80 (django template tag to remove querystrings from media images url)

PR #79, opened 2016-03-07
reword pip requirement for boto fork

jimabramson:
@maxrothman @zubair-arbi 

the prior phrasing seemed to cause pip to fall back to installing the latest release version, instead of the commit.

PR #83, opened 2016-03-08
Revert "use leonsim's fork of boto for fixing aws querysting issue"

jimabramson:
This reverts commit d60176274186825ed9d71ecdf8697681f1932946.

The workaround we needed was in configuration all along, but had the wrong name (fixed in https://github.com/edx/configuration/pull/2840)


jimabramson:
@zubair-arbi @awais786 I think we can get rid of #79, #80, and #81 also - once we check https://github.com/edx/configuration/pull/2840 and make sure it does the trick.

rlucioni:
:+1:

zubair-arbi:
Great :+1:

PR #87, opened 2016-03-15
Avoid setting COMPRESS_STORAGE equal to STATICFILES_STORAGE

rlucioni:
In some cases, such as when using Django's ManifestStaticFilesStorage, these settings must be different.

@jimabramson @maxrothman @zubair-arbi @awais786


jimabramson:
:+1: but let's see if it works before we merge

rlucioni:
For context: https://github.com/django-compressor/django-compressor/issues/746

PR #92, opened 2016-03-31
Add signatory organization info

ahsan-ul-haq:
ECOM-4009

@awais786 @tasawernawaz please review.

### DIFF
--- a/credentials/apps/credentials/models.py
+++ b/credentials/apps/credentials/models.py
@@ -88,6 +88,10 @@ class Signatory(TimeStampedModel):
     """
     name = models.CharField(max_length=255)
     title = models.CharField(max_length=255)
+    organization_name_override = models.CharField(
+        max_length=255, null=True, blank=True,
+        help_text=_('Signatory organization name if its different from issuing organization.')
+    )
     image = models.ImageField(
         help_text=_(
             'Image must be square PNG files. The file size should be under 250KB.'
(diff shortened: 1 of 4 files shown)


awais786:
Please add a test case and check the value in response.

awais786:
:+1:

ahsan-ul-haq:
@jimabramson @clintonb please review.

jimabramson:
@ahsan-ul-haq :+1:  to these changes.  Optional (not part of this change) could you take a quick look at the tests using `self._assert_signatory_data` - these seem like they might be misleading because in the `setUp`, `signatory_1` and `signatory_2` have the same values.

credentials/apps/credentials/models.py
```diff
@@ -88,6 +88,10 @@ class Signatory(TimeStampedModel):
     """
     name = models.CharField(max_length=255)
     title = models.CharField(max_length=255)
+    organization_name = models.CharField(
```
clintonb: I would call this `organization_name_override` to be completely explicit about the fact that this is not generally the normal use case.

credentials/templates/credentials/program_certificate.html
```diff
@@ -115,7 +115,11 @@ <h4 class="signatory-name">{{ signatory.name }}</h4>
 
                                         <p class="signatory-credentials">
                                             <span class="role">{{ signatory.title }}</span>
-                                            <span class="organization">{{ certificate_context.organization_data.name }}</span>
+                                            {% if signatory.organization_name %}
```
clintonb: Use [`firstof`](https://docs.djangoproject.com/en/1.9/ref/templates/builtins/#firstof).

clintonb:
:+1: once comments addressed

PR #93, opened 2016-03-31
Fixing print issue.

awais786:
ECOM-3781
Fixing print preview back-ground image issue.

### DIFF
--- a/credentials/static/sass/_views.scss
+++ b/credentials/static/sass/_views.scss
@@ -78,10 +78,11 @@
 
   .accomplishment-rendering {
     background-color: palette(grayscale, white-t);
-    background-image: url("#{$accomplishment-xseries-path}/xseries-seal-grayscale.png");
     background-repeat: no-repeat;
     background-position: center 30px;
     background-size: 80% auto;
+    overflow: hidden;
+    position: relative;
 
       @include susy-breakpoint($bp-screen-md, $susy) {
         background-position: center 40px;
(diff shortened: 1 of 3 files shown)


awais786:
@marcotuts Please check this fix.

marcotuts:
I don't believe this addresses the root problem. Is there a second"-darker" image uploaded to s3 in the correct location? 

This proposed fix uses only the image that has been working 

That should be the fix instead of additional styling changes. The existing styles should handle watermark placement as expected, including rtl and responsive use cases.

awais786:
@marcotuts 2nd image is also available.
[darker-image](https://github.com/edx/credentials/blob/master/credentials/static/images/xseries-seal-grayscale-darker.png)

And I have mentioned in the ticket image appear on print-preview also. But user need to click the print button and then cancel button twice. On third attempt image appears.

marcotuts:
Discussed offline, reminded of intermittent nature of the bug. Thumbs up with the current approach, jus need a biapp wrapper around that left rule.

marcotuts:
:thumbs:

credentials/static/sass/_views.scss
```diff
@@ -92,6 +92,14 @@
         background-position: -325px -50px;
         background-size: 100% auto;
       }
+    .background-image{
+      position: absolute;
+      z-index: 0;
+      left: -325px;
```
marcotuts: @include left(-325px);

ahsan-ul-haq:
:+1:

PR #94, opened 2016-04-01
Fix model permissions for UserCredential.

jimabramson:
ECOM-3978

@ahsan-ul-haq @awais786 @rlucioni  please review.

### DIFF
--- a/credentials/apps/api/permissions.py
+++ b/credentials/apps/api/permissions.py
@@ -1,0 +1,55 @@
+"""
+Custom permissions classes for use with DRF.
+"""
+from django.http import Http404
+from rest_framework import permissions
+
+
+class UserCredentialViewSetPermissions(permissions.DjangoModelPermissions):
+    """
+    Custom extension to DjangoModelPermissions for use with the
+    UserCredentialViewSet.
+
+    This permissions class uses an explicit 'view' permission, enabling support
+    for:
+        - explicitly granting certain users read access to any UserCredential
+        - implicitly granting any user read access to UserCredentials that were
+          awarded specifically to them
+        - denying read access (obscured by HTTP 404) in any other case
+
+    NOTE: users are allowed to read their own UserCredential records regardless
+        of their 'status' (i.e. even if revoked).
+
+    WARNING: this permissions implementation does not cover the 'list' method
+        of access.  The access control required under DRF for that use case is
+        presently implemented in the `list` method of the viewset itself.
+    """
+
+    # refer to the super() for more context on what this override is doing.
+    perms_map = permissions.DjangoModelPermissions.perms_map
+    perms_map.update({method: ['%(app_label)s.view_%(model_name)s'] for method in permissions.SAFE_METHODS})
+
+    def has_permission(self, request, view):
+        """
+        Relax the base's view-level permissions in the case of 'safe'
+        (read-only) methods, requiring only that the user be authenticated.
+
+        This lets us delay deciding whether or not read permission should be
+        implicitly granted, until after DRF has fetched the requested object.
+        """
+        return super(UserCredentialViewSetPermissions, self).has_permission(request, view) or (
+            request.user.is_authenticated() and request.method in permissions.SAFE_METHODS
+        )
+
+    def has_object_permission(self, request, view, obj):
+        """
+        Allow access to specific objects when granted explicitly (via model
+        permissions) or, if a read-only request, implicitly (via matching
+        username).
+        """
+        if super(UserCredentialViewSetPermissions, self).has_permission(request, view):
+            return True
+        elif request.user.username.lower() == obj.username.lower():
+            return True
+        else:
+            raise Http404
--- a/credentials/apps/api/tests/test_views.py
+++ b/credentials/apps/api/tests/test_views.py
@@ -5,13 +5,14 @@ from __future__ import unicode_literals
 import json
 
 import ddt
-from django.contrib.auth.models import Permission
+from django.contrib.auth.models import Group, Permission
 from django.core.urlresolvers import reverse
 from rest_framework.test import APITestCase, APIRequestFactory
 from testfixtures import LogCapture
 
 from credentials.apps.api.serializers import UserCredentialSerializer
 from credentials.apps.api.tests import factories
+from credentials.apps.core.constants import Role
 from credentials.apps.credentials.models import UserCredential
 
 
@@ -40,6 +41,11 @@ class UserCredentialViewSetTests(APITestCase):
         self.username = "test_user"
         self.request = APIRequestFactory().get('/')
 
+    def _add_permission(self, perm):
+        """ DRY helper to add usercredential model permissions to self.user """
+        # pylint: disable=no-member
+        self.user.user_permissions.add(Permission.objects.get(codename='{}_usercredential'.format(perm)))
+
     def _attempt_update_user_credential(self, data):
         """ Helper method that attempts to patch an existing credential object.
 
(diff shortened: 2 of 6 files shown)


credentials/apps/api/tests/test_views.py
```diff
@@ -441,6 +449,122 @@ def test_users_lists_access_by_authenticated_users(self):
...
+        ({}, 403),
+        ({'username': 'test-user'}, 403),
+    )
+    @ddt.unpack
+    def test_create(self, user_kwargs, expected_status):
+        """
+        The creation (POST) method requires the 'add' permission.
+        """
+        list_path = reverse('api:v1:usercredential-list')
+        program_certificate = factories.ProgramCertificateFactory(program_id=2)
```
rlucioni: Is it necessary to provide the `program_id` kwarg here?

credentials/apps/api/tests/test_views.py
```diff
@@ -441,6 +449,122 @@ def test_users_lists_access_by_authenticated_users(self):
...
+            'username': 'test-user',
+            'credential': {
+                'program_id': program_certificate.program_id
+            },
+            'attributes': [],
+        }
+
+        self.client.force_authenticate(self.make_user(**user_kwargs))  # pylint: disable=no-member
+        response = self.client.post(list_path, data=json.dumps(post_data), content_type=JSON_CONTENT_TYPE)
+        self.assertEqual(response.status_code, expected_status)
```
rlucioni: Would it also make sense to verify that the credential was (or was not) created?

credentials/apps/core/migrations/0003_auto_20160331_0218.py
```diff
@@ -0,0 +1,44 @@
...
+from django.contrib.contenttypes.models import ContentType
+from django.core.exceptions import ObjectDoesNotExist
+from django.db import migrations
+
+from credentials.apps.core.constants import Role
+
+
+def create_view_permission(apps, schema_editor):
+    """
+    """
```
rlucioni: Could you fill in this docstring and the one below (or remove them both)?

credentials/apps/api/tests/test_views.py
```diff
@@ -441,6 +449,122 @@ def test_users_lists_access_by_authenticated_users(self):
...
+
+    @ddt.data(
+        ({'group': Role.ADMINS}, 200),
+        ({'perm': 'view'}, 200),
+        ({'perm': 'add'}, 404),
+        ({'perm': 'change'}, 404),
+        ({'username': 'test-user'}, 200),
+        ({'username': 'TeSt-uSeR'}, 200),
```
rlucioni: Aren't Django's usernames are case-sensitive? Is it really correct to grant both `test-user` and `TeSt-uSeR` access to `test-user`'s objects?

credentials/apps/api/tests/test_views.py
```diff
@@ -441,6 +449,122 @@ def test_users_lists_access_by_authenticated_users(self):
...
+        """
+        program_cert = factories.ProgramCertificateFactory()
+        user_credential = factories.UserCredentialFactory.create(credential=program_cert, username='test-user')
+        detail_path = reverse("api:v1:usercredential-detail", args=[user_credential.id])
+        post_data = {
+            'username': 'test-user',
+            'credential': {
+                'program_id': program_cert.program_id
+            },
+            'attributes': [{"name": "dummy-attr-name", "value": "dummy-attr-value"}],
```
rlucioni: Nitpick: please stick to one quote style.

credentials/apps/api/permissions.py
```diff
@@ -0,0 +1,55 @@
...
+    def has_permission(self, request, view):
+        """
+        Relax the base's view-level permissions in the case of 'safe'
+        (read-only) methods, requiring only that the user be authenticated.
+
+        This lets us delay deciding whether or not read permission should be
+        implicitly granted, until after DRF has fetched the requested object.
+        """
+        return super(UserCredentialViewSetPermissions, self).has_permission(request, view) or (
+            request.user.is_authenticated() and request.method in permissions.SAFE_METHODS
```
rlucioni: I'm having a hard time understanding why this line is necessary and if it's correct. If [`super()`](https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/permissions.py#L110-L132) with an extended `perms_map` fails to grant the user permission, why perform this similar but more relaxed check? Given [line 50](https://github.com/edx/credentials/pull/94/files#diff-508c2850acfb8f9922ce96497ac7802eR50), it seems like this could end up granting a user permission to an object without checking usernames first.

rlucioni: I read http://www.django-rest-framework.org/api-guide/permissions/#custom-permissions but I'm still confused. According to the docs, "the instance-level `has_object_permission` method will only be called if the view-level `has_permission` checks have already passed." Here, `has_object_permission()` is calling the view-level `has_permission()` again, instead of assuming that it has passed.

credentials/apps/api/v1/views.py
```diff
@@ -19,16 +20,22 @@ class UserCredentialViewSet(viewsets.ModelViewSet):
...
     def list(self, request, *args, **kwargs):
-        if not self.request.query_params.get('username'):
+        if not request.query_params.get('username'):
             raise ValidationError(
                 {'error': 'A username query string parameter is required for filtering user credentials.'})
 
+        # provide an additional permission check related to the username
+        # query string parameter.  See also `UserCredentialViewSetPermissions`
+        if not request.user.has_perm('credentials.view_usercredential') and (
+                request.user.username.lower() != self.request.query_params['username'].lower()
```
schenedx: Why did you decide to make username case insensitive for permission?

jimabramson:
@tasawernawaz

credentials/apps/api/tests/test_views.py
jimabramson: it seems not - will remove

credentials/apps/api/tests/test_views.py
jimabramson: i'd prefer to leave this test strictly focused on permissions / status codes as I think we can leave verifying transaction results to the other test case.  If you feel strongly about this I'm willing to reconsider.

credentials/apps/api/tests/test_views.py
jimabramson: oops!  fixing

credentials/apps/core/migrations/0003_auto_20160331_0218.py
jimabramson: fixed

credentials/apps/api/tests/test_views.py
jimabramson: in fact they are case-insensitive.

credentials/apps/api/v1/views.py
jimabramson: usernames are case-insensitive in our system.

credentials/apps/api/permissions.py
jimabramson: This is the first of two permission checks.  It allows any authenticated user in case of a read request, or any user with explicit permission to perform the write operation associated with the requested view.  Unauthenticated users, or users attempting to make write requests without permission, will be denied access in the default way (i.e. returning `False`)

The second check, below, calls the super's view-level permission method (not its own).  It allows any kind of request to proceed if the user has explicit permission to perform it; anyone else making a request has only been allowed to get this far so that we can compare their username to the username associated with the data.

I couldn't come up with a more straightforward way to implement this in DRF's framework, but of course, that doesn't mean there isn't one...

jimabramson: It may also help to note that because this PR adds an explicit 'view' permission on `UserCredential`, no user would be able to make a read request unless they had that permission.  This is why we need to make an exception for non-privileged users in `has_permission`, and then enforce the username check in `has_object_permission`

rlucioni: Understood, thanks for the explanation.

rlucioni:
👍

awais786:
:+1:

credentials/apps/api/v1/views.py
schenedx: Why leave the `self` in the `request` reference in the second part of this comparison?

schenedx:
Besides a question, rest LGTM

credentials/apps/api/v1/views.py
jimabramson: unintentional - thanks for catching, will fix

PR #98, opened 2016-05-02
Added ability to override issuing organization for credentials.

waheedahmed:
[ECOM-4198] (https://openedx.atlassian.net/browse/ECOM-4198)

In order to support the existing use of wrapper-organizations such as VJx, added the ability to override the issuing organization on credentials (specifically XSeries, as courses already have this.)

### DIFF
--- a/credentials/apps/credentials/models.py
+++ b/credentials/apps/credentials/models.py
@@ -233,6 +233,12 @@ class ProgramCertificate(AbstractCertificate):
         object_id_field='credential_id',
         related_query_name='program_credentials'
     )
+    use_org_name = models.BooleanField(
+        default=False,
+        help_text=_("Display the associated organization's name (e.g. ACME University) "
+                    "instead of its short name (e.g. ACMEx)"),
+        verbose_name=_('Use organization name')
+    )
 
     class Meta(object):
         verbose_name = "Program certificate configuration"
--- a/credentials/apps/credentials/tests/test_views.py
+++ b/credentials/apps/credentials/tests/test_views.py
@@ -189,3 +189,22 @@ class RenderCredentialPageTests(TestCase):
         self.assertEqual(response.status_code, 200)
         self.assertContains(response, self.signatory_1.organization_name_override)
         self.assertNotContains(response, self.signatory_2.organization_name_override)
+
+    def test_issuing_organization_name_override(self):
+        """ Verify that the view response contains organization 'name' instead of 'short_name'
+        if 'use_org_name' is True."""
+        response = self._render_user_credential()
+
+        self.assertEqual(
+            response.context['certificate_context']['organization_name'],
+            OrganizationsDataMixin.ORGANIZATIONS_API_RESPONSE['short_name']
+        )
+
+        self.program_certificate.use_org_name = True
+        self.program_certificate.save()
+        response = self._render_user_credential()
+
+        self.assertEqual(
+            response.context['certificate_context']['organization_name'],
+            OrganizationsDataMixin.ORGANIZATIONS_API_RESPONSE['name']
+        )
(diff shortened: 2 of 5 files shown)


waheedahmed:
@zubair-arbi @tasawernawaz @ahsan-ul-haq @awais786 Please review.

credentials/templates/credentials/program_certificate.html
```diff
@@ -96,7 +96,7 @@ <h4 class="organization-name">{{ certificate_context.organization_data.name }}</
 
                                 <span class="accomplishment-statement-detail copy">
                                     {# Translators: course_count is an integer value, and name is the display name for the provided organization e.g Test Organization. #}
-                                    {% blocktrans with course_count=certificate_context.programs_data.course_count name=certificate_context.organization_data.short_name platform_name=platform_name %}
+                                    {% blocktrans with course_count=certificate_context.programs_data.course_count name=certificate_context.programs_data.organization_name platform_name=platform_name %}
```
ahsan-ul-haq: In case when we do not want to use overdid organization name we would need `certificate_context.organization_data.short_name` please use something like https://github.com/edx/credentials/blob/waheed/ecom-4198-override-issuing-organization/credentials/templates/credentials/program_certificate.html#L119

ahsan-ul-haq: sorry for confusion got that.

waheedahmed:
@clintonb Please review.

ahsan-ul-haq:
LGTM 👍

clintonb:
I do not approve. The solution is valid, but I don't think this is the best solution. We should not be storing organization-related data on credentials models. The Organizations API already returns the exact data you need: https://courses.stage.edx.org/api/organizations/v0/organizations/vjx/ in the name field.

Instead of storing this data, you need only store a boolean (e.g. `use_org_name`) to toggle the usage of `short_name` over `name`. This will be much cleaner, and avoid spreading data around unnecessarily. Ultimately we want to consolidate all of our course/program metadata.

waheedahmed:
@clintonb Please take another look.

credentials/apps/credentials/models.py
```diff
@@ -233,6 +233,7 @@ class ProgramCertificate(AbstractCertificate):
         object_id_field='credential_id',
         related_query_name='program_credentials'
     )
+    use_org_name = models.BooleanField(default=False)
```
clintonb: Add help text explaining the purpose of this field.

credentials/apps/credentials/tests/test_views.py
```diff
@@ -189,3 +189,15 @@ def test_signatory_organization_name_override(self):
         self.assertEqual(response.status_code, 200)
         self.assertContains(response, self.signatory_1.organization_name_override)
         self.assertNotContains(response, self.signatory_2.organization_name_override)
+
+    def test_issuing_organization_name_override(self):
+        """ Verify that the view response contains organization 'name' instead of 'short_name'
+        if 'use_org_name' is True."""
+        response = self._render_user_credential()
+        self.assertContains(response, OrganizationsDataMixin.ORGANIZATIONS_API_RESPONSE['short_name'])
```
clintonb: Check `response.context`.

waheedahmed:
@clintonb Does it looks good now?

credentials/apps/credentials/models.py
```diff
@@ -233,6 +233,11 @@ class ProgramCertificate(AbstractCertificate):
         object_id_field='credential_id',
         related_query_name='program_credentials'
     )
+    use_org_name = models.BooleanField(
+        default=False,
+        help_text='Display organization name instead of its short name.',
```
clintonb: Mark for translation.

`Display the associated organization's name (e.g. ACME University) instead of its short name (e.g. ACMEx)`

credentials/apps/credentials/models.py
```diff
@@ -233,6 +233,11 @@ class ProgramCertificate(AbstractCertificate):
...
+        verbose_name='Use organization name'
```
clintonb: Mark for translation.

clintonb:
👍

PR #99, opened 2016-05-03
bump edx-auth-backends and remove python-social-auth pinning

zubair-arbi:
@waheedahmed @ahsan-ul-haq @tasawernawaz @awais786 

* Bump `edx-auth-backends` from version `0.2.1` to `0.2.3`
* Remove `python-social-auth` pin at version `0.2.14` due to the latest version `0.2.19`

FYI: @clintonb @rlucioni


ahsan-ul-haq:
Once build passes 👍

awais786:
👍

PR #101, opened 2016-05-09
Move Credentials service docs to edx-documentation repo

srpearce:
@rlucioni I'm finally getting around to moving all of the docs for individual IDAs into the edx-documentation repo. In this PR, the readme.rst file explains that the documentation for the Credentials service is moving to the edx-documentation repo and that anyone who wants to update those docs should add commits to https://github.com/edx/edx-documentation/pull/922. This PR also moves all the current files in the docs folder into an "old docs - do not use" folder, just in case we end up needing them (we shouldn't).

I want to add similar readme files to the other IDA doc folders (e.g., ecommerce) and to cookiecutter-django-ida/{{cookiecutter.repo_name}}/docs/. Can you help me determine which IDAs have docs that this readme should replace?


rlucioni:
@srpearce, @clintonb is having second thoughts about centralizing IDA docs. FWIW, I'd prefer docs live closest to the code they describe, but I can see the benefit of hosting the docs together. I'd like us all to be on the same page before we decide to move more docs.

clintonb:
Closing due to lack of movement. We can re-open after discussing the home of IDA docs.

PR #106, opened 2016-06-22
Added single signout

ahsan-ul-haq:
@clintonb Please review


clintonb:
👍

awais786:
@ahsan-ul-haq you have to change in the configuration repo too.  Since you have removed the accounts from urls.
[credentials_logout_url](https://github.com/edx/configuration/pull/3123/files#diff-240da08108c4d3cce14b2abd689b924eR51)

FYI @clintonb

credentials/apps/api/tests/test_views.py
```diff
@@ -66,6 +66,7 @@ def test_get(self):
         path = reverse("api:v1:usercredential-detail", args=[self.user_credential.id])
         response = self.client.get(path)
         self.assertEqual(response.status_code, 200)
+        # pylint: disable=no-member
```
clintonb: Write this once at the top of the file.

clintonb:
@awais786 nice catch!

PR #112, opened 2016-08-26
Changed requirements to pull edx-i18n-tools from PyPI and .gitignore …

Dillon-Dumesnil:
…to include en files

@clintonb @bderusha Please review


requirements/local.txt
```diff
@@ -6,7 +6,7 @@ django-debug-toolbar==1.4
 
 # i18n
 transifex-client==0.11
-git+https://github.com/edx/i18n-tools.git@v0.1.4#egg=i18n_tools==0.1.4
+edx-i18n-tools>=0.3.3
```
nedbat: I'm making the same comment in all the repos, sorry: don't we want to pin exact versions for IDA's?

iivic:
@clintonb @mjfrey Please review.

clintonb:
Include the English .po files.

iivic:
Review once more.

clintonb:
👍 (after .mo file added and commits squashed)

PR #118, opened 2016-11-07
Update New Relic Agent to avoid bug

mikedikan:
ECOM-6316

@rlucioni for review
FYI @edx/ecommerce


mikedikan:
Ignoring VersionEye per @clintonb

PR #120, opened 2016-11-16
Update Transifex config to avoid searching for nonexistent PO file

rlucioni:
Credentials has no djangojs.po.

@clintonb please review. @edx/ecommerce FYI.


rlucioni:
In the same vein as https://github.com/edx/programs/pull/172.

PR #124, opened 2016-11-29
Added UUID field to ProgramCertificate model

clintonb:
ECOM-6481


clintonb:
@edx/ecommerce please review

amangano-edx:
This seems to include a lot of extra stuff other than just the UUID change (translations files, etc). Would it make sense to break out some of this other stuff into separate PRs?

clintonb:
@amangano-edx the cleanup work is done in a separate commit. I'm not convinced a separate PR would add much value; but, if you feel strongly, I can create one.

credentials/apps/credentials/tests/test_models.py
```diff
@@ -143,3 +143,14 @@ def test_invalid_course_key(self):
             ).full_clean()
 
         self.assertEqual(context.exception.message_dict, {'course_id': ['Invalid course key.']})
+
+
+class ProgramCertificateTests(TestCase):
+    """ Tests for the ProgramCertificate model. """
+
+    def test_str(self):
```
amangano-edx: Would it make sense to add a docstring here to clarify what `str` is expected to return?

credentials/conf/locale/eo/LC_MESSAGES/django.po
```diff
@@ -0,0 +1,319 @@
+# SOME DESCRIPTIVE TITLE.
+# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
+# This file is distributed under the same license as the PACKAGE package.
+# FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
```
amangano-edx: Does this comment need to be updated?

credentials/conf/locale/eo/LC_MESSAGES/django.po
```diff
@@ -0,0 +1,319 @@
...
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: PACKAGE VERSION\n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2016-11-29 07:22+0000\n"
+"PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
+"Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
+"Language-Team: LANGUAGE <LL@li.org>\n"
```
amangano-edx: Does any of this need to be updated?

amangano-edx:
@clintonb I don't feel strongly about it here, but in general I've found it helpful to keep PRs as small and focused as possible to minimize the risk of regressions slipping through and to make it easier for reviewers to catch them.

credentials/apps/credentials/tests/test_models.py
clintonb: I don't think so. [`__str__()`](https://docs.python.org/2/reference/datamodel.html#object.__str__) is a pretty common built-in method.

PR #125, opened 2016-11-30
Retrieving program details from the Catalog API

clintonb:
Program details are now retrieved from the Catalog API when a program's UUID is known. If no UUID is available, we fall back to the Programs API.

ECOM-6483


credentials/apps/credentials/models.py
```diff
@@ -223,10 +231,14 @@ class Meta(object):
         verbose_name = "Course certificate configuration"
 
 
+ProgramDetails = namedtuple('ProgramDetails', ('uuid', 'title', 'type', 'organization_keys', 'course_count',))
```
schenedx: Is this a module level variable? Is this a constant? Shouldn't constants be capitalized?

PR #126, opened 2016-11-30
Updated front-end configuration

clintonb:
We load about 65KB of JavaScript simply to create an event handler for the print button. That's ridiculous. All of the JS pipeline, and most of the third-party libraries, have been removed from this repository. That overhead is unnecessary.

Additionally, I have added an endpoint to render an example certificate so that it is easy to verify/preview these changes in the future.


clintonb:
@edx/ecommerce please review

credentials/templates/credentials/program_certificate.html
```diff
@@ -38,7 +39,7 @@ <h2 class="message-title">
                     <div class="message-actions">
                         <div class="message-actions-box">
                             <h3 class="sr-only">{% trans "Print or share your certificate:" %}</h3>
-                            <button title="{% trans 'Print' %}" class="action action-print btn icon-only" id="action-print-view">
+                            <button title="{% trans 'Print' %}" class="action action-print btn icon-only" onclick="window.print();">
```
AlasdairSwan: What is this, Angular? Inline functions grumble, grumble, grumble. But is this is the only JS in the whole IDA I guess it's fair enough. @clintonb do you know the long term vision for this repo? i.e. can we be sure that there won't be future JS requirements beyond a print button?

clintonb: If we do add social sharing buttons—LinkedIn, Twitter, Facebook—we may want to track those. I don't foresee this service becoming very JavaScript-heavy. When we get to the point of adding the buttons, we can decide if there are any merits to the overhead of a JS pipeline (whether it be RequireJS, Webpack, etc.). For now, the overhead is overkill.

AlasdairSwan: I agree that the current overhead is overkill, I'm just debating whether we should have a separate JS file for the click handler that we could easily extend instead of ripping out all JS completely.

clintonb: If/when we get to the point of needing to extend, we should add a file. For now, I'm not convinced we even need a separate `<script>` tag for the print button.

PR #129, opened 2016-12-02
Added mgmt command to add UUIDs to program certificates table.

davec-edx:
ECOM-6485

@edx/ecommerce 

Please review
thanks

### DIFF
--- a/credentials/apps/credentials/management/commands/add_uuids_to_program_certificates.py
+++ b/credentials/apps/credentials/management/commands/add_uuids_to_program_certificates.py
@@ -1,0 +1,50 @@
+import logging
+
+from django.core.management.base import BaseCommand
+from slumber.exceptions import HttpNotFoundError
+
+from credentials.apps.credentials.models import ProgramCertificate
+from credentials.apps.credentials.utils import get_program
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    """
+    This management comand is essentially a one-off.  We are adding this as an extra step to
+    populate the newly added UUID field in the ProgramCertificate table.  After it has been
+    run to do the initial population, it will not need to be run again.
+    """
+    help = 'Add missing UUIDs to the programCertificate table'
+
+    def get_uuid_from_api(self, program_id):
+        """ Return the UUID for the provided program ID """
+
+        # If there is no ID (meaning the program_id is empty), that is a problem,
+        # so report it, this likely means there is an issue with the Db
+        if program_id is None:
+            raise AttributeError('Need an ID')
+
+        # Now, get the program id.
+        item = get_program(program_id)
+
+        return item['uuid']
+
+    def handle(self, *args, **options):
+        """ Add missing UUIDs to the programCertificate table"""
+
+        program_certificate_table = ProgramCertificate.objects.filter(program_uuid__isnull=True)
+
+        for program_entry in program_certificate_table:
+            try:
+                logger.info('Getting UUID for %s', program_entry.program_id)
+                program_entry.program_uuid = self.get_uuid_from_api(program_entry.program_id)
+
+                program_entry.save()
+            except HttpNotFoundError:
+                logger.warning('ID not found: %s', program_entry.program_id)
+            except AttributeError:
+                logger.warning('Bad ID: %s', program_entry.program_id, exc_info=True)
+            except Exception:  # pylint: disable=broad-except
+                logger.exception('Error detected trying to get the ID (%s) from the program service',
+                                 program_entry.program_id)


MatthewPiatetsky:
you have some pylint errors, if you're not familiar with how our travis works, click on details on the failed build and scroll to the bottom to see the error

MatthewPiatetsky:
Could you add a brief docstring explaining at a high level what this command is doing/why?

credentials/apps/credentials/management/commands/add_uuids_to_program_certificates.py
```diff
@@ -0,0 +1,34 @@
...
+        program_certificate_table = ProgramCertificate.objects.filter(program_uuid__isnull=True)
+
+        for program_entry in program_certificate_table:
+            try:
+                logger.info('Getting UUID for {}'.format(program_entry.program_id))
+                program_entry.program_uuid = self.get_uuid_from_api(program_entry.program_id)
+
+                program_entry.save()
+            except AttributeError:
+                pass
```
MatthewPiatetsky: why pass in this case?

credentials/apps/credentials/management/commands/add_uuids_to_program_certificates.py
```diff
@@ -0,0 +1,34 @@
...
+    help = 'Add missing UUIDs to the programCertificate table'
+
+    def get_uuid_from_api(self, id=None):
+        if id is None:
+            raise AttributeError('Need an ID')
+
+        item = get_program(id)
+        return item['uuid']
+
+    def handle(self, *args, **options):
```
amangano-edx: I don't have much context on this, but is `handle` a special name that something else is expecting to call? If not, could we name this something more descriptive, like `add_missing_uuids` or something?

credentials/apps/credentials/management/commands/add_uuids_to_program_certificates.py
amangano-edx: When is this raised? Does it make sense to log when this happens so that we know something went wrong?

credentials/apps/credentials/management/commands/add_uuids_to_program_certificates.py
```diff
@@ -0,0 +1,34 @@
...
+        for program_entry in program_certificate_table:
+            try:
+                logger.info('Getting UUID for {}'.format(program_entry.program_id))
+                program_entry.program_uuid = self.get_uuid_from_api(program_entry.program_id)
+
+                program_entry.save()
+            except AttributeError:
+                pass
+            except Exception:
+                logger.exception('Error detected trying to get the ID from the program service')
```
amangano-edx: Should we log some more context so that we know what happened?
```
except Exception as e:
    logger.exception('Error detected trying to get the ID from the program service. Message: {}'.format(e.message))

File: credentials/apps/credentials/management/commands/add_uuids_to_program_certificates.py [hunk 4]
```diff
@@ -0,0 +1,34 @@
...
+
+class Command(BaseCommand):
+    help = 'Add missing UUIDs to the programCertificate table'
+
+    def get_uuid_from_api(self, id=None):
+        if id is None:
+            raise AttributeError('Need an ID')
+
+        item = get_program(id)
+        return item['uuid']
```
schenedx: In this case, I don't think the program can have empty `uuid` field. So we should blow up when that happens. So I am fine with not handling the empty case.

credentials/apps/credentials/management/commands/add_uuids_to_program_certificates.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def handle(self, *args, **options):
+        """ Add missing UUIDs to the programCertificate table"""
+
+        program_certificate_table = ProgramCertificate.objects.filter(program_uuid__isnull=True)
+
+        for program_entry in program_certificate_table:
+            try:
+                logger.info('Getting UUID for {}'.format(program_entry.program_id))
+                program_entry.program_uuid = self.get_uuid_from_api(program_entry.program_id)
```
schenedx: Instead of getting the program one by one from the API, why can't we just go ahead and get all programs from the API and store that in django cache or in memory dictionary? Then when you are editing each program,  you can use the dictionary to find the UUID. This avoids so many API calls

credentials/apps/credentials/management/commands/add_uuids_to_program_certificates.py
clintonb: `handle` is special. See https://docs.djangoproject.com/en/1.10/howto/custom-management-commands/.

credentials/apps/credentials/management/commands/add_uuids_to_program_certificates.py
clintonb: I'm on the fence here. We are making about 42 API calls. I'm not sure if it's worth optimizing the code for that.

schenedx: I thought about the same. Will this Management command only run once for the foreseeable future? If that's the case, I am ok for leaving it this state.

credentials/apps/credentials/management/commands/add_uuids_to_program_certificates.py
davec-edx: Item should never be empty, and if it was the error would be caught in the exception handling in the calling method where it would be reported.

credentials/apps/credentials/management/commands/add_uuids_to_program_certificates.py
davec-edx: fixed

davec-edx: fixed

credentials/apps/credentials/management/commands/add_uuids_to_program_certificates.py
davec-edx: The exception processing should log the exception and provide a stacktrace.  However, I've added the program ID that we're trying to retrieve to the log statement.

credentials/apps/credentials/management/commands/add_uuids_to_program_certificates.py
davec-edx: Yes, it's a one-time call.  Clinton and I talked about this one when it was being tested.

credentials/apps/credentials/management/commands/add_uuids_to_program_certificates.py
```diff
@@ -0,0 +1,34 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = 'Add missing UUIDs to the programCertificate table'
+
+    def get_uuid_from_api(self, id=None):
+        if id is None:
+            raise AttributeError('Need an ID')
```
amangano-edx: To me, default values suggest 'this value is optional'. If the value is not optional, I feel we should require it to always be explicitly passed.

schenedx:
How come there are no unit test for this command?

credentials/apps/credentials/management/commands/add_uuids_to_program_certificates.py
```diff
@@ -0,0 +1,50 @@
...
+        program_certificate_table = ProgramCertificate.objects.filter(program_uuid__isnull=True)
+
+        for program_entry in program_certificate_table:
+            try:
+                logger.info('Getting UUID for {}'.format(program_entry.program_id))
+                program_entry.program_uuid = self.get_uuid_from_api(program_entry.program_id)
+
+                program_entry.save()
+            except HttpNotFoundError:
+                logger.warning('ID not found: {}'.format(program_entry.program_id))
```
clintonb: `logger.warning('Program ID %d was not found for ProgramCertificate %s', program_entry.program_id, program_entry.id)`

credentials/apps/credentials/management/commands/add_uuids_to_program_certificates.py
```diff
@@ -0,0 +1,50 @@
...
+            except AttributeError as e:
```
clintonb: Remove this in favor of validating `program_entry` above.

credentials/apps/credentials/management/commands/add_uuids_to_program_certificates.py
```diff
@@ -0,0 +1,50 @@
...
+    run to do the initial population, it will not need to be run again.
+    """
+    help = 'Add missing UUIDs to the programCertificate table'
+
+    def get_uuid_from_api(self, id):
+        """ Return the UUID for the provided program ID """
+
+        # If there is no ID (meaning the program_id is empty), that is a problem,
+        # so report it, this likely means there is an issue with the Db
+        if id is None:
```
clintonb: I still think this is a bit of defensive coding overkill. We can look in the DB and verify that all `ProgramCertificate` entries have a `program_id`. However, if you must keep this validation, update `handle` to log an error if that assertion fails.

credentials/apps/credentials/management/commands/add_uuids_to_program_certificates.py
```diff
@@ -0,0 +1,50 @@
...
+                logger.info('Getting UUID for {}'.format(program_entry.program_id))
+                program_entry.program_uuid = self.get_uuid_from_api(program_entry.program_id)
+
+                program_entry.save()
+            except HttpNotFoundError:
+                logger.warning('ID not found: {}'.format(program_entry.program_id))
+            except AttributeError as e:
+                logger.warning('Bad ID: {} - {}'.format(program_entry.program_id, e))
+            except Exception:
+                logger.exception('Error detected trying to get the ID ({}) from the program service'
```
clintonb: The program ID is less important than the ID of the `ProgramCertificate` object.

`logger.exception('An error occurred while updating ProgramCertificate %d.', program_entry.id)`

credentials/apps/credentials/management/commands/add_uuids_to_program_certificates.py
```diff
@@ -0,0 +1,50 @@
...
+from django.core.management.base import BaseCommand
+from slumber.exceptions import HttpNotFoundError
+
+from credentials.apps.credentials.models import ProgramCertificate
+from credentials.apps.credentials.utils import get_program
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
```
clintonb: Add `# pragma: no cover` to appease coverage.

MatthewPiatetsky:
👍

davec-edx:
merged

PR #131, opened 2016-12-08
Replaced the Program and Organization APIs with the Catalog API

clintonb:
ECOM-6501


clintonb:
@edx/ecommerce please review

clintonb:
@edx/ecommerce please review

credentials/apps/credentials/views.py
```diff
@@ -50,19 +50,13 @@ def get_certificate_context(self, user_credential):
...
 
         # pylint: disable=no-member
         return {
             'credential_type': _('{program_type} Certificate').format(program_type=program_details.type),
             'credential_title': user_credential.credential.title,
             'user_data': get_user_data(user_credential.username),
-            'program_details': program_details,
-            'organization_data': organization_data,
-            'organization_name': organization_name,
+            'program_details': user_credential.credential.program_details,
```
rlucioni: Can this be simplified?
```py
'program_details': program_details,
```

credentials/apps/credentials/models.py
```diff
@@ -264,28 +264,26 @@ class Meta(object):
...
-            data = get_program(self.program_id)
-            return ProgramDetails(
-                uuid=data['uuid'],
-                title=data['name'],
-                type=data['category'],
-                organization_keys=[organization['key'] for organization in data['organizations']],
-                course_count=len(data['course_codes'])
-            )
+        client = self.site.siteconfiguration.catalog_api_client
+        data = client.programs(self.program_uuid.hex).get()
```
rlucioni: Thoughts on caching this data? It won't change very often.

clintonb: Hmm...I must have missed restoring that. Yes, this should be cached.

credentials/apps/credentials/views.py
rlucioni: The variable in question is already being used above on line 52: https://github.com/edx/credentials/pull/131/files/0f241a654f98c7e226da792d64d4f06dd4c2dea1#diff-ff7a742772c12a35314c107251af5a80R52.

clintonb:
@rlucioni thanks for the review. Feedback addressed.

MatthewPiatetsky:
@clintonb is this affected by what Lauren said regarding XSeries?

credentials/templates/credentials/program_certificate.html
```diff
@@ -31,7 +31,7 @@ <h2 class="message-title">
                         </h2>
 
                         <p class="message-copy">
-                            {% blocktrans with organization_name=certificate_context.organization_data.short_name %}
+                            {% blocktrans with organization_name=certificate_context.program_details.organizations.0.key %}
```
clintonb: Yes.

credentials/apps/credentials/models.py
```diff
@@ -261,31 +262,43 @@ def __str__(self):
...
+        program_uuid = self.program_uuid.hex
+        cache_key = 'programs.api.data.{uuid}'.format(uuid=program_uuid)
+        program = cache.get(cache_key)
+
+        if program:
+            return program
+
+        client = self.site.siteconfiguration.catalog_api_client
+        program = client.programs(program_uuid).get()
+        cache.set(cache_key, program, settings.PROGRAMS_CACHE_TTL)
```
rlucioni: This could be addressed separately, but I wanted to at least note that this TTL is currently hardcoded to 30 minutes. It should probably be at least an hour, and it also seems like a natural fit for django-solo.

PR #132, opened 2016-12-09
Updated configuration for Docker-based devstack

clintonb:
ECOM-6571


clintonb:
@rlucioni @davec-edx @jmbowman please review
FYI @edx/devops

Makefile
```diff
@@ -14,6 +15,7 @@ help:
 	@echo "  help                       display this help message"
 	@echo "  html_coverage              generate and view HTML coverage report"
 	@echo "  migrate                    apply database migrations"
+	@echo "  production-requirements    install requirements for production"
```
davec-edx: Should be an _ instead of a - in the name to be consistent with the other commands

Makefile
```diff
@@ -40,6 +42,9 @@ clean:
 clean_static:
 	rm -rf credentials/assets/
 
+production-requirements: requirements.js
```
davec-edx: use _ instead of -

PR #136, opened 2016-12-20
Added V2 functionality to handle queries against a program_uuid inste…

davec-edx:
Part of the conversion from program_id to program_uuid...   Added a V2 endpoint for credentials that would function using program_uuid instead of a program_id.   

Added unit tests for this V2 functionality
Adjusted V1 unit tests to confirm it is a V1 request not a V2
ECOM-6482
@edx/ecommerce

### DIFF
--- a/credentials/apps/api/filters.py
+++ b/credentials/apps/api/filters.py
@@ -5,17 +5,6 @@ import django_filters
 from credentials.apps.credentials.models import UserCredential
 
 
-class ProgramFilter(django_filters.FilterSet):
-    """ Allows for filtering program credentials by their program_id and status
-    using a query string argument.
-    """
-    program_id = django_filters.NumberFilter(name="program_credentials__program_id")
-
-    class Meta:
-        model = UserCredential
-        fields = ['program_id', 'status']
-
-
 class CourseFilter(django_filters.FilterSet):
     """ Allows for filtering course credentials by their course_id, status and
     certificate_type using a query string argument.
--- a/credentials/apps/api/tests/test_views.py
+++ b/credentials/apps/api/tests/test_views.py
@@ -1,14 +1,10 @@
-"""
-Tests for credentials service views.
-"""
-# pylint: disable=no-member
 from __future__ import unicode_literals
 import json
 
 import ddt
 from django.contrib.auth.models import Group, Permission
 from django.core.urlresolvers import reverse
-from rest_framework.test import APITestCase, APIRequestFactory
+from rest_framework.test import APIRequestFactory
 from testfixtures import LogCapture
 
 from credentials.apps.api.serializers import UserCredentialSerializer
@@ -17,20 +13,20 @@ from credentials.apps.core.tests.factories import UserFactory
 from credentials.apps.credentials.models import UserCredential
 from credentials.apps.credentials.tests import factories
 
-
 JSON_CONTENT_TYPE = 'application/json'
 LOGGER_NAME = 'credentials.apps.credentials.issuers'
 LOGGER_NAME_SERIALIZER = 'credentials.apps.api.serializers'
 
 
 @ddt.ddt
-class UserCredentialViewSetTests(APITestCase):
+class BaseUserCredentialViewSetTests(object):
     """ Tests for GenerateCredentialView. """
+    # pylint: disable=no-member
 
-    list_path = reverse("api:v1:usercredential-list")
+    list_path = None
 
     def setUp(self):
-        super(UserCredentialViewSetTests, self).setUp()
+        super(BaseUserCredentialViewSetTests, self).setUp()
 
         self.user = UserFactory()
         self.client.force_authenticate(self.user)
@@ -449,10 +445,13 @@ class UserCredentialViewSetTests(APITestCase):
 
 
 @ddt.ddt
-class UserCredentialViewSetPermissionsTests(APITestCase):
+class BaseUserCredentialViewSetPermissionsTests(object):
     """
     Thoroughly exercise the custom view- and object-level permissions for this viewset.
     """
+    # pylint: disable=no-member
+
+    list_path = None
 
     def make_user(self, group=None, perm=None, **kwargs):
         """ DRY helper to create users with specific groups and/or permissions. """
@@ -478,10 +477,9 @@ class UserCredentialViewSetPermissionsTests(APITestCase):
         The list method (GET) requires either 'view' permission, or for the
         'username' query parameter to match that of the requesting user.
         """
-        list_path = reverse("api:v1:usercredential-list")
 
         self.client.force_authenticate(self.make_user(**user_kwargs))
-        response = self.client.get(list_path, {'username': 'test-user'})
+        response = self.client.get(self.list_path, {'username': 'test-user'})
         self.assertEqual(response.status_code, expected_status)
 
     @ddt.data(
@@ -497,7 +495,6 @@ class UserCredentialViewSetPermissionsTests(APITestCase):
         """
         The creation (POST) method requires the 'add' permission.
         """
-        list_path = reverse('api:v1:usercredential-list')
         program_certificate = factories.ProgramCertificateFactory()
         post_data = {
             'username': 'test-user',
@@ -508,7 +505,7 @@ class UserCredentialViewSetPermissionsTests(APITestCase):
         }
 
         self.client.force_authenticate(self.make_user(**user_kwargs))
-        response = self.client.post(list_path, data=json.dumps(post_data), content_type=JSON_CONTENT_TYPE)
+        response = self.client.post(self.list_path, data=json.dumps(post_data), content_type=JSON_CONTENT_TYPE)
         self.assertEqual(response.status_code, expected_status)
 
     @ddt.data(
@@ -563,95 +560,14 @@ class UserCredentialViewSetPermissionsTests(APITestCase):
         self.assertEqual(response.status_code, expected_status)
 
 
-class CredentialViewSetTests(APITestCase):
-    """ Base Class for ProgramCredentialViewSetTests and CourseCredentialViewSetTests. """
+class BaseCourseCredentialViewSetTests(object):
+    """ Tests for CourseCredentialViewSetTests. """
+    # pylint: disable=no-member
 
     list_path = None
-    user_credential = None
 
     def setUp(self):
-        super(CredentialViewSetTests, self).setUp()
-
-        self.user = UserFactory()
-        self.user.groups.add(Group.objects.get(name=Role.ADMINS))
-        self.client.force_authenticate(self.user)
-        self.request = APIRequestFactory().get('/')
-
-    def assert_permission_required(self, data):
-        """
-        Ensure access to these APIs is restricted to those with explicit model
-        permissions.
-        """
-        self.client.force_authenticate(user=UserFactory())
-        response = self.client.get(self.list_path, data)
-        self.assertEqual(response.status_code, 403)
-
-    def assert_list_without_id_filter(self, path, expected):
-        """Helper method used for making request and assertions. """
-        response = self.client.get(path)
-        self.assertEqual(response.status_code, 400)
-        self.assertEqual(response.data, expected)
-
-    def assert_list_with_id_filter(self, data):
-        """Helper method used for making request and assertions. """
-        expected = {'count': 1, 'next': None, 'previous': None,
-                    'results': [UserCredentialSerializer(self.user_credential, context={'request': self.request}).data]}
-        response = self.client.get(self.list_path, data)
-        self.assertEqual(response.status_code, 200)
-        self.assertEqual(response.data, expected)
-
-    def assert_list_with_status_filter(self, data):
-        """Helper method for making request and assertions. """
-        expected = {'count': 1, 'next': None, 'previous': None,
-                    'results': [UserCredentialSerializer(self.user_credential, context={'request': self.request}).data]}
-        response = self.client.get(self.list_path, data, expected)
-        self.assertEqual(json.loads(response.content), expected)
-
-
-class ProgramCredentialViewSetTests(CredentialViewSetTests):
-    """ Tests for ProgramCredentialViewSetTests. """
-
-    list_path = reverse("api:v1:programcredential-list")
-
-    def setUp(self):
-        super(ProgramCredentialViewSetTests, self).setUp()
-
-        self.program_certificate = factories.ProgramCertificateFactory()
-        self.program_id = self.program_certificate.program_id
-        self.user_credential = factories.UserCredentialFactory.create(credential=self.program_certificate)
-        self.request = APIRequestFactory().get('/')
-
-    def test_list_without_program_id(self):
-        """ Verify a list end point of program credentials will work only with
-        program_id filter.
-        """
-        self.assert_list_without_id_filter(path=self.list_path, expected={
-            'error': 'A program_id query string parameter is required for filtering program credentials.'
-        })
-
-    def test_list_with_program_id_filter(self):
-        """ Verify the list endpoint supports filter data by program_id."""
-        program_cert = factories.ProgramCertificateFactory(program_id=001)
-        factories.UserCredentialFactory.create(credential=program_cert)
-        self.assert_list_with_id_filter(data={'program_id': self.program_id})
-
-    def test_list_with_status_filter(self):
-        """ Verify the list endpoint supports filtering by status."""
-        factories.UserCredentialFactory.create_batch(2, status="revoked", username=self.user_credential.username)
-        self.assert_list_with_status_filter(data={'program_id': self.program_id, 'status': UserCredential.AWARDED}, )
-
-    def test_permission_required(self):
-        """ Verify that requests require explicit model permissions. """
-        self.assert_permission_required({'program_id': self.program_id, 'status': UserCredential.AWARDED})
-
-
-class CourseCredentialViewSetTests(CredentialViewSetTests):
-    """ Tests for CourseCredentialViewSetTests. """
-
-    list_path = reverse("api:v1:coursecredential-list")
-
-    def setUp(self):
-        super(CourseCredentialViewSetTests, self).setUp()
+        super(BaseCourseCredentialViewSetTests, self).setUp()
 
         self.course_certificate = factories.CourseCertificateFactory()
         self.course_id = self.course_certificate.course_id
--- a/credentials/apps/api/v2/tests/test_views.py
+++ b/credentials/apps/api/v2/tests/test_views.py
@@ -1,0 +1,94 @@
+"""
+Tests for credentials service views.
+"""
+from __future__ import unicode_literals
+
+from django.core.urlresolvers import reverse
+from rest_framework.test import APIRequestFactory, APITestCase
+
+from credentials.apps.api.tests.mixins import CredentialViewSetTestsMixin
+from credentials.apps.api.tests.test_views import (
+    BaseUserCredentialViewSetTests, BaseUserCredentialViewSetPermissionsTests, BaseCourseCredentialViewSetTests,
+)
+from credentials.apps.credentials.models import UserCredential
+from credentials.apps.credentials.tests import factories
+
+JSON_CONTENT_TYPE = 'application/json'
+LOGGER_NAME = 'credentials.apps.credentials.issuers'
+LOGGER_NAME_SERIALIZER = 'credentials.apps.api.serializers'
+
+
+class ProgramCredentialViewSetTests(CredentialViewSetTestsMixin, APITestCase):
+    """ Tests for ProgramCredentialViewSetTests. """
+
+    list_path = reverse("api:v2:programcredential-list")
+
+    def setUp(self):
+        super(ProgramCredentialViewSetTests, self).setUp()
+
+        self.program_certificate = factories.ProgramCertificateFactory()
+        self.program_id = self.program_certificate.program_id
+        self.program_uuid = self.program_certificate.program_uuid
+        self.user_credential = factories.UserCredentialFactory.create(credential=self.program_certificate)
+        self.request = APIRequestFactory().get('/')
+
+    def test_list_without_uuid(self):
+        """ Verify a list end point of program credentials will work with
+        program_uuid filter.
+        """
+        error_message = {'error': 'A UUID query string parameter is required for filtering program credentials.'}
+        self.assert_list_without_id_filter(path=self.list_path, expected=error_message)
+
+    def test_list_without_uuid_but_with_id(self):
+        """ Verify a list end point of program credentials will work with
+        program_uuid filter.
+        """
+        error_message = {'error': 'A UUID query string parameter is required for filtering program credentials.'}
+        self.assert_list_without_id_filter(path=self.list_path,
+                                           data={'program_id': self.program_id},
+                                           expected=error_message)
+
+    def test_list_with_uuid_and_id(self):
+        """ Verify a list end point of program credentials will not work with
+        program_id filter.
+        """
+        error_message = {'error': 'A program_id query string parameter was found in a V2 API request.'}
+        self.assert_list_without_id_filter(path=self.list_path,
+                                           data={'program_uuid': self.program_uuid, 'program_id': self.program_id},
+                                           expected=error_message)
+
+    def test_list_with_program_uuid_filter(self):
+        """ Verify the list endpoint supports filter data by program_uuid."""
+        self.assert_list_with_id_filter(data={'program_uuid': self.program_uuid})
+
+    def test_list_with_invalid_uuid(self):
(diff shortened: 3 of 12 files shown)


credentials/apps/api/filters.py
```diff
@@ -5,6 +5,17 @@
 from credentials.apps.credentials.models import UserCredential
 
 
+class ProgramFilterByUuid(django_filters.FilterSet):
```
clintonb: This should be `UserCredentialFilter` going forward since it is a filter for `UserCredential` objects. Sub-class it, add the `program_id` filter, call it something else, and use the sub-class for the v1 API. You could also move the sub-class to `/api/v1/filters.py`.

credentials/apps/api/tests/test_views.py
```diff
@@ -629,22 +636,114 @@ def test_list_without_program_id(self):
...
+        program_uuid filter.
+        """
+        self.assert_list_without_id_filter(path=self.list_path, data={'program_uuid': self.program_id,
+                                                                      'program_id': self.program_id}, expected={
+            'error': 'A program_uuid query string parameter should not appear in V1 queries.'
+        })
+
     def test_list_with_program_id_filter(self):
         """ Verify the list endpoint supports filter data by program_id."""
         program_cert = factories.ProgramCertificateFactory(program_id=001)
```
clintonb: `1`

credentials/apps/api/tests/test_views.py
```diff
@@ -629,22 +636,114 @@ def test_list_without_program_id(self):
...
         factories.UserCredentialFactory.create(credential=program_cert)
         self.assert_list_with_id_filter(data={'program_id': self.program_id})
 
+    def test_list_with_program_invalid_id_filter(self):
+        """ Verify the list endpoint supports filter data by program_id."""
+        program_cert = factories.ProgramCertificateFactory(program_id=001)
```
clintonb: `1`

credentials/apps/api/v1/views.py
```diff
@@ -54,6 +54,10 @@ def list(self, request, *args, **kwargs):
             raise ValidationError(
                 {'error': 'A program_id query string parameter is required for filtering program credentials.'})
 
+        if self.request.query_params.get('program_uuid'):
+            raise ValidationError(
+                {'error': 'A program_uuid query string parameter should not appear in V1 queries.'})
```
clintonb: Why not? We can be forwards-compatible.

credentials/apps/api/v2/tests/__init__.py
```diff
@@ -0,0 +1 @@
+# Create your tests in sub-packages prefixed with "test_" (e.g. test_views).
```
clintonb: Useless comment is useless. Also, where are the tests for `ProgramsCredentialsViewSet`?

credentials/apps/api/v2/urls.py
```diff
@@ -0,0 +1,11 @@
+""" API v2 URLs. """
+from rest_framework.routers import DefaultRouter
+
+from credentials.apps.api.v2 import views
+
+
+router = DefaultRouter()  # pylint: disable=invalid-name
+# URL can not have hyphen as it is not currently supported by slumber
```
clintonb: URLs cannot contain hyphens since they are not supported by Slumber. See https://github.com/samgiles/slumber/issues/44.

credentials/apps/api/v2/urls.py
```diff
@@ -0,0 +1,11 @@
...
+# as mentioned https://github.com/samgiles/slumber/issues/44
+router.register(r'program_credentials', views.ProgramsCredentialsViewSet, base_name='programcredential')
```
clintonb: You also need to include the other resources from API v1.

credentials/apps/api/v2/views.py
```diff
@@ -0,0 +1,37 @@
+"""
+Credentials service API views (v1).
```
clintonb: These docstrings are rarely useful. Remove them.

credentials/apps/api/v2/urls.py
```diff
@@ -0,0 +1,11 @@
+""" API v2 URLs. """
```
clintonb: Remove.

credentials/apps/api/v2/views.py
```diff
@@ -0,0 +1,37 @@
...
+    serializer_class = UserCredentialSerializer
+
+    def list(self, request, *args, **kwargs):
+        # Validate that we do have a program_uuid to use
+        if not self.request.query_params.get('program_uuid'):
+            raise ValidationError(
+                {'error': 'A UUID query string parameter is required for filtering program credentials.'})
+
+        # Confirmation that we are not supplying both parameters. We should only be providing the program_uuid in V2
+        if self.request.query_params.get('program_id'):
```
clintonb: This check is not useful. Just ignore the parameter.

credentials/apps/api/tests/test_views.py
```diff
@@ -629,22 +636,114 @@ def test_list_without_program_id(self):
...
+        """ Verify the list endpoint supports filtering by status."""
+        self.assert_list_with_status_filter(data={'program_id': self.program_id, 'status': UserCredential.REVOKED},
+                                            should_exist=False)
 
     def test_permission_required(self):
         """ Verify that requests require explicit model permissions. """
         self.assert_permission_required({'program_id': self.program_id, 'status': UserCredential.AWARDED})
 
 
+class ProgramCredentialViewSetTestsV2(CredentialViewSetTests):
```
clintonb: This belongs in the `v2` package.

credentials/apps/api/v1/views.py
davec-edx: We shouldn't be injecting additional query parameters, the consumer is either calling a V1 or a V2, if they have the uuid parameter it should be V2.  If we do this, then what's the point of versioning the api?

credentials/apps/api/v2/views.py
clintonb: I would treat this as any other arbitrary parameter. We don't explicitly block calls made with other parameters. Why block this one? If they aren't using UUID, we'll error out. If they are, they will get the expected result. The odds of anyone using both are low. This is unnecessary defensive coding.

credentials/apps/api/tests/test_views.py
```diff
@@ -17,20 +14,71 @@
 from credentials.apps.credentials.models import UserCredential
 from credentials.apps.credentials.tests import factories
 
-
 JSON_CONTENT_TYPE = 'application/json'
 LOGGER_NAME = 'credentials.apps.credentials.issuers'
 LOGGER_NAME_SERIALIZER = 'credentials.apps.api.serializers'
 
 
+class CredentialViewSetTests(APITestCase):
```
clintonb: This should be a mixin—inherit `object`—to avoid executing `setUp` unnecessarily. The test runner considers anything that inherits `TestCase` to be a test class. Also, rename to `CredentialViewSetTestMixin`.

credentials/apps/api/v1/tests/test_api.py
```diff
@@ -0,0 +1,91 @@
+"""
```
clintonb: This docstring can go away.

credentials/apps/api/v1/tests/test_api.py
```diff
@@ -0,0 +1,91 @@
...
+Tests for credentials service views.
+"""
+# pylint: disable=no-member
```
clintonb: We generally discourage file-wide disabling of PyLint warnings. Does this need to be disabled for anything outside of `setUp` (which uses factories)?

credentials/apps/api/v1/tests/test_api.py
```diff
@@ -0,0 +1,91 @@
...
+from rest_framework.test import APIRequestFactory, APITestCase
+
+from credentials.apps.api.tests.test_views import CredentialViewSetTests, BaseUserCredentialViewSetTests, \
+    BaseUserCredentialViewSetPermissionsTests, BaseCourseCredentialViewSetTests
+from credentials.apps.credentials.models import UserCredential
+from credentials.apps.credentials.tests import factories
+
+
+class ProgramCredentialViewSetTests(CredentialViewSetTests):
+    """ Tests for ProgramCredentialViewSetTests. """
```
clintonb: This docstring doesn't add any value. It can go.

credentials/apps/api/v1/tests/test_api.py
```diff
@@ -0,0 +1,91 @@
...
+
+    list_path = reverse("api:v1:programcredential-list")
```
clintonb: Single quotes.

credentials/apps/api/v1/tests/test_api.py
```diff
@@ -0,0 +1,91 @@
...
+        super(ProgramCredentialViewSetTests, self).setUp()
+
+        self.program_certificate = factories.ProgramCertificateFactory()
+        self.program_id = self.program_certificate.program_id
+        self.user_credential = factories.UserCredentialFactory.create(credential=self.program_certificate)
+        self.request = APIRequestFactory().get('/')
+
+    def test_list_without_program_id(self):
+        """ Verify a list end point of program credentials will work only with
+        program_id filter.
```
clintonb: Can this fit on the line above?

credentials/apps/api/v1/tests/test_api.py
```diff
@@ -0,0 +1,91 @@
+"""
+Tests for credentials service views.
+"""
+# pylint: disable=no-member
+from __future__ import unicode_literals
+
+from django.core.urlresolvers import reverse
+from rest_framework.test import APIRequestFactory, APITestCase
+
+from credentials.apps.api.tests.test_views import CredentialViewSetTests, BaseUserCredentialViewSetTests, \
```
clintonb: We generally wrap imports in parentheses and avoid the slash. The newline is still needed to avoid going over 120 characters.

```python
from credentials.apps.api.tests.test_views import (CredentialViewSetTests, BaseUserCredentialViewSetTests, BaseUserCredentialViewSetPermissionsTests, BaseCourseCredentialViewSetTests)
```

credentials/apps/api/v1/tests/test_api.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def test_list_without_program_id(self):
+        """ Verify a list end point of program credentials will work only with
+        program_id filter.
+        """
+        self.assert_list_without_id_filter(path=self.list_path, expected={
+            'error': 'A program_id query string parameter is required for filtering program credentials.'
+        })
+
+    def test_list_with_uuid_but_not_id(self):
```
clintonb: I don't see much value in testing this specific case. This endpoint has never used the UUID, so back-porting the test case doesn't do anything for us.

credentials/apps/api/v1/tests/test_api.py
```diff
@@ -0,0 +1,91 @@
...
+        """ Verify a list end point of program credentials will work with
+        program_uuid filter.
+        """
+        self.assert_list_without_id_filter(path=self.list_path, data={'program_uuid': self.program_id}, expected={
+            'error': 'A program_id query string parameter is required for filtering program credentials.'
+        })
+
+    def test_list_with_both_uuid_and_id(self):
```
clintonb: Same comment as above. There is no reason to expect including the `program_uuid` parameter will change the behavior of the v1 API, so there is no need to test it.

credentials/apps/api/v1/views.py
clintonb: I would leave the v1 API alone. It is already setup properly to respond to requests that don't include the `program_id` parameter. If a client passes the `program_uuid` parameter, the v1 API shouldn't care. It's the same as if the client passed the `foobar` parameter.

credentials/apps/api/v2/tests/test_api.py
```diff
@@ -0,0 +1,93 @@
+"""
+Tests for credentials service views.
+"""
+# pylint: disable=no-member
```
clintonb: See my previous comment about disabling PyLint warnings for the entire file.

credentials/apps/api/v2/tests/test_api.py
```diff
@@ -0,0 +1,93 @@
...
+from __future__ import unicode_literals
+
+from django.core.urlresolvers import reverse
+from rest_framework.test import APIRequestFactory, APITestCase
+
+from credentials.apps.api.tests.test_views import CredentialViewSetTests, BaseUserCredentialViewSetTests, \
```
clintonb: See previous comment about parentheses.

credentials/apps/api/v2/tests/test_api.py
```diff
@@ -0,0 +1,93 @@
+"""
```
clintonb: This file should be `test_views.py`.

credentials/apps/api/tests/test_views.py
```diff
@@ -17,20 +14,71 @@
...
+    def assert_permission_required(self, data):
+        """
+        Ensure access to these APIs is restricted to those with explicit model
+        permissions.
+        """
+        self.client.force_authenticate(user=UserFactory())
+        response = self.client.get(self.list_path, data)
+        self.assertEqual(response.status_code, 403)
+
+    def assert_list_without_id_filter(self, path, expected, data=None):
```
clintonb: `assert_bad_request`

credentials/apps/api/v2/tests/test_api.py
```diff
@@ -0,0 +1,93 @@
...
+        self.request = APIRequestFactory().get('/')
+
+    def test_list_without_uuid(self):
+        """ Verify a list end point of program credentials will work with
+        program_uuid filter.
+        """
+        error_message = {'error': 'A UUID query string parameter is required for filtering program credentials.'}
+        self.assert_list_without_id_filter(path=self.list_path, expected=error_message)
+
+    def test_list_without_uuid_but_with_id(self):
```
clintonb: I still don't see value in validating parameters don't work. Also, this becomes one more bit of code that needs to go away when we ultimately get rid of the `program_id` attribute/column. I understand why you're testing this; but, given that this API is only used externally and v1 will almost immediately be deprecated/removed once this is deployed, there is no need to worry about adding tests for `program_id` in API v2.

credentials/apps/api/v2/tests/test_api.py
```diff
@@ -0,0 +1,93 @@
...
+        """ Verify a list end point of program credentials will work with
+        program_uuid filter.
+        """
+        error_message = {'error': 'A UUID query string parameter is required for filtering program credentials.'}
+        self.assert_list_without_id_filter(path=self.list_path,
+                                           data={'program_id': self.program_id},
+                                           expected=error_message)
+
+    def test_list_with_uuid_and_id(self):
```
clintonb: See note above.

credentials/apps/api/tests/test_views.py
```diff
@@ -17,20 +14,71 @@
...
+        """Helper method used for making request and assertions. """
+        response = self.client.get(path, data)
+
+        self.assertEqual(response.status_code, 400)
+        self.assertEqual(response.data, expected)
+
+    def assert_list_with_id_filter(self, data=None, should_exist=True):
```
clintonb: `assert_list_path_success`

credentials/apps/api/tests/test_views.py
```diff
@@ -17,20 +14,71 @@
...
+        """Helper method used for making request and assertions. """
+        expected = self._generate_results(should_exist)
+        response = self.client.get(self.list_path, data)
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.data, expected)
+
+    def assert_list_with_status_filter(self, data, should_exist=True):
```
clintonb: This method is weird given that it hides the status filter in the `data` argument. Find a way to combine this with the method above. Both seem to do almost the same thing.

credentials/apps/api/v2/tests/test_api.py
```diff
@@ -0,0 +1,93 @@
...
+        self.assert_list_with_id_filter(data={'program_uuid': self.program_uuid})
+
+    def test_list_with_invalid_uuid(self):
+        """ Verify the list endpoint will fail if given a bad uuid."""
+        self.program_uuid = '12345678=0DAC-CAD0-ABCD-fedcba987654'
+        self.assert_list_with_id_filter(data={'program_uuid': self.program_uuid}, should_exist=False)
+
+    def test_list_with_status_filter(self):
+        """ Verify the list endpoint supports filtering by status."""
+        factories.UserCredentialFactory.create_batch(2, status="revoked", username=self.user_credential.username)
```
clintonb: UserCredential.REVOKED

credentials/apps/api/v2/tests/test_api.py
```diff
@@ -0,0 +1,93 @@
...
+        self.assert_list_with_status_filter(data={'program_uuid': self.program_uuid, 'status': UserCredential.AWARDED})
```
clintonb: I don't follow this test, which is a side-effect of these helper methods being used. It would actually be somewhat clearer to explicitly create the data, make the request, and validate the response in each test. The reliance on the helper methods was good in theory to de-dupe code. However, we seem to have taken that to an extreme that complicates understanding some of the tests.

credentials/apps/api/v2/views.py
```diff
@@ -0,0 +1,78 @@
...
+from django.http import Http404
+
+from rest_framework import mixins, viewsets
+from rest_framework.exceptions import ValidationError
+from credentials.apps.api.filters import CourseFilter
+
+from credentials.apps.api.permissions import UserCredentialViewSetPermissions
+from credentials.apps.api.serializers import UserCredentialCreationSerializer, UserCredentialSerializer
+from credentials.apps.credentials.models import UserCredential
+
```
clintonb: Remove this new line.

credentials/apps/api/v2/views.py
```diff
@@ -0,0 +1,78 @@
+import logging
+
+from django.http import Http404
+
+from rest_framework import mixins, viewsets
+from rest_framework.exceptions import ValidationError
+from credentials.apps.api.filters import CourseFilter
```
clintonb: This should be grouped below.

credentials/apps/api/v2/views.py
```diff
@@ -0,0 +1,78 @@
+import logging
+
+from django.http import Http404
```
clintonb: This should be grouped with third-party imports (e.g. `rest_framework`).

davec-edx: fixed

credentials/apps/api/v2/tests/test_api.py
davec-edx: fixed in v1 and v2

credentials/apps/api/v2/views.py
davec-edx: fixed in v1 and v2

credentials/apps/api/v2/views.py
davec-edx: fixed v1 and v2

clintonb:
@davec-edx please rebase to prepare for merging.

credentials/apps/api/v2/views.py
```diff
@@ -0,0 +1,77 @@
...
+from rest_framework import mixins, viewsets
+from rest_framework.exceptions import ValidationError
+
+from credentials.apps.api.filters import CourseFilter
+from credentials.apps.api.permissions import UserCredentialViewSetPermissions
+from credentials.apps.api.serializers import UserCredentialCreationSerializer, UserCredentialSerializer
+from credentials.apps.api.v2.filters import UserCredentialFilter
+
```
clintonb: Remove this newline

credentials/apps/api/v2/views.py
clintonb: This still needs to be removed, along with associated tests.

credentials/apps/api/v2/tests/test_views.py
```diff
@@ -0,0 +1,93 @@
+"""
+Tests for credentials service views.
+"""
+from __future__ import unicode_literals
+
+from django.core.urlresolvers import reverse
+from rest_framework.test import APIRequestFactory, APITestCase
+
+from credentials.apps.api.tests.test_views import (CredentialViewSetTests, BaseUserCredentialViewSetTests,
```
clintonb: Move the contents within the parentheses to ease indentation issues:

```python
from credentials.apps.api.tests.test_views import (
  CredentialViewSetTests, BaseUserCredentialViewSetTests, BaseUserCredentialViewSetPermissionsTests,
  BaseCourseCredentialViewSetTests
)
```

credentials/apps/api/v1/views.py
clintonb: This still needs to be removed.

credentials/apps/api/v1/views.py
```diff
@@ -1,17 +1,15 @@
...
 from django.http import Http404
 from rest_framework import mixins, viewsets
 from rest_framework.exceptions import ValidationError
-from credentials.apps.api.filters import ProgramFilter, CourseFilter
 
+from credentials.apps.api.filters import CourseFilter
 from credentials.apps.api.permissions import UserCredentialViewSetPermissions
 from credentials.apps.api.serializers import UserCredentialCreationSerializer, UserCredentialSerializer
-from credentials.apps.credentials.models import UserCredential
+from credentials.apps.api.v1.filters import UserCredentialFilter
```
clintonb: Remove the newline.

credentials/apps/api/tests/test_views.py
clintonb: This still needs to be addressed.

credentials/apps/api/tests/test_views.py
clintonb: This can be ignored. We'll fix it later.

credentials/apps/api/tests/test_views.py
clintonb: This can be ignored. We'll fix it later.

credentials/apps/api/tests/test_views.py
clintonb: This can be ignored. We'll fix it later.

credentials/apps/api/tests/test_views.py
davec-edx: fixed

credentials/apps/api/v2/views.py
davec-edx: We talked about this on Tuesday and thought the resolution was different.  API security dictates that we should A) Validate that all parameters/options that are given are of the correct type, and B) Validate that extra parameters/options are not given.  This is to prevent injections that could be potentially harmful.  Additionally, in a versioned API we can add helper methods for deprecation of an older version so that consumers of the API are checked to ensure they are not passing "old" parameters.  We should add a subtask to the deprecation of the old version to remove this.

credentials/apps/api/v1/views.py
davec-edx: This was a helper for the upgrade to V2.  You are correct, the parameter will be ignored, but it will result in the wrong type of error (incorrect parameter vs missing).  When we deprecate it it will go away.  
Removing it now.

credentials/apps/api/v2/views.py
davec-edx: fixed

credentials/apps/api/v2/views.py
clintonb: The query string parameters have nothing to do with security. Yes, we will validate parameters we do care about. When we get parameters we don't care about, we just ignore them. This is the case for all of our APIs. The only place where we drop requests with unexpected parameters are for APIs behind Amazon's API Gateway, and that is only due to a (now lifted) restriction of API Gateway.

We (edx.org) are the only consumers of this API. There is no need for the helpers. The work to modify edx-platform to use this new version of the API is already in the review state: https://github.com/edx/edx-platform/pull/14259.

If this were an API with a larger user base, such as the Catalog API, I would agree that we should be more careful. Since we have exactly one consumer, all you're doing here is adding debt.

credentials/apps/api/v2/views.py
```diff
@@ -0,0 +1,76 @@
...
+        return super(UserCredentialViewSet, self).list(request, *args, **kwargs)  # pylint: disable=maybe-no-member
+
+    def create(self, request, *args, **kwargs):
+        self.serializer_class = UserCredentialCreationSerializer
+        return super(UserCredentialViewSet, self).create(request, *args, **kwargs)
+
+
+class ProgramsCredentialsViewSet(mixins.ListModelMixin, viewsets.GenericViewSet):
+    """It will return the all credentials for programs based on the program_uuid."""
```
rlucioni: Could you clean up this docstring?

credentials/apps/api/v2/views.py
rlucioni: @davec-edx I'd like to second @clintonb here. If a parameter we don't expect or otherwise need is provided, it'll be ignored. I don't see the value in including this.

PR #140, opened 2017-01-04
Add template for generic program credential

rlucioni:
Removes hardcoded references to XSeries and cleans up unused assets. This template can be used to render credentials for XSeries, MicroMasters, and Professional Certificate programs. Note how the program type is slugified to assist with asset retrieval. This is a trick borrowed from Django itself that makes this template significantly cleaner.

ECOM-6531

@edx/ecommerce please review.


credentials/templates/credentials/program.html
```diff
@@ -50,13 +48,16 @@ <h3 class="sr-only">{% trans "Print or share your certificate:" %}</h3>
...
-                    <img src="{% static 'images/xseries-seal-grayscale.png' %}" class="background-image">
-                    <div class="wrapper-accomplishment-title xseries">
+                    {% with certificate_context.program_details.type|slugify as program_type %}
+                    {# http://stackoverflow.com/a/16656232 #}
+                    <img class="watermark" src="{% static 'images/'|add:program_type|add:'-watermark.png' %}">
+                    {% endwith %}
+                    <div class="wrapper-accomplishment-title">
                         <h2 class="accomplishment-title">
-                            <span class="accomplishment-title-label">{% trans "XSeries" %}</span>
+                            {# TODO: Replace with SVG logo specific to program type. #}
```
MatthewPiatetsky: What is this todo?  
Does it block merging this PR? 
Should it be a ticket?

MatthewPiatetsky:
are the assets here the actual ones or temporary? If they are temporary, can you create a ticket as part of the prof cert epic to get the real assets to make sure we don't forget

credentials/static/sass/_views.scss
```diff
@@ -1,99 +1,23 @@
...
-
-// ------------------------------
-// #HONOR CODE CERT
-// ------------------------------
-
-
-// ------------------------------
-// #VERIFIED CERT
-// ------------------------------
-.certificate-verified {
```
MatthewPiatetsky: it seems like you are deleting css in a bunch of places. Are you sure that these styles aren't also used elsewhere?

credentials/templates/credentials/program.html
```diff
@@ -170,14 +171,16 @@ <h3 class="accomplishment-metadata-title">{% trans "About edX" %}</h3>
                     </div>
 
                     <div class="accomplishment-metadata-bit accomplishment-metadata-aboutcert">
-                        <h3 class="accomplishment-metadata-title">{% trans "About edX XSeries Certificates" %}</h3>
-
+                        <h3 class="accomplishment-metadata-title">
+                        {% blocktrans with certificate_name=certificate_context.credential_type %}
+                            About edX {{ certificate_name }}s
```
MatthewPiatetsky: does this work with XSeries and MicroMasters? XSeriess MicroMasterss

credentials/templates/credentials/program.html
rlucioni: This is tracked by subtasks on ECOM-4639. The comment is to make the work easier for whoever ends up picking up those subtasks. I could have done it here but held off; I wanted to make it clear that this is where the SVG should be included.

credentials/static/sass/_views.scss
rlucioni: Yes, believe it or not. This and other code like it was copied wholesale from another repo and makes it hard to work in this repo.

credentials/templates/credentials/program.html
rlucioni: Take a look at how certificate_name is defined. It's not the same as the program type, it's the name of the credential (e.g., "XSeries Certificate," "MicroMasters Certificate").

rlucioni:
@MatthewPiatetsky any assets which survive this PR are legitimate. The others that are missing will be provided as part of other subtasks on the parent ticket.

MatthewPiatetsky:
👍

credentials/static/sass/_components-rendering.scss
```diff
@@ -15,149 +15,37 @@
...
   }
 
   .accomplishment-title-label {
+    @include margin-left(13px); // fix letterspacing shift
     display: block;
-    font-size: rem(40);
+    font-size: rem(52);
+    letter-spacing: .26em; // letterspacing relates to element font-size
 
     @include susy-breakpoint($bp-screen-md, $susy) {
```
clintonb: Any chance we can kill off susy, and upgrade to the latest version of [Pattern Library](http://ux.edx.org/)? This would greatly decrease static compilation times, and allow us to move to Webpack.

credentials/static/sass/_utilities.scss
```diff
@@ -1,39 +1,8 @@
...
-// ------------------------------
-
-// ------------------------------
-// #MIXINS
-// ------------------------------
-
-// ------------------------------
-// #HELPERS
-// ------------------------------
+$em-base: 16; // deliberately sets bourbon-based em-base (http://bourbon.io/docs/#em-base)
```
clintonb: Why not maintain the previous separation of these values?

credentials/apps/credentials/tests/test_views.py
```diff
@@ -148,7 +148,7 @@ def _assert_html_data(self, response):
         self.assertContains(response, 'offers interactive online classes and MOOCs from the')
         self.assertContains(
             response,
-            'An {platform_name} XSeries certificate signifies that the learner has'.format(
+            'An {platform_name} XSeries Certificate signifies that the learner has'.format(
```
rht: I've got a question: this assumes that `{platform_name}` always starts with a vowel-sounding char. If `settings.platform_name` defaults to 'edX', what about the the hardcoded 'edX' string elsewhere in the case when it refers to the platform instead of the repo?

credentials/static/sass/_views.scss
clintonb: The majority of these styles are related to course certificates, which we don't have here.

credentials/templates/credentials/program.html
clintonb: I share the same concern. Also, "edX" should not be hard-coded.

credentials/static/sass/_views.scss
rht: If this could be checked w/ https://www.npmjs.com/package/find-unused-css :question:

credentials/static/sass/_components-rendering.scss
rlucioni: @clintonb I'm sure we could, especially given that the surface area in this project is so small right now.

credentials/static/sass/_utilities.scss
rlucioni: I did away with the sections that were here because most of them were empty. Those that weren't didn't add much value. We can add them back if they become helpful in the future, but as it stands they were just noise.

credentials/templates/credentials/program.html
rlucioni: @clintonb can you elaborate on your concern? As things stand, `certificate_name` should always be the singular name for the credential, which is distinct from the string describing the program type. See [here](https://github.com/edx/credentials/blob/master/credentials/apps/credentials/views.py#L59).

clintonb: My biggest concern is that this cannot be translated properly. Adding an 's' works for pluralization in English and a few other languages. It doesn't work for many other languages. You should pass the pluralized name here.

rlucioni: Good point, I forgot about that. Resolving this correctly seems tricky. I don't think passing the pluralized name is enough, since the pluralization would still be in English. Ungettext also doesn't really seem right here. I can shuffle the structure of the sentence to make the pluralization unnecessary, so I'll go with that for now.

PR #141, opened 2017-01-04
Remove hardcoded references to edX

rlucioni:
@clintonb thanks for catching these.


rlucioni:
@clintonb I tacked an additional commit on here addressing the hard-to-translate plural form of "certificates" which you spotted. My change does away with the need for the plural altogether.

credentials/templates/credentials/program.html
```diff
@@ -172,7 +172,7 @@ <h3 class="accomplishment-metadata-title">{% blocktrans%}About {{ platform_name
                     <div class="accomplishment-metadata-bit accomplishment-metadata-aboutcert">
                         <h3 class="accomplishment-metadata-title">
                         {% blocktrans with certificate_name=certificate_context.credential_type %}
-                            About {{ platform_name }} {{ certificate_name }}s
+                            About your {{ platform_name }} {{ certificate_name }}
```
clintonb: Ehh...it's not really "your" certificate if I'm not the recipient. How about "the"?

rlucioni: Updated.

PR #143, opened 2017-01-05
Create an instance of the base program certificate for MicroMasters, XSeries and Professional Certificate

MatthewPiatetsky:
@edx/ecommerce


credentials/apps/credentials/views.py
```diff
@@ -68,7 +68,7 @@ class ExampleCredential(TemplateView):
     """ Example certificate. """
     template_name = 'credentials/base.html'
 
-    def get_context_data(self, **kwargs):
+    def get_context_data(self, program_type='MicroMasters', **kwargs):
```
rlucioni: Like I said earlier about the catalog management command, we should be careful about hardcoding brand names in open-source code. If there's value in per-type example certificates, you should at least default to `None` here and use `Example` as the type when someone GETs `/credentials/example`.

Related: I understand what you're after here, but it doesn't seem wise to assume that whatever is in the path is a program type, then try to generate a certificate. By doing this, we open ourselves up to a whole class of errors we wouldn't otherwise have to worry about.

clintonb: Don't modify the signature of this method. Use `kwargs` to pass values.

credentials/templates/credentials/images/micromasters-watermark.svg
```diff
@@ -1,4 +1,4 @@
-<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 1847.86 2677.65" role="img" aria-hidden="true">
+<svg class="background-watermark" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 1847.86 2677.65" role="img" aria-hidden="true">
```
clintonb: I can foresee this causing issues if we ever use this image for another purpose. Rather than set this class here, set it where the image is actually used.

credentials/templates/credentials/program.html
```diff
@@ -49,21 +49,15 @@ <h3 class="sr-only">{% trans "Print or share your certificate:" %}</h3>
 
         <main class="accomplishment accomplishment-main">
             <div class="wrapper-accomplishment-rendering">
-                <div class="accomplishment-rendering">
+                <div class="accomplishment-rendering" >
                     {% with certificate_context.program_details.type|slugify as program_type %}
                     {# http://stackoverflow.com/a/16656232 #}
```
clintonb: This comment is not very useful. Please remove it.

credentials/templates/credentials/program.html
```diff
@@ -49,21 +49,15 @@ <h3 class="sr-only">{% trans "Print or share your certificate:" %}</h3>
 
         <main class="accomplishment accomplishment-main">
             <div class="wrapper-accomplishment-rendering">
-                <div class="accomplishment-rendering">
+                <div class="accomplishment-rendering" >
```
clintonb: This is unnecessary and should be reverted.

credentials/templates/credentials/program.html
```diff
@@ -49,21 +49,15 @@ <h3 class="sr-only">{% trans "Print or share your certificate:" %}</h3>
...
                     {% with certificate_context.program_details.type|slugify as program_type %}
                     {# http://stackoverflow.com/a/16656232 #}
-                    <img class="watermark" src="{% static 'images/'|add:program_type|add:'-watermark.png' %}">
-                    {% endwith %}
+                    {% include 'credentials/images/'|add:program_type|add:'-watermark.svg' %}
```
clintonb: Why are we including the raw SVG source rather than loading this as an image?

mikedikan: Because we have to if we want the styles applied - this is a discussion that I had with @AlasdairSwan & @rlucioni and they can probably give the whole picture.

credentials/apps/credentials/tests/test_views.py
```diff
@@ -179,3 +179,19 @@ def test_get(self):
...
+        self.assertEqual(response.status_code, 200)
+        self.assertContains(response, 'micromasters-watermark')
+        self.assertContains(response, 'micromasters-logo')
+
+        response = self.client.get('/credentials/example/xseries')
+        self.assertEqual(response.status_code, 200)
+        self.assertContains(response, 'xseries-watermark')
+        self.assertContains(response, 'xseries-logo')
+
+        response = self.client.get('/credentials/example/professional%20certificate')
```
MatthewPiatetsky: changed

credentials/templates/credentials/images/micromasters-watermark.svg
MatthewPiatetsky: changed

credentials/templates/credentials/program.html
MatthewPiatetsky: fixed

credentials/templates/credentials/program.html
MatthewPiatetsky: removed

credentials/apps/credentials/views.py
MatthewPiatetsky: @rlucioni I changed program_type to a url parameter. This way we don't make assumptions about it being a program unless the param is passed (and I default program_type to None). Does that work?

credentials/apps/credentials/views.py
```diff
@@ -99,7 +100,7 @@ def get_context_data(self, **kwargs):
                 'program_details': ProgramDetails(
                     uuid=uuid.uuid4(),
                     title='Completely Example Program',
-                    type='Example',
+                    type=program_type,
```
rlucioni: If you change this to
```py
type=program_type or 'Example',
```
You'll get a nicer looking example credential when no program type is specified.

credentials/templates/credentials/program.html
```diff
@@ -51,19 +51,16 @@ <h3 class="sr-only">{% trans "Print or share your certificate:" %}</h3>
...
-                        <h2 class="accomplishment-title">
-                            {# TODO: Replace with SVG logo specific to program type. #}
-                            <span class="accomplishment-title-label">{% trans certificate_context.program_details.type %}</span>
-                            <span class="accomplishment-title-type">
-                                {% blocktrans with span_start='<span class="deco">' span_end='</span>' %}
-                                    Certificate {{ span_start }} of {{ span_end }} Achievement
-                                {% endblocktrans %}
-                            </span>
-                        </h2>
+                        {% if program_type != 'none' %}
```
rlucioni: If you take my suggestion above, you can also get rid of this conditional.

MatthewPiatetsky: but there is no watermark that looks like none-watermark or example-watermark. (same with logo)

credentials/apps/credentials/tests/test_views.py
```diff
@@ -179,3 +179,19 @@ def test_get(self):
         """ Verify the view loads. """
         response = self.client.get(reverse('credentials:example'))
         self.assertEqual(response.status_code, 200)
+
+    def test_get_all_types(self):
```
rlucioni: `ddt` is your friend here.
```py
@ddt.data('micromasters', 'xseries', 'professional-certificate')
def test_get_example_credentials(self, program_type):
    response = self.client.get('{}?program_type={}'.format(reverse('credentials:example'), program_type)
    for asset in ('watermark', 'logo'):
        self.assertContains(response, '{}-{}'.format(program_type, asset))
```

Also, this would be more complete with a test covering the empty case (when no program type is provided).

credentials/templates/credentials/program.html
```diff
@@ -51,19 +51,16 @@ <h3 class="sr-only">{% trans "Print or share your certificate:" %}</h3>
...
-                        <h2 class="accomplishment-title">
-                            {# TODO: Replace with SVG logo specific to program type. #}
-                            <span class="accomplishment-title-label">{% trans certificate_context.program_details.type %}</span>
-                            <span class="accomplishment-title-type">
-                                {% blocktrans with span_start='<span class="deco">' span_end='</span>' %}
-                                    Certificate {{ span_start }} of {{ span_end }} Achievement
-                                {% endblocktrans %}
-                            </span>
-                        </h2>
+                        {% if program_type != 'example' %}
```
rlucioni: Why is this necessary? Consider adding an `example-` watermark and logo so that the basic example credential looks good.

credentials/apps/credentials/tests/test_views.py
MatthewPiatetsky: thanks, changed. didn't do this previously when i had a space in professional certificate but now this works

credentials/templates/credentials/program.html
MatthewPiatetsky: added example watermark and logo

credentials/templates/credentials/program.html
rlucioni: @clintonb our commitment to supporting IE11 prevents us from using more modern methods of styling SVGs (e.g., CSS masks). An SVG included via CSS as a background image can't be easily styled without these. For more, see https://codepen.io/noahblon/post/coloring-svgs-in-css-background-images.

PR #145, opened 2017-01-06
Tweak program type logo css

MatthewPiatetsky:
The program type logos and edx logo were too big, so this makes them smaller. 
@edx/ecommerce


AlasdairSwan:
@MatthewPiatetsky we should not be styling elements directly, classes should be used instead.

PR #149, opened 2017-01-10
Removed all references to program ID in API v2

clintonb:
API v2 should only be utilizing program UUID when referring to programs. This fixes the references made for UserCredential creation.

ECOM-6482

### DIFF
--- a/credentials/apps/api/v2/serializers.py
+++ b/credentials/apps/api/v2/serializers.py
@@ -1,0 +1,134 @@
+"""
+Serializers for data manipulated by the credentials service APIs.
+"""
+import logging
+
+from django.core.exceptions import ObjectDoesNotExist
+from rest_framework import serializers
+from rest_framework.exceptions import ValidationError
+from rest_framework.reverse import reverse
+
+from credentials.apps.api.accreditors import Accreditor
+from credentials.apps.credentials.models import (
+    CourseCertificate, ProgramCertificate, UserCredential, UserCredentialAttribute
+)
+from credentials.apps.credentials.utils import validate_duplicate_attributes
+
+logger = logging.getLogger(__name__)
+
+
+class CredentialField(serializers.Field):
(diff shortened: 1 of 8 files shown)


clintonb:
@edx/ecommerce please review

credentials/apps/core/tests/mixins.py
```diff
@@ -19,7 +19,7 @@ def setUp(self):
         domain = 'testserver.fake'
         self.client = self.client_class(SERVER_NAME=domain)
 
-        Site.objects.all().delete()
+        Site.objects.all().delete()  # pylint: disable=no-member
```
rlucioni: Are you running a newer version of pylint than what's installed by the project requirements? Wondering why this is suddenly being flagged.

credentials/apps/api/v2/serializers.py
```diff
@@ -0,0 +1,134 @@
...
+
+class CredentialField(serializers.Field):
+    """ Field identifying the credential type and identifiers."""
+
+    def to_internal_value(self, data):
+        credential_id = None
+        try:
+            if 'program_uuid' in data and data.get('program_uuid'):
+                credential_id = data['program_uuid']
+                return ProgramCertificate.objects.get(program_uuid=credential_id, is_active=True)
```
rlucioni: The variable names being used here are confusing. `program_uuid` becomes `credential_id` which is used to filter the value of `program_uuid`?

credentials/apps/api/v2/serializers.py
```diff
@@ -0,0 +1,134 @@
...
+            elif 'course_id' in data and data.get('course_id') and data.get('certificate_type'):
+                credential_id = data['course_id']
```
rlucioni: Seeing as how this service doesn't host course credentials, you might be able to get rid of this. If not, I would fix the issue with calling a `course_id` a `credential_id`, as above.

PR #151, opened 2017-01-11
Removed unnecessary endpoints from API v2

clintonb:
Two of the three endpoints are solely used to filter UserCredentials, for which an endpoint already exists. It makes more sense to add filtering to the existing user_credentials endpoint rather than continue supporting extra endpoints that essentially duplicate existing behavior.

ECOM-6808


clintonb:
@edx/ecommerce please review

credentials/apps/api/v2/serializers.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+class CredentialField(serializers.Field):
+    """ Field identifying the credential type and identifiers."""
+
+    def to_internal_value(self, data):
+        program_uuid = data.get('program_uuid')
+
+        if not program_uuid:
+            raise ValidationError('Credential ID is missing.')
```
rlucioni: For clarity, should this refer to the UUID?

credentials/apps/api/v2/serializers.py
```diff
@@ -0,0 +1,121 @@
...
+
+        try:
+            return ProgramCertificate.objects.get(program_uuid=program_uuid, is_active=True)
+        except ObjectDoesNotExist:
+            msg = 'No active ProgramCertificate exists for program [{}]'.format(program_uuid)
+            logger.exception(msg)
+            raise ValidationError({'program_uuid': msg})
```
rlucioni: Is there a need to put this message in a dict? I see you're not doing it above in the previous `ValidationError`.

credentials/apps/api/v2/tests/test_serializers.py
```diff
@@ -0,0 +1,139 @@
...
+            self.assertEqual(ex.detail, expected)
+
+    def test_to_internal_value_with_empty_program_uuid(self):
+        """ Verify an error is raised if no program UUID is provided. """
+
+        with self.assertRaisesRegexp(ValidationError, 'Credential ID is missing'):
+            self.field_instance.to_internal_value({'program_uuid': ''})
+
+    def test_to_internal_value_with_invalid_program_uuid(self, ):
+        """Verify that it will return error message if program-id does not exist in db."""
```
rlucioni: Please update this docstring (and any others in this module) that still refer to the program ID.

credentials/apps/api/v2/urls.py
```diff
@@ -2,12 +2,9 @@
...
 from credentials.apps.api.v2 import views
 
-
 router = DefaultRouter()  # pylint: disable=invalid-name
 # URLs can not have hyphen as it is not currently supported by slumber
 # as mentioned https://github.com/samgiles/slumber/issues/44
-router.register(r'user_credentials', views.UserCredentialViewSet)
-router.register(r'program_credentials', views.ProgramsCredentialsViewSet, base_name='programcredential')
-router.register(r'course_credentials', views.CourseCredentialsViewSet, base_name='coursecredential')
+router.register(r'credentials', views.CredentialViewSet, base_name='credentials')
```
rlucioni: @MatthewPiatetsky note that this will require some changes in your PR. Currently edx-platform makes requests in a few places to the `user_credentials` resource.

credentials/apps/credentials/models.py
```diff
@@ -197,14 +197,19 @@ class UserCredential(TimeStampedModel):
         max_length=255, blank=True, null=True,
         help_text=_('Download URL for the PDFs.')
     )
-    uuid = models.UUIDField(default=uuid.uuid4, editable=False)
+    uuid = models.UUIDField(default=uuid.uuid4, editable=False, unique=True)
```
rlucioni: Nice catch.

credentials/apps/api/v2/serializers.py
clintonb: Not quite. ID in this case is the identifier of the credential type being awarded. For programs this is program_uuid. For courses, this will be course key and certificate type. I have changed "ID" to "identifier".

rlucioni: @clintonb not trying to split hairs, but is it the credential identifier that's missing, or the program/course identifier that's missing?

credentials/apps/api/v2/serializers.py
clintonb: That's a great question. Since this field, `credential`, is a `dict` rather than a single value, a `dict` is pass to the error class so that the error message is formatted to inform the user which key/nested field is invalid.

So, if I post an invalid body 
```json
{
   "username":"",
   "credential":{
      "program_uuid":123
   },
   "status":null,
   "attributes":[

   ]
}
```
I get a more detailed response like 
```json
{
    "username": [
        "This field may not be blank."
    ],
    "credential": {
        "program_uuid": "No active ProgramCertificate exists for program [123]"
    },
    "status": [
        "This field may not be null."
    ]
}
```
instead of 

```json
{
    "username": [
        "This field may not be blank."
    ],
    "credential": [
        "No active ProgramCertificate exists for program [123]"
    ],
    "status": [
        "This field may not be null."
    ]
}
```

credentials/apps/api/v2/tests/test_serializers.py
clintonb: Good catch. Done.

credentials/apps/api/v2/serializers.py
rlucioni: Got it, makes sense. Thanks for the explanation.

credentials/apps/api/v2/serializers.py
clintonb: Technically they are the same thing. This will be clearer when we restore course credentials. See https://github.com/edx/credentials/blob/master/credentials/apps/api/serializers.py#L23-L40.

PR #152, opened 2017-01-12
Added Django REST Swagger

clintonb:
ECOM-6808

Review #151 first


clintonb:
@edx/ecommerce please review

credentials/apps/api/v2/views.py
```diff
@@ -19,15 +19,53 @@ class CredentialViewSet(viewsets.ModelViewSet):
...
+        ---
+        omit_parameters:
+            - query
+        """
+        return super(CredentialViewSet, self).retrieve(request, *args, **kwargs)
+
+    def update(self, request, *args, **kwargs):
+        """ Update a credential.
+        ---
+        omit_parameters:
```
rlucioni: What does `omit_parameters` do?

clintonb: This prevents the query parameters that would show up due to the `filter_class`. We only need form parameters (from the serializers).

rlucioni: Maybe I'm misunderstanding, but wouldn't it be good to have query parameters from the `filter_class` (`UserCredentialFilter`, in this case) appear in the docs? Examples being `program_uuid`, `status`, and `username`.

clintonb: For the list endpoint, yes. However, those filters don't apply when modifying a single instance (obtained via UUID).

rlucioni: Oh, I completely missed that these were endpoints for modifying single instances. Thanks for the clarification!

PR #153, opened 2017-01-12
Corrected pagination class for API v2

clintonb:
All of our APIs should use page number pagination. This is now the default for the project. API v1 still retains limit-offset pagination to avoid breaking existing clients.

ECOM-6808


clintonb:
FYI @edx/ecommerce

PR #155, opened 2017-01-19
Added program validation to ProgramCertificateAdmin

clintonb:
ProgramCertificate objects can only be saved if the associated program's authoring organizations have certificate logos.

ECOM-6861


clintonb:
@edx/ecommerce please review

PR #157, opened 2017-01-20
Converted INSTALLED_APPS setting from a tuple to a list

clintonb:
This data type allows us to include additional custom apps from YAML in production.

ECOM-4639

### DIFF
--- a/credentials/settings/base.py
+++ b/credentials/settings/base.py
@@ -22,7 +22,9 @@ SITE_ID = 1
 
 # Application definition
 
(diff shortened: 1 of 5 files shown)


clintonb:
FYI @edx/ecommerce

e0d:
@clintonb nit, can you add a comment explaining the importance of the data type.  If there's a reason configuration should support tuples, can you open an OPS ticket for that?

👍

clintonb:
@e0d done

PR #158, opened 2017-01-20
Can't concatenate lists to tuples.

feanil:
This broke in our [CI environment yesterday](http://jenkins.edx.org:8080/job/ansible-provision/10892/console) with the following error:

`TypeError: can only concatenate tuple (not "list") to tuple`

@MatthewPiatetsky @clintonb Please review.  Also, can someone on your team look at why this slipped through other forms of testing?


clintonb:
@feanil this has been fixed with https://github.com/edx/credentials/pull/157.

This slipped through tests because our tests don't use `production.py`; they use `test.py`. We will fix this when we update the Docker image for Docker-based devstack.

feanil:
:+1:

PR #159, opened 2017-01-20
Adding exception throwing if logo image is missing

mikedikan:
ECOM-6863

Adding exception if logo image is missing in the program details.  This is expected to only be shown to PCs and edX staff during certificate configuration and is not likely to be learner-facing.

@edx/ecommerce for review

### DIFF
--- a/credentials/apps/credentials/exceptions.py
+++ b/credentials/apps/credentials/exceptions.py
@@ -1,0 +1,5 @@
+class MissingCertificateLogoError(Exception):
+    """ Raised when a Program fetched as part of a program certificate configuration
+     has an organization without a defined certificate logo image url
+    """
+    pass
--- a/credentials/apps/credentials/tests/test_views.py
+++ b/credentials/apps/credentials/tests/test_views.py
@@ -12,6 +12,7 @@ from mock import patch
 import responses
 
 from credentials.apps.core.tests.mixins import SiteMixin
+from credentials.apps.credentials.exceptions import MissingCertificateLogoError
 from credentials.apps.credentials.models import UserCredential
 from credentials.apps.credentials.tests import factories
 from credentials.apps.credentials.tests.mixins import UserDataMixin
@@ -21,7 +22,7 @@ from credentials.apps.credentials.tests.mixins import UserDataMixin
 class RenderCredentialPageTests(SiteMixin, TestCase):
     """ Tests for credential rendering view. """
     PRIMARY_ORGANIZATION_KEY = 'ACMEx'
-    PRIMARY_ORGANIZATION_LOGO_URL = 'http://example.com/image.jpg'
+    PRIMARY_ORGANIZATION_CERTIFICATE_LOGO_URL = 'http://example.com/image.jpg'
     PRIMARY_ORGANIZATION_NAME = 'ACME University'
 
     def setUp(self):
@@ -33,8 +34,12 @@ class RenderCredentialPageTests(SiteMixin, TestCase):
         self.user_credential = factories.UserCredentialFactory(credential=self.program_certificate)
         self.platform_name = self.site.siteconfiguration.platform_name
 
-    def _render_user_credential(self):
+    def _render_user_credential(self, use_proper_logo_url=True):
         """ Helper method to render a user certificate."""
+        if use_proper_logo_url:
+            certificate_logo_image_url = self.PRIMARY_ORGANIZATION_CERTIFICATE_LOGO_URL
+        else:
+            certificate_logo_image_url = None
         program_uuid = self.program_certificate.program_uuid
         program_endpoint = 'programs/{uuid}/'.format(uuid=program_uuid.hex)
         body = {
@@ -46,14 +51,14 @@ class RenderCredentialPageTests(SiteMixin, TestCase):
                     'uuid': uuid.uuid4().hex,
                     'key': self.PRIMARY_ORGANIZATION_KEY,
                     'name': self.PRIMARY_ORGANIZATION_NAME,
-                    'logo_image_url': self.PRIMARY_ORGANIZATION_LOGO_URL
+                    'certificate_logo_image_url': certificate_logo_image_url
 
                 },
                 {
                     'uuid': uuid.uuid4().hex,
                     'key': 'FakeX',
                     'name': 'Fake University',
-                    'logo_image_url': 'http://example.com/image.jpg'
+                    'certificate_logo_image_url': 'http://example.com/image.jpg'
                 }
             ],
             'courses': [
@@ -122,7 +127,7 @@ class RenderCredentialPageTests(SiteMixin, TestCase):
 
         # test organization related data.
         self.assertContains(response, self.PRIMARY_ORGANIZATION_KEY)
-        self.assertContains(response, self.PRIMARY_ORGANIZATION_LOGO_URL)
+        self.assertContains(response, self.PRIMARY_ORGANIZATION_CERTIFICATE_LOGO_URL)
 
         # test programs data
         self.assertContains(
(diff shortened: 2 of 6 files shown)


credentials/apps/credentials/views.py
```diff
@@ -40,6 +41,11 @@ def get_context_data(self, **kwargs):
             'certificate_context': self.get_certificate_context(user_credential),
         })
 
+        program_details = user_credential.credential.program_details
+        if not program_details.organizations[0].logo_image_url:
+            raise ImproperlyConfigured("No credential image logo defined for program: {program_name}".
```
davec-edx: Single quotes

davec-edx: Should the validation of the logo image be done just after checking that valid data was returned (a few lines above)?  No sense in creating an object if we're about to raise an error.

credentials/apps/credentials/views.py
```diff
@@ -40,6 +41,11 @@ def get_context_data(self, **kwargs):
             'certificate_context': self.get_certificate_context(user_credential),
         })
 
+        program_details = user_credential.credential.program_details
```
rlucioni: This data is already available below in `get_certificate_context`. I'd perform this check in that method.

credentials/apps/credentials/views.py
```diff
@@ -40,6 +41,11 @@ def get_context_data(self, **kwargs):
...
+        if not program_details.organizations[0].logo_image_url:
```
rlucioni: Iterate across all organizations. When possible, don't assume there's only one.

rlucioni: You should be checking for the presence of the `certificate_logo_image_url`.

credentials/apps/credentials/views.py
```diff
@@ -3,6 +3,7 @@
 import logging
 import uuid
 
+from django.core.exceptions import ImproperlyConfigured
```
rlucioni: `ImproperlyConfigured` is usually reserved for problems with settings (Django configuration). To avoid confusion and aid readability, create and raise your own exception:
```py
class MissingCertificateLogoError(Exception):
    pass

...

raise MissingCertificateLogoError
```

credentials/apps/credentials/views.py
mikedikan: I don't see certificate_logo_image_url in the data - only logo_image_url

rlucioni: You should see it, @clintonb added it a few days ago. https://github.com/edx/course-discovery/commit/8021f8ef7c8d623992fe8251e6be5cbd8444270a

clintonb: You don't see it because you haven't added it. Ironically, we made subtasks to update the API with the image; but, we didn't make a subtask to actually use the image on the certificate. It makes sense to go ahead and do that here.

mikedikan: yup, gonna do that now

mikedikan: @rlucioni in the view template - we assume there's only one - i was being consistent.  Should we update the code in both places?  Should we render two logos?  Should we use the first one we see (as currently implemented in the template).

rlucioni: @mikedikan fixing that assumption in the template is out of scope. We'll need to address it sooner or later, but we shouldn't repeat the same bad assumption here.

credentials/apps/credentials/views.py
mikedikan: @rlucioni I plan to take @davec-edx's approach below, in which case, the logic would happen before get_certificate_context is called.

mikedikan:
@edx/ecommerce for review

credentials/apps/credentials/tests/test_views.py
```diff
@@ -174,6 +179,14 @@ def test_signatory_organization_name_override(self):
         self.assertContains(response, self.signatory_1.organization_name_override)
         self.assertNotContains(response, self.signatory_2.organization_name_override)
 
+    @responses.activate
+    def test_logo_missing_exception(self):
+        user_credential = self.user_credential
+        with self.assertRaises(MissingCertificateLogoError) as context:
+            self._render_user_credential(use_proper_logo_url=False)
+
+        self.assertIn('No certificate image logo defined for program', context.exception.message)
```
clintonb: Use [`assertRaisesRegexp`](https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertRaisesRegexp).

credentials/apps/credentials/views.py
```diff
@@ -30,6 +30,12 @@ def get_context_data(self, **kwargs):
             status=UserCredential.AWARDED
         )
 
+        program_details = user_credential.credential.program_details
+        for organization in program_details.organizations:
+            if not organization.certificate_logo_image_url:
+                raise MissingCertificateLogoError('No certificate image logo defined for program: {program_name}'.
+                                                  format(program_name=program_details.title))
```
clintonb: Use the program's UUID. Also, we tend to wrap specific details in square brackets.

`No certificate image logo defined for program [{program_uuid}].`

credentials/apps/credentials/views.py
```diff
@@ -118,3 +124,7 @@ def get_context_data(self, **kwargs):
         })
 
         return context
+
+
+class MissingCertificateLogoError(Exception):
```
clintonb: This belongs in `exceptions.py`, and should have a docstring explaining when it should be used.

mikedikan:
@clintonb I believe your two comments were addressed, please have a look

credentials/apps/credentials/exceptions.py
```diff
@@ -0,0 +1,9 @@
+"""Exceptions and error messages used by the credentials application."""
+from __future__ import unicode_literals
```
clintonb: There is no need for this.

credentials/apps/credentials/exceptions.py
```diff
@@ -0,0 +1,9 @@
+"""Exceptions and error messages used by the credentials application."""
```
clintonb: Obvious do strong is obvious. This can go away.

credentials/apps/credentials/tests/test_views.py
```diff
@@ -15,13 +15,14 @@
 from credentials.apps.credentials.models import UserCredential
 from credentials.apps.credentials.tests import factories
 from credentials.apps.credentials.tests.mixins import UserDataMixin
+from credentials.apps.credentials.exceptions import MissingCertificateLogoError
```
clintonb: Alphabetize.

credentials/apps/credentials/tests/test_views.py
```diff
@@ -174,6 +179,13 @@ def test_signatory_organization_name_override(self):
         self.assertContains(response, self.signatory_1.organization_name_override)
         self.assertNotContains(response, self.signatory_2.organization_name_override)
 
+    @responses.activate
+    def test_logo_missing_exception(self):
+        with self.assertRaises(MissingCertificateLogoError) as context:
```
clintonb: See my previous comment. Use assertRaisesRegexp.

credentials/apps/credentials/views.py
```diff
@@ -11,6 +11,7 @@
 
 from credentials.apps.credentials.models import UserCredential, ProgramCertificate, ProgramDetails, OrganizationDetails
 from credentials.apps.credentials.utils import get_user_data
+from credentials.apps.credentials.exceptions import MissingCertificateLogoError
```
clintonb: Alphabetize.

credentials/apps/credentials/exceptions.py
mikedikan: huh?  You told me to add this in the previous PR review - what are you asking for here?

PR #163, opened 2017-02-01
Updated JWT Authentication settings

clintonb:
These changes bring the service in line with our other services.

ECOM-6958


clintonb:
FYI @edx/ecommerce @edx/devops

PR #164, opened 2017-02-02
Permissions Fixes

clintonb:


clintonb:
@edx/ecommerce please review

PR #165, opened 2017-02-02
Upgrade nodeenv to 1.1.1

bjacobel:
This fixes a bug in the Ansible provisioning process where an existing nodeenv could fail to be created when the node version was changed.

See edx/configuration#3648 and ekalinin/nodeenv#183 for more details.


gsong:
Adding myself to this thread so I can monitor progress.

PR #168, opened 2017-02-14
[ECOM-7453] Enable Configuration of a White Label site

marjev:
@mjfrey Here's the command that should be ran to update SiteConfiguration with WL specific information. I would like to get some feedback whether or not I'm on the right track here. There's a to-do list below showing what needs to be done before this can be merged.

JIRA story: https://openedx.atlassian.net/browse/ECOM-7453

To-do:
- [x] Write tests
- [x] Update documentation

### DIFF
--- a/credentials/apps/core/management/commands/create_or_update_site.py
+++ b/credentials/apps/core/management/commands/create_or_update_site.py
@@ -1,0 +1,143 @@
+""" Creates or updates a Site including SiteConfiguration data. """
+from __future__ import unicode_literals
+
+import logging
+
+from django.contrib.sites.models import Site
+from django.core.management import BaseCommand
+
+from credentials.apps.core.models import SiteConfiguration
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = 'Create or update Site and SiteConfiguration'
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            '--site-id',
+            action='store',
+            dest='site_id',
+            type=int,
+            help='ID of the Site to update.'
+        )
+        parser.add_argument(
+            '--site-domain',
+            action='store',
+            dest='site_domain',
+            type=str,
+            required=True,
+            help='Domain of the Site to create or update.'
+        )
+        parser.add_argument(
+            '--site-name',
+            action='store',
+            dest='site_name',
+            type=str,
+            required=True,
+            help='Name of the Site to create or update.'
+        )
+        parser.add_argument(
+            '--platform-name',
+            action='store',
+            dest='platform_name',
+            type=str,
+            required=True,
+            help='Name of your Open edX platform.'
+        )
+        parser.add_argument(
+            '--lms-url-root',
+            action='store',
+            dest='lms_url_root',
+            type=str,
+            required=True,
+            help='Root URL of LMS (e.g. https://localhost:8000)'
+        )
+        parser.add_argument(
+            '--catalog-api-url',
+            action='store',
+            dest='catalog_api_url',
+            type=str,
+            required=True,
+            help='Root URL of the Catalog API (e.g. https://api.edx.org/catalog/v1/)'
+        )
+        parser.add_argument(
+            '--tos-url',
+            action='store',
+            dest='tos_url',
+            type=str,
+            required=True,
+            help='Terms of Service URL'
+        )
+        parser.add_argument(
+            '--privacy-policy-url',
+            action='store',
+            dest='privacy_policy_url',
+            type=str,
+            required=True,
+            help='Privacy Policy URL'
+        )
+        parser.add_argument(
+            '--homepage-url',
+            action='store',
+            dest='homepage_url',
+            type=str,
+            required=True,
+            help='Homepage URL'
+        )
+        parser.add_argument(
+            '--company-name',
+            action='store',
+            dest='company_name',
+            type=str,
+            required=True,
+            help='Company Name'
+        )
+        parser.add_argument(
+            '--verified-certificate-url',
+            action='store',
+            dest='verified_certificate_url',
+            type=str,
+            required=True,
+            help='Verified Certificate URL'
+        )
+        parser.add_argument(
+            '--certificate-help-url',
+            action='store',
+            dest='certificate_help_url',
+            type=str,
+            required=True,
+            help='Certificate Help URL'
+        )
+
+    def handle(self, *args, **options):
+        site_id = options.get('site_id')
+        site_domain = options.get('site_domain')
+        site_name = options.get('site_name')
+
+        try:
+            site = Site.objects.get(id=site_id)
+        except Site.DoesNotExist:
+            site, site_created = Site.objects.get_or_create(domain=site_domain)
+            if site_created:
+                logger.info('Created Site [%d] with domain [%s]', site.id, site.domain)
+
+        site.domain = site_domain
+        site.name = site_name
+        site.save()
+
+        SiteConfiguration.objects.update_or_create(
+            site=site,
+            defaults={
+                'platform_name': options.get('platform_name'),
+                'lms_url_root': options.get('lms_url_root'),
+                'catalog_api_url': options.get('catalog_api_url'),
+                'tos_url': options.get('tos_url'),
+                'privacy_policy_url': options.get('privacy_policy_url'),
+                'homepage_url': options.get('homepage_url'),
+                'company_name': options.get('company_name'),
+                'verified_certificate_url': options.get('verified_certificate_url'),
+                'certificate_help_url': options.get('certificate_help_url')
+            }
+        )
--- a/credentials/apps/core/tests/test_commands.py
+++ b/credentials/apps/core/tests/test_commands.py
@@ -1,0 +1,129 @@
+from ddt import data, ddt
+
+from django.contrib.sites.models import Site
+from django.core.management import CommandError, call_command
+from django.test import TestCase
+
+from credentials.apps.core.tests.factories import SiteConfigurationFactory, SiteFactory
+
+
+@ddt
+class CreateOrUpdateSiteCommandTests(TestCase):
+    command_name = 'create_or_update_site'
+
+    def setUp(self):
+        super(CreateOrUpdateSiteCommandTests, self).setUp()
+
+        self.site_configuration = SiteConfigurationFactory.build()
+
+    def _call_command(self, site_domain, site_name, site_id=None):
+        """
+        Internal helper method for interacting with the create_or_update_site management command
+        """
+        # Required arguments
+        command_args = [
+            '--site-domain={site_domain}'.format(site_domain=site_domain),
+            '--platform-name={platform_name}'.format(platform_name=self.site_configuration.platform_name),
+            '--lms-url-root={lms_url_root}'.format(lms_url_root=self.site_configuration.lms_url_root),
+            '--catalog-api-url={catalog_api_url}'.format(catalog_api_url=self.site_configuration.catalog_api_url),
+            '--site-name={site_name}'.format(site_name=site_name),
+            '--tos-url={tos_url}'.format(tos_url=self.site_configuration.tos_url),
+            '--privacy-policy-url={privacy_policy_url}'.format(
+                privacy_policy_url=self.site_configuration.privacy_policy_url
+            ),
+            '--homepage-url={homepage_url}'.format(homepage_url=self.site_configuration.homepage_url),
+            '--company-name={company_name}'.format(company_name=self.site_configuration.company_name),
+            '--verified-certificate-url={verified_certificate_url}'.format(
+                verified_certificate_url=self.site_configuration.verified_certificate_url
+            ),
+            '--certificate-help-url={certificate_help_url}'.format(
+                certificate_help_url=self.site_configuration.certificate_help_url
+            )
+        ]
+
+        if site_id:
+            command_args.append('--site-id={site_id}'.format(site_id=site_id))
+
+        call_command(self.command_name, *command_args)
+
+    def _check_site_configuration(self, site_configuration):
+        """
+        Helper method for verifying that the Site is properly configured.
+
+        Args:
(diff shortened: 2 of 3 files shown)


marjev:
I've added the tests for the command. Before I update the documentation, I would like to get example values for the fields we're updating with this command.

During the manual testing I used dummy values, but I think we want something representative to be available in the documentation.

credentials/apps/core/management/commands/create_or_update_site.py
```diff
@@ -0,0 +1,147 @@
+""" Creates or updates a Site including SiteConfiguration data. """
+
+from __future__ import unicode_literals
+import logging
```
vkaracic: Add a space between the imports.

credentials/apps/core/management/commands/create_or_update_site.py
```diff
@@ -0,0 +1,147 @@
...
+        privacy_policy_url = options.get('privacy_policy_url')
+        homepage_url = options.get('homepage_url')
+        company_name = options.get('company_name')
+        verified_certificate_url = options.get('verified_certificate_url')
+        certificate_help_url = options.get('certificate_help_url')
+
+        try:
+            site = Site.objects.get(id=site_id)
+        except Site.DoesNotExist:
+            site, site_created = Site.objects.get_or_create(domain=site_domain)
```
vkaracic: Is ``site_id`` necessary? Because the ``domain`` field in ``Site`` is unique so you could just do the ``get_or_create`` without the try-except block.

credentials/apps/core/tests/test_commands.py
```diff
@@ -0,0 +1,146 @@
+from ddt import ddt, data
+from django.core.management import call_command, CommandError
+from django.contrib.sites.models import Site
```
vkaracic: _Alphabetize._
--  I.I.

credentials/apps/core/tests/test_commands.py
```diff
@@ -0,0 +1,146 @@
...
+            )
+
+        call_command(self.command_name, *command_args)
+
+    def _check_site_configuration(self, site):
+        """
+        Helper method for verifying that the Site is properly configured.
+
+        Args:
+            site (Site): Site for which the SiteConfiguration is being verified
```
vkaracic: ``Site whose SiteConfiguration is being verified.`` ?

credentials/apps/core/tests/test_commands.py
```diff
@@ -0,0 +1,146 @@
...
+        self.catalog_api_url = 'http://catalog.fake.server'
+        self.site_name = 'Test Site'
+        self.tos_url = 'http://fake.server/tos'
+        self.privacy_policy_url = 'http://fake.server/privacy-policy'
+        self.homepage_url = 'http://fake.server'
+        self.company_name = 'Test Company'
+        self.verified_certificate_url = 'http://fake.server/verified-certificate'
+        self.certificate_help_url = 'http://fake.server/certificate-help'
+
+    def _call_command(self, site_domain, platform_name, lms_url_root, catalog_api_url,
```
vkaracic: Why not default the arguments to the variables in ``setUp``?

credentials/apps/core/tests/test_commands.py
```diff
@@ -0,0 +1,146 @@
...
+
+        site.refresh_from_db()
+
+        self.assertEqual(site.domain, updated_site_domain)
+        self.assertEqual(site.name, updated_site_name)
+        self._check_site_configuration(site)
+
+    @data(
+        ['--site-id=1'],
+        ['--site-id=1'],
```
vkaracic: This seems to be a duplicate.

credentials/apps/core/management/commands/create_or_update_site.py
```diff
@@ -0,0 +1,147 @@
...
+            type=str,
+            required=True,
+            help='Domain of the Site to create or update.'
+        )
+        parser.add_argument(
+            '--site-name',
+            action='store',
+            dest='site_name',
+            type=str,
+            default='',
```
clintonb: This should be required.

clintonb: Actually, every field except site-id should be required. Otherwise, sites may be created that have broken links. The alternative is to update the templates to properly handle missing fields, and *not* display those links.

credentials/apps/core/management/commands/create_or_update_site.py
```diff
@@ -0,0 +1,147 @@
...
+        company_name = options.get('company_name')
+        verified_certificate_url = options.get('verified_certificate_url')
+        certificate_help_url = options.get('certificate_help_url')
+
+        try:
+            site = Site.objects.get(id=site_id)
+        except Site.DoesNotExist:
+            site, site_created = Site.objects.get_or_create(domain=site_domain)
+            if site_created:
+                logger.info('Site created with domain %s', site_domain)
```
clintonb: `logger.info('Created Site [%d] with domain [%s]', site.id, site.domain)`

credentials/apps/core/tests/test_commands.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+@ddt
+class CreateOrUpdateSiteCommandTests(TestCase):
+    command_name = 'create_or_update_site'
+
+    def setUp(self):
+        super(CreateOrUpdateSiteCommandTests, self).setUp()
+
+        self.platform_name = 'Test Platform'
```
clintonb: You can avoid a lot of this by using `SiteConfigurationFactory.build()`. See http://factoryboy.readthedocs.io/en/latest/index.html#using-factories.

credentials/apps/core/management/commands/create_or_update_site.py
marjev: I don't have a ticket with the list of use cases for this, so I had to copy paste the command from the ecommerce.

I think the idea there was to use site ID for updating. If both ID and domain are present in command args, it would mean: `Fetch the site with that ID and change its domain that value.`

@mjfrey Is this correct?

marjev:
@clintonb, @vkaracic Comments addressed. Please take another look before I start updating documentation. Thanks!

credentials/apps/core/management/commands/create_or_update_site.py
```diff
@@ -0,0 +1,154 @@
...
+
+        try:
+            site = Site.objects.get(id=site_id)
+        except Site.DoesNotExist:
+            site, site_created = Site.objects.get_or_create(domain=site_domain)
+            if site_created:
+                logger.info('Created Site [%d] with domain [%s]', site.id, site.domain)
+
+        site.domain = site_domain
+        if site_name:
```
vkaracic: If ``site_name`` is required, do you still need to the check?

credentials/apps/core/tests/test_commands.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+@ddt
+class CreateOrUpdateSiteCommandTests(TestCase):
+    command_name = 'create_or_update_site'
+
+    def setUp(self):
+        super(CreateOrUpdateSiteCommandTests, self).setUp()
+
+        self.site_configuration = SiteConfigurationFactory.build()
```
vkaracic: I think you can omit ``.build()``.

credentials/apps/core/tests/test_commands.py
```diff
@@ -0,0 +1,129 @@
...
+        site_configuration = site.siteconfiguration
+        self.assertEqual(site_configuration.lms_url_root, self.site_configuration.lms_url_root)
+        self.assertEqual(site_configuration.platform_name, self.site_configuration.platform_name)
+        self.assertEqual(site_configuration.catalog_api_url, self.site_configuration.catalog_api_url)
+        self.assertEqual(site_configuration.tos_url, self.site_configuration.tos_url)
+        self.assertEqual(site_configuration.privacy_policy_url, self.site_configuration.privacy_policy_url)
+        self.assertEqual(site_configuration.homepage_url, self.site_configuration.homepage_url)
+        self.assertEqual(site_configuration.company_name, self.site_configuration.company_name)
+        self.assertEqual(site_configuration.verified_certificate_url, self.site_configuration.verified_certificate_url)
+        self.assertEqual(site_configuration.certificate_help_url, self.site_configuration.certificate_help_url)
```
vkaracic: What do you think about:
```
attrs = [attr for attr in site.siteconfiguration if not attr.startswith('_')]
for attr in attrs:
  self.assertEqual(getattr(site.siteconfiguration, attr), getattr(self.site_configuration, attr))
```

Also if you're only testing the site configuration why aren't you accepting the ``siteconfiguration`` object instead?

credentials/apps/core/tests/test_commands.py
```diff
@@ -0,0 +1,129 @@
...
+         '--catalog-api-url=fake.catalog.server', '--site-name=Fake Site Name', '--tos-url=fake.tos.url'
+         '--privacy-policy-url=fake.privacy.policy', '--homepage-url=fake.home.page'],
+        ['--site-id=1', '--site-domain=fake.server', '--platform-name=Fake Name', '--lms-url-root=fake.lms.server'
+         '--catalog-api-url=fake.catalog.server', '--site-name=Fake Site Name', '--tos-url=fake.tos.url'
+         '--privacy-policy-url=fake.privacy.policy', '--homepage-url=fake.home.page', '--company-name=Fake Company'],
+        ['--site-id=1', '--site-domain=fake.server', '--platform-name=Fake Name', '--lms-url-root=fake.lms.server'
+         '--catalog-api-url=fake.catalog.server', '--site-name=Fake Site Name', '--tos-url=fake.tos.url'
+         '--privacy-policy-url=fake.privacy.policy', '--homepage-url=fake.home.page', '--company-name=Fake Company'
+         '--verified-certificate-url=fake.verified.url'],
+        ['--site-id=1', '--site-domain=fake.server', '--platform-name=Fake Name', '--lms-url-root=fake.lms.server'
```
vkaracic: This one is a duplicate.

vkaracic: No it's not, my bad.

credentials/apps/core/tests/test_commands.py
marjev: I think the idea is to build and not create `SiteConfiguration` object. We don't want to save the data in the database, but simply use the model to contain related data for later usage.

credentials/apps/core/tests/test_commands.py
marjev: SiteConfiguration object is not iterable.

vkaracic: ``attrs = [attr for attr in site.siteconfiguration.__dict__ if not attr.startswith('_')]``
should work.

marjev: Site ID won't match.

credentials/apps/core/tests/test_commands.py
vkaracic: Alright.

credentials/apps/core/tests/test_commands.py
vkaracic: If you're fine with not covering an object attribute in your test to save a db commit then keep it as is. I'd still change the method parameter to accept site configuration directly, though.

marjev: I did that.

vkaracic: Sorry, I was looking at the wrong commit.

credentials/apps/core/management/commands/create_or_update_site.py
clintonb: Remove the default since this value is required.

credentials/apps/core/management/commands/create_or_update_site.py
```diff
@@ -0,0 +1,153 @@
...
+            if site_created:
+                logger.info('Created Site [%d] with domain [%s]', site.id, site.domain)
+
+        site.domain = site_domain
+        site.name = site_name
+        site.save()
+
+        SiteConfiguration.objects.update_or_create(
+            site=site,
+            defaults={
```
clintonb: This is just `options`. It might be worth exploring passing that value instead of making individual variables and essentially recreating `options`.

credentials/apps/core/tests/test_commands.py
```diff
@@ -0,0 +1,128 @@
...
+        super(CreateOrUpdateSiteCommandTests, self).setUp()
+
+        self.site_configuration = SiteConfigurationFactory.build()
+
+    def _call_command(self, site_domain, site_name, site_id=None):
+        """
+        Internal helper method for interacting with the create_or_update_site management command
+        """
+        # Required arguments
+        command_args = [
```
clintonb: You can drop all of the string formatting and simply pass kwargs to `call_command`:

```python
# Build it manually
command_kwargs = {
  'site_name': site_name,
  ...
}

# BEGIN Create automatically
from django.forms.models import model_to_dict
command_kwargs = model_to_dict(self.site_configuration)

# Allow overrides
command_kwargs.update({
  'site_name': site_name,
})

if site_id:
  command_kwargs['site_id'] = site_id
# END Create automatically

call_command(self.command_name, **command_kwargs)
```

clintonb:
Remember to squash your commits and include a ticket number on the last line of your commit message.

credentials/apps/core/management/commands/create_or_update_site.py
marjev: I tried to do it that way, but it seems that `options` have other keys as well.
`settings` is one of the keys, which `SiteConfiguration` doesn't know what to do with and throws the following error: `TypeError: 'settings' is an invalid keyword argument for this function`.

However, I moved all the `get` invocations down to `defaults` argument since most of them are used only there.

credentials/apps/core/tests/test_commands.py
marjev: It seems that the `call_command` expects all the required arguments to be sent as `args` and not `kwargs`. I get the following error:
`CommandError: Error: argument --site-name is required`.

marjev:
@mjfrey @clintonb Latest commit contains the changes I did in the documentation. Please review the change. Thanks!

PR #169, opened 2017-02-15
Upgraded to Django 1.9

clintonb:
ECOM-7093


clintonb:
@edx/ecommerce please review

MatthewPiatetsky:
are all the other requirement updates a direct result of the django 1.9 upgrade or are you just upgrading those packages?

clintonb:
The dependency updates are (a) to ensure the dependencies remain up-to-date and (b) to ensure the dependencies support Django 1.9+.

PR #170, opened 2017-02-15
ECOM-6762 Remove credentials api/v1 and program_id field

MatthewPiatetsky:
@edx/ecommerce 

Tests pass and i've retrieved and created a certificate locally, though I'm not sure if there are things I may be braking and not testing with this code

Also not sure if there is additional code I should be removing. 
views/serializers/filters/permissions/urls are defined in both api and api/v2.
I've removed serializers from the top level, unsure if we want to consolidate any of the other code to only be in the versioned folder


davec-edx:
When removing v1, was it already marked as deprecated? I don't see it.  Do we know for sure there are not services currently referencing it (not just in our code)?

MatthewPiatetsky:
I did see that program_id was marked as deprecated.
@clintonb said "We should appropriately communicate that all users need to upgrade to the current version, run the command to convert program ID to UUID, and then upgrade to the latest version."  
I can't speak to the decision to do this more than that

credentials/apps/credentials/models.py
```diff
@@ -245,8 +245,6 @@ class ProgramCertificate(AbstractCertificate):
     """
 
     program_uuid = models.UUIDField(db_index=True, unique=True, null=True, blank=False, verbose_name=_('Program UUID'))
-    program_id = models.PositiveIntegerField(db_index=True, unique=True, null=True, blank=True,
```
clintonb: The model changes need to occur *after* the removal of all consumers of the `program_id` field. Otherwise, you run the risk of the migrations being run, removing the field, with references (e.g. Django admin) still being made by web code that has not been updated.

Commit/Release Order:
1. Set `null=False` on program_uuid, and remove all references to program_id.
2. Remove the program_id field.

docs/credentials_api.rst
```diff
@@ -57,7 +57,7 @@ To update the status of a user credential, use ``status``.
 
 .. code-block:: json
 
-    url: /api/v1/user_credentials/:id
+    url: /api/v2/user_credentials/:id
```
clintonb: The endpoint is `/api/v2/credentials/`.

docs/credentials_api.rst
```diff
@@ -256,21 +256,21 @@ Get a List of All Credentials for a Program
...
 
 You can filter the returned list of credentials by using
-the ``program_id`` or ``status`` parameters in the query string.
+the ``program_uuid`` or ``status`` parameters in the query string.
 
 **Example Requests**
 
 .. code-block:: bash
 
-    api/v1/program_credentials/?program_id=<program_id>
```
clintonb: This endpoint doesn't exist.

MatthewPiatetsky: so this is correctly removed?

MatthewPiatetsky:
@clintonb I've put program_id back in the models.py  
have I addressed your comments?

credentials/apps/credentials/models.py
```diff
@@ -261,7 +261,7 @@ class ProgramCertificate(AbstractCertificate):
     )
 
     def __str__(self):
-        return 'ProgramCertificate: {uuid}'.format(uuid=(self.program_uuid or self.program_id))
+        return 'ProgramCertificate: {uuid}'.format(uuid=(self.program_uuid))
```
clintonb: No need for the extra parentheses.

docs/credentials_api.rst
clintonb: No. There literally is no `api/v2/program_credentials/` endpoint. There is only `api/v2/credentials/`.

docs/credentials_api.rst
```diff
@@ -305,7 +305,7 @@ the ``program_id`` or ``status`` parameters in the query string.
     }
 
 **Note:**
-If you do not include the ``program_id`` parameter, you receive the following
+If you do not include the ``program_uuid`` parameter, you receive the following
```
clintonb: This note is incorrect. the `program_uuid` parameter is not required.

docs/credentials_api.rst
```diff
@@ -256,21 +256,21 @@ Get a List of All Credentials for a Program
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 To get a list of all credentials that users have earned for a specific program,
-use ``program_credentials``. The query string must include the ``program_id``
+use ``program_credentials``. The query string must include the ``program_uuid``
```
clintonb: This is not correct. This endpoint non longer exists.

docs/credentials_api.rst
```diff
@@ -207,10 +207,10 @@ string.
 
 .. code-block:: bash
 
-    api/v1/course_credentials/?course_id=<course_id>
-    api/v1/course_credentials/?course_id=<course_id>&status=<status>
-    api/v1/course_credentials/?course_id=<course_id>&certificate_type=<certificate_type>
-    api/v1/course_credentials/?course_id=<course_id>&status=<status>&certificate_type=<certificate_type>
+    api/v2/course_credentials/?course_id=<course_id>
```
clintonb: As with programs, this endpoint doesn't exist. This entire section about course credentials should be removed. We will restore it once we have actually implemented course credentials.

docs/credentials_api.rst
MatthewPiatetsky: Ah ok I understand now I see the wrong info in docs around this line, I got confused by the comment only showing the removed line

MatthewPiatetsky:
@clintonb thanks, sorry that I didn't get the docs comments at first.  did I address your comments?

first commit has the ticket number, i will fixup the rest once PR is ready since team said we don't want to fixup during review process

docs/credentials_api.rst
```diff
@@ -57,7 +57,7 @@ To update the status of a user credential, use ``status``.
 
 .. code-block:: json
 
-    url: /api/v1/user_credentials/:id
+    url: /api/v2/credentials/:id
     Content-Type: application/json
     Method: PATCH
```
davec-edx: Is this block still correct?  with 'id' not 'uuid'?

docs/credentials_api.rst
```diff
@@ -98,7 +94,7 @@ To get information about a specific credential for a single user, use ``credenti
 
 .. code-block:: bash
 
-    /api/v1/user_credentials/1
+    /api/v2/credentials/a2810ab0-c084-43de-a9db-fa484fcc82bc
```
davec-edx: line 91 says use 'credential_id', which in the response is an int.  But in the example it is a UUID.

docs/credentials_api.rst
MatthewPiatetsky: good point i think this should be uuid, changed

docs/credentials_api.rst
MatthewPiatetsky: Not sure why that is but updated to say credential uuid

davec-edx: 👍

docs/credentials_api.rst
```diff
@@ -188,124 +182,3 @@ If you do not include the ``username`` parameter, you receive the following
...
-    }
-
-**Note:**
-If you do not include the ``course_id`` parameter, you receive the following
-``status_code=400`` error message:
-
-``A course_id query string parameter is required for filtering user credentials.``
-
-
-Get a List of All Credentials for a Program
```
clintonb: This section is still valid. See https://github.com/edx/credentials/blob/master/credentials/apps/api/v2/filters.py#L8.

clintonb:
@MatthewPiatetsky remember to squash your commits, please.

PR #174, opened 2017-02-21
ECOM-7199 Remove program_id field

MatthewPiatetsky:
@edx/ecommerce


credentials/apps/credentials/models.py
```diff
@@ -245,8 +245,6 @@ class ProgramCertificate(AbstractCertificate):
     """
 
     program_uuid = models.UUIDField(db_index=True, unique=True, null=False, blank=False, verbose_name=_('Program UUID'))
```
davec-edx: I realize this is a previous commit...but there should be help text for the UUID.

davec-edx:
other than my comment, LGTM

credentials/apps/credentials/models.py
MatthewPiatetsky: Did you have something in mind?  
I don't think we always put help text unless it is actually helpful

clintonb: There doesn't need to be help text. The users of this application know what the field is for. If they don't...they probably shouldn't be using the application. That may seem a bit snarky, but it's true.

MatthewPiatetsky:
thanks, I missed that last one, removed

credentials/apps/credentials/models.py
davec-edx: We had it on program_id, maybe the same text....maybe just a comment that if they don't know what the field is, maybe they shouldn't use it :-)

MatthewPiatetsky: I agree with Clinton, id/uuid fields should be self explanatory (and the help text we currently had just said that the old field was deprecated)

PR #175, opened 2017-02-21
Fixed SASS issue

clintonb:
ECOM-7260

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -15,3 +15,7 @@ edx-opaque-keys==0.4
 edx-rest-api-client==1.7.1
 pillow==4.0.0
 pytz==2016.10
+
+# The package django-libsass requires libsass; however, later versions of libsass are stricter regarding SASS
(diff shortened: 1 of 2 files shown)


clintonb:
FYI @edx/ecommerce @edx/devops

requirements/base.txt
```diff
@@ -15,3 +15,7 @@ edx-opaque-keys==0.4
 edx-rest-api-client==1.7.1
 pillow==4.0.0
 pytz==2016.10
+
+# The package django-libsass requires libsass; however, later versions of libsass are stricter regarding SASS
+# validation. We are stuck with the older version until we fix our SASS files to implement RTL properly.
+libsass==0.10.0
```
schenedx: This looks like a band-aid solution. Do we need this solution to make the migration pipeline happy for this IDA?

clintonb: This *is* a Band-Aid. It has been in place for a while, but was removed with the Django 1.9 upgrade. See the comment.

PR #176, opened 2017-02-22
Added view to clear cache

clintonb:
This view will come in handy when we want to clear the cache without relying on a management command or script.

ECOM-7068


clintonb:
@edx/ecommerce please review

davec-edx:
My only comment is that this only gives one option that I can see.  The ticket says to help the user understand their options.  This does provide a mgmt web interface to clearing the cache...is that what the ticket is asking for?

clintonb:
@davec-edX this is the best option I have given that this is not planned, and I need this functionality now.

credentials/templates/management.html
```diff
@@ -0,0 +1,24 @@
...
+    {% for message in messages %}
+      <li{% if message.tags %} class="{{ message.tags }}"{% endif %}>{{ message }}</li>
+    {% endfor %}
+  </ul>
+{% endif %}
+<h1>Caching</h1>
+<form method="post">
+  {% csrf_token %}
+  <input type="hidden" name="action" value="clear_cache">
+  <input type="submit" value="{% trans "Clear cache" %}">
```
rlucioni: The nested quotes aren't a problem here?

credentials/urls.py
```diff
@@ -34,6 +34,7 @@
     url(r'^credentials/', include('credentials.apps.credentials.urls', namespace='credentials')),
     url(r'^health/$', core_views.health, name='health'),
     url(r'^jsi18n/$', 'django.views.i18n.javascript_catalog', ''),
+    url(r'^management/', include('credentials.apps.edx_django_extensions.urls', namespace='management')),
```
rlucioni: Why not to terminate with `$` here and route to the view directly?

credentials/templates/management.html
clintonb: Nope. The inner quotes will only apply to the `{% trans %}` tag.

credentials/urls.py
clintonb: This allows us to handle additional paths in the future without having to touch this file.

PR #179, opened 2017-03-06
[ECOM-7345] User Credential API endpoint should be site aware

marjev:
@mjfrey @vkaracic @iivic This is the PR for making the User Credential API endpoint (WL) site aware.

The only reason this is marked as WIP is the fact that we still don't know to what extent should the endpoint be site-aware. I made it absolutely site-aware, which means that it doesn't matter which user sends the request to the endpoint, but from which Site the request is being sent from.
If we decide to remove the Site awareness for a specific use case, this PR can easily be modified.

JIRA ticket: https://openedx.atlassian.net/browse/ECOM-7345

### DIFF
--- a/credentials/apps/api/v2/tests/test_views.py
+++ b/credentials/apps/api/v2/tests/test_views.py
@@ -10,6 +10,7 @@ from rest_framework.test import APITestCase, APIRequestFactory
 
 from credentials.apps.api.v2.serializers import UserCredentialSerializer, UserCredentialAttributeSerializer
 from credentials.apps.core.tests.factories import UserFactory, USER_PASSWORD
+from credentials.apps.core.tests.mixins import SiteMixin
 from credentials.apps.credentials.models import UserCredential
 from credentials.apps.credentials.tests.factories import (
     UserCredentialFactory, ProgramCertificateFactory, UserCredentialAttributeFactory
@@ -23,7 +24,7 @@ LOGGER_NAME_SERIALIZER = 'credentials.apps.api.v2.serializers'
 # pylint: disable=no-member
 
 @ddt.ddt
-class CredentialViewSetTests(APITestCase):
+class CredentialViewSetTests(SiteMixin, APITestCase):
     list_path = reverse('api:v2:credentials-list')
 
     def setUp(self):
@@ -32,7 +33,7 @@ class CredentialViewSetTests(APITestCase):
 
     def serialize_user_credential(self, user_credential, many=False):
         """ Serialize the given UserCredential object(s). """
-        request = APIRequestFactory().get('/')
+        request = APIRequestFactory(SERVER_NAME=self.site.domain).get('/')
         return UserCredentialSerializer(user_credential, context={'request': request}, many=many).data
 
     def authenticate_user(self, user):
@@ -160,7 +161,11 @@ class CredentialViewSetTests(APITestCase):
 
     def test_destroy(self):
         """ Verify the endpoint does NOT support the DELETE operation. """
-        credential = UserCredentialFactory(status=UserCredential.AWARDED, username=self.user.username)
+        credential = UserCredentialFactory(
+            credential__site=self.site,
+            status=UserCredential.AWARDED,
+            username=self.user.username
+        )
         path = reverse('api:v2:credentials-detail', kwargs={'uuid': credential.uuid})
 
         # Verify users without the view permission are denied access
@@ -177,7 +182,10 @@ class CredentialViewSetTests(APITestCase):
 
     def test_retrieve(self):
         """ Verify the endpoint returns data for a single UserCredential. """
-        credential = UserCredentialFactory(username=self.user.username)
+        credential = UserCredentialFactory(
+            credential__site=self.site,
+            username=self.user.username
+        )
         path = reverse('api:v2:credentials-detail', kwargs={'uuid': credential.uuid})
 
         # Verify users without the view permission are denied access
@@ -205,8 +213,8 @@ class CredentialViewSetTests(APITestCase):
 
     def test_list_status_filtering(self):
         """ Verify the endpoint returns data for all UserCredentials that match the specified status. """
-        awarded = UserCredentialFactory.create_batch(3, status=UserCredential.AWARDED)
-        revoked = UserCredentialFactory.create_batch(3, status=UserCredential.REVOKED)
+        awarded = UserCredentialFactory.create_batch(3, credential__site=self.site, status=UserCredential.AWARDED)
+        revoked = UserCredentialFactory.create_batch(3, credential__site=self.site, status=UserCredential.REVOKED)
 
         self.authenticate_user(self.user)
         self.add_user_permission(self.user, 'view_usercredential')
@@ -224,19 +232,18 @@ class CredentialViewSetTests(APITestCase):
 
     def test_list_username_filtering(self):
         """ Verify the endpoint returns data for all UserCredentials awarded to the user matching the username. """
-
-        UserCredentialFactory.create_batch(3)
+        UserCredentialFactory.create_batch(3, credential__site=self.site)
 
         self.authenticate_user(self.user)
 
         # Users should be able to view their own credentials without additional permissions
         username = self.user.username
-        expected = UserCredentialFactory.create_batch(3, username=username)
+        expected = UserCredentialFactory.create_batch(3, credential__site=self.site, username=username)
         self.assert_list_username_filter_request_succeeds(username, expected)
 
         # Privileged users should be able to view all credentials
         username = 'test_user'  # pylint: disable=redefined-variable-type
-        expected = UserCredentialFactory.create_batch(3, username=username)
+        expected = UserCredentialFactory.create_batch(3, credential__site=self.site, username=username)
         self.add_user_permission(self.user, 'view_usercredential')
 
         self.assert_list_username_filter_request_succeeds(username, expected)
@@ -244,7 +251,7 @@ class CredentialViewSetTests(APITestCase):
     def test_list_program_uuid_filtering(self):
         """ Verify the endpoint returns data for all UserCredentials awarded for the given program. """
         UserCredentialFactory.create_batch(3)
-        program_certificate = ProgramCertificateFactory()
+        program_certificate = ProgramCertificateFactory(site=self.site)
         expected = UserCredentialFactory.create_batch(3, credential=program_certificate)
 
         self.authenticate_user(self.user)
@@ -257,7 +264,10 @@ class CredentialViewSetTests(APITestCase):
     @ddt.data('put', 'patch')
     def test_update(self, method):
         """ Verify the endpoint supports updating the status of a UserCredential, but no other fields. """
-        credential = UserCredentialFactory(username=self.user.username)
+        credential = UserCredentialFactory(
+            credential__site=self.site,
+            username=self.user.username
+        )
         path = reverse('api:v2:credentials-detail', kwargs={'uuid': credential.uuid})
         expected_status = UserCredential.REVOKED
         data = {'status': expected_status}
@@ -273,3 +283,16 @@ class CredentialViewSetTests(APITestCase):
         self.assertEqual(credential.status, expected_status)
         self.assertEqual(response.status_code, 200)
         self.assertEqual(response.data, self.serialize_user_credential(credential))
+
+    def test_site_filtering(self):
+        """ Verify the endpoint only returns credentials linked to a single site. """
+        credential = UserCredentialFactory(credential__site=self.site)
+        UserCredentialFactory()
+
+        self.authenticate_user(self.user)
+        self.add_user_permission(self.user, 'view_usercredential')
+
+        response = self.client.get(self.list_path)
+        self.assertEqual(UserCredential.objects.count(), 2)
+        self.assertEqual(response.data['count'], 1)
+        self.assertEqual(response.data['results'][0], self.serialize_user_credential(credential))
--- a/credentials/apps/api/v2/views.py
+++ b/credentials/apps/api/v2/views.py
@@ -1,5 +1,6 @@
 import logging
 
+from django.db.models import Q
 from rest_framework import viewsets
 from rest_framework.response import Response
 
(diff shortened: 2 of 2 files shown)


credentials/apps/credentials/tests/factories.py
```diff
@@ -51,6 +52,16 @@ class Meta(object):
     download_url = factory.Faker('url')
     uuid = factory.LazyFunction(uuid.uuid4)
 
+    @classmethod
+    def create_site_aware_batch(cls, size, **kwargs):
+        """ Create a batch of UserCredential objects from the same site. """
+        user_credentials = []
+        for _ in range(size):
+            site = kwargs.get('site', Site.objects.get_current())
```
vkaracic: You don't need to get the site in every iteration.

credentials/apps/api/v2/tests/test_views.py
```diff
@@ -225,39 +234,46 @@ def assert_list_username_filter_request_succeeds(self, username, expected):
     def test_list_username_filtering(self):
         """ Verify the endpoint returns data for all UserCredentials awarded to the user matching the username. """
 
-        UserCredentialFactory.create_batch(3)
+        UserCredentialFactory.create_batch(3, credential=self.current_site_program_certificate)
```
vkaracic: Doesn't ``create_site_aware_batch`` do this automatically?

credentials/apps/api/v2/tests/test_views.py
```diff
@@ -273,3 +289,16 @@ def test_update(self, method):
...
         self.assertEqual(response.status_code, 200)
         self.assertEqual(response.data, self.serialize_user_credential(credential))
+
+    def test_get_queryset(self):
+        """ Verify the endpoint is site-aware. """
+        credential = UserCredentialFactory(credential=self.current_site_program_certificate)
+        UserCredentialFactory(credential=ProgramCertificateFactory(site=SiteFactory()))
+
+        self.authenticate_user(self.user)
+        self.add_user_permission(self.user, 'view_usercredential')
```
vkaracic: Doesn't concern your changes, but shouldn't the user be able to view user credentials by default?

credentials/apps/api/v2/views.py
```diff
@@ -15,9 +16,21 @@ class CredentialViewSet(viewsets.ModelViewSet):
...
+        """ Get queryset containing only UserCredentials that belong to the current site.
+
+        UserCredential currently relates to these two models:
+            - ProgramCredential
+            - CourseCredential
+        """
+        site = self.request.site
+        return UserCredential.objects.filter(
+            Q(program_credentials__site=site) |
+            Q(course_credentials__site=site)
```
vkaracic: Nit: this can fit in one line.

credentials/apps/api/v2/tests/test_views.py
marjev: There's a subtle difference.
`create_batch` will create n UserCredentials pointing to the same credential. (1 to many)
`create_site_aware_batch` will create both n `UserCredentials` and `ProgramCertificates`. (1 to 1).

credentials/apps/api/v2/tests/test_views.py
marjev: Only the ones (s)he owns, AFAIK.

marjev:
Comments addressed.

credentials/apps/api/v2/tests/test_views.py
vkaracic: I don't think so: https://github.com/FactoryBoy/factory_boy/blob/v2.8.1/factory/base.py#L626

credentials/apps/api/v2/tests/test_views.py
```diff
@@ -28,6 +29,7 @@ class CredentialViewSetTests(APITestCase):
 
     def setUp(self):
         super(CredentialViewSetTests, self).setUp()
+        self.current_site_program_certificate = ProgramCertificateFactory(site=Site.objects.get_current())
```
clintonb: The factory already associates a `Site` with `ProgramCertificate` instances: https://github.com/edx/credentials/blob/352cb74284e38281977e68ebab9537cdb432703f/credentials/apps/credentials/tests/factories.py#L23. If instances need to be associated with a specific `Site`, create `self.site = SiteFactory()` and use `self.site` where needed.

credentials/apps/api/v2/tests/test_views.py
```diff
@@ -205,8 +214,8 @@ def test_list(self):
 
     def test_list_status_filtering(self):
         """ Verify the endpoint returns data for all UserCredentials that match the specified status. """
-        awarded = UserCredentialFactory.create_batch(3, status=UserCredential.AWARDED)
-        revoked = UserCredentialFactory.create_batch(3, status=UserCredential.REVOKED)
+        awarded = UserCredentialFactory.create_site_aware_batch(3, status=UserCredential.AWARDED)
```
clintonb: The new method is not necessary. Simply pass the site on the `credential__site` kwarg.

credentials/apps/api/v2/tests/test_views.py
clintonb: All you need here is `credential__site`. The specific credential doesn't matter.

credentials/apps/api/v2/tests/test_views.py
```diff
@@ -225,39 +234,46 @@ def assert_list_username_filter_request_succeeds(self, username, expected):
...
         """ Verify the endpoint returns data for all UserCredentials awarded for the given program. """
         UserCredentialFactory.create_batch(3)
-        program_certificate = ProgramCertificateFactory()
-        expected = UserCredentialFactory.create_batch(3, credential=program_certificate)
+        expected = UserCredentialFactory.create_batch(3, credential=self.current_site_program_certificate)
 
         self.authenticate_user(self.user)
         self.add_user_permission(self.user, 'view_usercredential')
 
-        response = self.client.get(self.list_path + '?program_uuid={}'.format(program_certificate.program_uuid))
```
clintonb: Revert, and modify `program_certificate`.

credentials/apps/api/v2/tests/test_views.py
```diff
@@ -225,39 +234,46 @@ def assert_list_username_filter_request_succeeds(self, username, expected):
...
         self.add_user_permission(self.user, 'view_usercredential')
 
         self.assert_list_username_filter_request_succeeds(username, expected)
 
     def test_list_program_uuid_filtering(self):
         """ Verify the endpoint returns data for all UserCredentials awarded for the given program. """
         UserCredentialFactory.create_batch(3)
-        program_certificate = ProgramCertificateFactory()
-        expected = UserCredentialFactory.create_batch(3, credential=program_certificate)
+        expected = UserCredentialFactory.create_batch(3, credential=self.current_site_program_certificate)
```
clintonb: Revert, and modify `program_certificate`.

credentials/apps/api/v2/tests/test_views.py
```diff
@@ -273,3 +289,16 @@ def test_update(self, method):
         self.assertEqual(credential.status, expected_status)
         self.assertEqual(response.status_code, 200)
         self.assertEqual(response.data, self.serialize_user_credential(credential))
+
+    def test_get_queryset(self):
```
clintonb: `test_site_filtering`

credentials/apps/api/v2/tests/test_views.py
```diff
@@ -273,3 +289,16 @@ def test_update(self, method):
...
+        """ Verify the endpoint is site-aware. """
```
clintonb: Verify the endpoint only returns credentials linked to a single site.

credentials/apps/api/v2/tests/test_views.py
```diff
@@ -273,3 +289,16 @@ def test_update(self, method):
...
+        credential = UserCredentialFactory(credential=self.current_site_program_certificate)
+        UserCredentialFactory(credential=ProgramCertificateFactory(site=SiteFactory()))
+
+        self.authenticate_user(self.user)
+        self.add_user_permission(self.user, 'view_usercredential')
+
+        response = self.client.get(self.list_path)
+        self.assertEqual(UserCredential.objects.all().count(), 2)
```
clintonb: No need for `.all()`.

credentials/apps/api/v2/views.py
```diff
@@ -15,9 +16,18 @@ class CredentialViewSet(viewsets.ModelViewSet):
     filter_class = UserCredentialFilter
     lookup_field = 'uuid'
     permission_classes = (UserCredentialPermissions,)
-    queryset = UserCredential.objects.all()
     serializer_class = UserCredentialSerializer
 
+    def get_queryset(self):
+        """ Get queryset containing only UserCredentials that belong to the current site.
```
clintonb: No need for the docstring. The parent method already has one.

credentials/apps/credentials/tests/factories.py
```diff
@@ -51,6 +52,16 @@ class Meta(object):
     download_url = factory.Faker('url')
     uuid = factory.LazyFunction(uuid.uuid4)
 
+    @classmethod
+    def create_site_aware_batch(cls, size, **kwargs):
```
clintonb: This is not needed.

credentials/apps/api/v2/views.py
```diff
@@ -15,9 +16,18 @@ class CredentialViewSet(viewsets.ModelViewSet):
...
+
+        UserCredential currently relates to these two models:
+            - ProgramCredential
+            - CourseCredential
+        """
+        site = self.request.site
+        return UserCredential.objects.filter(Q(program_credentials__site=site) | Q(course_credentials__site=site))
```
clintonb: You need to make use of the super-class:

```python
def get_queryset(self):
    site = self.request.site
    qs = super(CredentialViewSet, self).get_queryset()

    # We have to filter on the explicit credential models because we cannot set a GenericRelation
    # field on the Site model.
    qs = qs.filter(Q(program_credentials__site=site) | Q(course_credentials__site=site))

    return qs
```

credentials/apps/api/v2/views.py
clintonb: Add an inline comment instead: We have to filter on the explicit credential models because we cannot set a GenericRelation field on the Site model.

credentials/apps/api/v2/tests/test_views.py
marjev: I want to use the current `Site` that the `Client` request will be associated with. I tried to use `SiteFactory` but I didn't find a way to specify which `Site` should be used by the middleware.
Do you know of a way to modify the `Client` to accomplish that?
I don't want to mock the request using the `RequestFactory` and call the view directly.

As a short term solution, I'll do the following (similar to what's done in the ecommerce within the `SiteMixin`):
```
Site.objects.all().delete()
self.site = SiteFactory(id=settings.SITE_ID)
```

marjev:
@clintonb, I've addressed your comments. Can you take a look at the 
[`setUp` implementation](https://github.com/edx/credentials/pull/179/files#diff-4ddad45f4c95e4fa7ed3a89da2f04116R33) because I would make sure it's acceptable to do it like that? Thanks!

credentials/apps/api/v2/tests/test_views.py
clintonb: https://github.com/edx/ecommerce/blob/master/ecommerce/tests/mixins.py#L250-L266

clintonb: I would avoid relying on `settings.SITE_ID` if possible. That just seems brittle.

marjev: The problem I'm having is that I don't know how to specify a site created by the `SiteFactory` as the site for middleware to use.
Almost every result of my Google search suggests using the `RequestFactory` for this, but that would mean I'd have to call the view directly as a method and not rely on Django's middleware and routing.

Regarding the URL you provided, I saw that, but do we need `SiteConfiguration` since it's not used in the `CredentialViewSetTests`?

clintonb: The middleware uses the domain name. You only need the `Site` model. Note that the arguments for the `Site` model are prefixes with `site__` since it is a foreign key. See the following lines for the domain:

* https://github.com/edx/ecommerce/blob/master/ecommerce/tests/mixins.py#L252
* https://github.com/edx/ecommerce/blob/master/ecommerce/tests/mixins.py#L264

marjev: I know what the middleware uses once it has the site, but I don't know how to specify a site to be used by the middleware. Removing `settings.SITE_ID` throws `Site DoesNotExist` my way, because of this https://github.com/django/django/blob/master/django/contrib/sites/models.py#L58. Someone specified `SITE_ID` in the `base.py` and it's being used to get the current site for middleware to use.

Also, I was asking why do we need to indirectly create a Site through `SiteConfigurationFactory`, since we're not using `SiteConfiguration` in this `TestCase`. Can't we simply create a site using the `SiteFactory`?

clintonb: You don't *need* to use SiteConfiguration. I was simply pointing to an example. That being said the system is setup such that every Site has a SiteConfiguration. You may run into issues if the view expects a SiteConfiguration and one does not exist.

marjev: I think it's addressed.

credentials/apps/api/v2/tests/test_views.py
```diff
@@ -28,11 +30,22 @@ class CredentialViewSetTests(APITestCase):
 
     def setUp(self):
         super(CredentialViewSetTests, self).setUp()
+
+        # Set the domain used for all test requests
+        domain = 'testserver.fake'
+        self.client = self.client_class(SERVER_NAME=domain)
+
+        Site.objects.all().delete()
```
clintonb: Is this still necessary? If so, please add a comment explaining why.

clintonb:
Remember to squash your commits.

marjev:
@clintonb While I was exploring the credentials repo, I ran into `SiteMixin` in the most expecting place, core app. So, I used that and removed everything from the `setUp` method.

PR #180, opened 2017-03-06
Site-specific templates

vkaracic:
I've added a header and footer template to the site configuration so that each site can configure their own, and rendering of the header, footer and certificate template if they are set, else use the defaults.

I'm working under the assumption that the templates are suppose to be added via the Django admin panel and not custom HTML files. Let me know if that changes, @mjfrey


clintonb:
What's going on here? Templates via Django admin? I've been on vacation. What ticket is this related to? What's the architectural vision?

mjfrey:
@clintonb Let's chat when you get in, we where just completing the work that was already started in order to allow for template overrides for sites.

PR #182, opened 2017-03-06
Updated makefile to only pull reviewed strings from Transifex

BenjiLee:
Updated makefile to only pull reviewed strings from Transifex.

Verification of upcoming commit that will run by the jenkins job. https://github.com/edx/credentials/pull/181

Context:
 
The jenkins job that automatically pulls and merges translations from Transifex, but will create a PR in case it requires manual resolution. https://github.com/edx/credentials/pull/167 . Rather than manually fixing each string (there many of them), this PR changes the pull step to only pull strings that have been reviewed. Since there are no user facing translations in use, the changes https://github.com/edx/credentials/pull/181 should not effect the UX when removing the non-reviewed translations. The process for reviewing translations does not have to be resolved in this PR. By only pulling reviewed strings, there will be less overhead for fixing strings in this step of the process.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,7 +87,7 @@ compile_translations:
 fake_translations: extract_translations dummy_translations compile_translations
 
 pull_translations:
-	tx pull -af
+	tx pull -af --mode reviewed
 
 push_translations:
 	tx push -s


BenjiLee:
@clintonb Are you volunteering :D?

clintonb:
Nope. However, you should work with @edx/open-source-team (or whomever "owns" i18n) to ensure we have a process in place to review translations. Clearly no one has reviewed the Spanish translations for this service.

rlucioni:
@BenjiLee this would most likely fix the problems we see from time to time with broken translations, but I second @clintonb's suggestion. I don't know if anyone currently reviews translations.

BenjiLee:
Strings can be reviewed by Reviewers and Coordinators (and any role above those). The number of these reviewers is dependent on the community in transifex. For es_419, we have 7 combined reviewers and a fairly large contributor base. We could either promote contributors to a reviewing roles to increase the number of reviewers, but that's a separate process that needs to be figured out.

Alternatively, we pay for professional translations on transifex.

BenjiLee:
To reiterate the points:
* This repository should not be accepting unreviewed strings.
* There needs to be some discussion on how we get reviewed strings but that is out of scope of this PR.

clintonb:
@BenjiLee please make these changes to the other services, and the cookiecutter template.

PR #184, opened 2017-03-11
Added isort for import order validation

clintonb:


clintonb:
@edx/ecommerce please review

.isort.cfg
```diff
@@ -0,0 +1,2 @@
+[settings]
+line_length=120
```
rlucioni: Should this be 119, to match our pep8 config? pep8 treats 119 as the max legal line length; 120 is over that limit.

PR #187, opened 2017-03-14
Added support for sharing certificates to LinkedIn and Twitter

clintonb:
ECOM-6578


clintonb:
@edx/ecommerce please review

credentials/static/sass/_components.scss
```diff
@@ -75,6 +75,11 @@
       font-size: font-size(x-large);
       color: palette(grayscale, dark);
 
+      a {
```
AlasdairSwan: Styles should be applied to classes instead of elements.

clintonb: I cannot think of a scenario where we would make an exception to styling `button.action a` in this manner. Adding the class seems redundant on the surface. What am I missing? Also, do we have a style guide I can reference in the future?

AlasdairSwan: @clintonb this is the closest to an edX CSS/Sass guide that I could find https://github.com/edx/ux-pattern-library/wiki/Styleguide:-Sass-&-CSS
It's a pretty common practice to use classes instead of styling elements directly to prevent issues where you want different implementation of the same basic element to differ in various scenarios. Styling elements directly leads to issues of specificity, which are overcome by either overly long/specific selectors or use of ```!important```. I don't think it's redundant to use classes even when there is little UI in a repo because I think of styling elements directly as adding unnecessary tech-debt.

clintonb: Updated.

PR #188, opened 2017-03-14
Docs Cleanup

clintonb:


clintonb:
FYI @edx/ecommerce

rlucioni:
@clintonb looks like docs aren't building automatically in this repo either. Were you able to fix that for edx/ecommerce? If so, this repo could probably use the same treatment.

clintonb:
@rlucioni you fixed it with the webhook. I'll take care of that.

clintonb:
@rlucioni correction: this was due to our not install requirements when building docs.

rlucioni:
@clintonb cool, thanks for fixing this! I should have mentioned that I checked the webhook and saw it was already active.

PR #189, opened 2017-03-15
Corrected RTL SASS

clintonb:
ECOM-6783


clintonb:
@edx/ecommerce please review

PR #191, opened 2017-03-15
Using Webpack to assemble static assets

clintonb:


clintonb:
@edx/ecommerce please review

FYI @edx/fedx-team

PR #192, opened 2017-03-15
Added sharing via Facebook

clintonb:
Includes #193


clintonb:
@edx/ecommerce please review

credentials/apps/core/admin.py
```diff
@@ -23,6 +26,29 @@ class SiteConfigurationAdmin(admin.ModelAdmin):
...
+        }),
+        (_('Social Sharing'), {
+            'fields': (
+                'facebook_app_id', 'twitter_username', 'enable_facebook_sharing', 'enable_linkedin_sharing',
+                'enable_twitter_sharing',
+            )
+        }),
+    )
+
+    def save_model(self, request, obj, form, change):
```
clintonb: FYI @mjfrey @douglashall @mattdrayer this should be incorporated into ecommerce and other services where site caching is an issue.

douglashall: @clintonb This works great in an environment running a single gunicorn work process. In a multiprocess environment, not so much. Consider installing the latest version of edx-django-sites-extensions which provides a more robust solution to this problem (Set a TTL on the Site cache). Note that ecommerce currently uses an older version of edx-django-sites-extensions which does not include the TTL on the Site cache. We should updgrade to the latest version of edx-django-sites-extensions so we can take advantage of the TTL.

clintonb: Using `edx_django_sites_extensions` is a non-starter due to https://github.com/edx/edx-django-sites-extensions/issues/11.

douglashall: Hmm, I think https://github.com/edx/edx-django-sites-extensions/blob/master/README.rst contains the information you are looking for. Let me know if I can help clarify anything.

credentials/settings/base.py
```diff
@@ -43,6 +42,7 @@
     'social.apps.django_app.default',
     'waffle',
     'storages',
+    'webpack_loader',
```
mikedikan: why was this moved to 3rd party?

docs/credentials_admin.rst
```diff
@@ -25,92 +25,21 @@ Each site can have only one configuration.
 
 
 Add Another Site and Site Configuration
------------------------------------------
-
-If you want to add more sites, partners, and site configurations, you can use
-the ``create_or_update_site`` command. The following options are available for
-this command.
```
mikedikan: why was this all removed?

credentials/settings/base.py
clintonb: It's a third-party app.

docs/credentials_admin.rst
clintonb: It's not needed, and will grow stale over time. If you need to know the command args, use `--help`.

credentials/settings/base.py
mikedikan: and it didn't used to be?

clintonb: It always has been. It was put in the wrong place initially.

PR #194, opened 2017-03-16
DO NOT MERGE: Updated API Authentication

clintonb:
- Using JWT authentication class from edx_rest_framework library
- Removed bearer authentication
- Removed unused social auth pipeline code

ECOM-6578


clintonb:
@edx/ecommerce @edx/helio please review

credentials/apps/api/authentication.py
```diff
@@ -1,79 +0,0 @@
...
-from rest_framework.exceptions import AuthenticationFailed
-from rest_framework_jwt.authentication import JSONWebTokenAuthentication
-
-from credentials.apps.core.constants import Role
-from credentials.apps.core.models import User
-
-logger = logging.getLogger(__name__)
-
-
-def _set_user_roles(user, payload):
```
rlucioni: I would double-check that this isn't being used. edx-drf-extensions doesn't appear to provide this same functionality.

rlucioni: If this disappears, where else will we assign roles? If they're not used, it seems like the entire `Role` enum can be removed.

clintonb:
This is not yet ready to be merged. We will need to do a phased rollout:

1. Update service (this repo + edx/configuration) to use the current issuer/secret *and* the new credentials (actually the same credentials we use everywhere else).
2. Update LMS to use the new credentials.
3. Remove the old credentials from configuration.

credentials/apps/api/authentication.py
clintonb: We can manually assign that role. We don't necessarily want/need every administrator being able to issue credentials. The service user's role is set by a migration (although, that should be moved to a management command).

clintonb:
Related to https://github.com/edx/configuration/pull/3665.

PR #195, opened 2017-03-16
Minor fixes

clintonb:


clintonb:
FYI @rlucioni

docs/credentials_admin.rst
```diff
@@ -127,4 +127,5 @@ The following image shows the ``ProgramCertificate`` model in the Django adminis
     :width: 600px
     :alt: Populated program certificate model
 
-If `Use organization name` is checked then organization full name will be displayed on the certificate else short name would be displayed.
+If `Use organization name` is checked then organization full name will be displayed on the certificate else short name
```
rlucioni: Since you're touching this, it should probably be more along the lines of:

> If `Use organization name` is checked, then the organization's full name will be displayed on the certificate. Otherwise, the organization's short name will be used.

Also, here's an updated screenshot you can use: https://i.imgur.com/uZIIUHP.png.

clintonb: Uhh...I've already moved on from this. If you feel strongly about the change, please file a ticket or a PR.

PR #196, opened 2017-03-16
Added Analytics via Segment

clintonb:


clintonb:
@edx/ecommerce please review

credentials/static/js/analytics.js
```diff
@@ -0,0 +1,14 @@
+!function(){var analytics=window.analytics=window.analytics||[];if(!analytics.initialize)if(analytics.invoked)window.console&&console.error&&console.error("Segment snippet included twice.");else{analytics.invoked=!0;analytics.methods=["trackSubmit","trackClick","trackLink","trackForm","pageview","identify","reset","group","track","ready","alias","debug","page","once","off","on"];analytics.factory=function(t){return function(){var e=Array.prototype.slice.call(arguments);e.unshift(t);analytics.push(e);return analytics}};for(var t=0;t<analytics.methods.length;t++){var e=analytics.methods[t];analytics[e]=analytics.factory(e)}analytics.load=function(t){var e=document.createElement("script");e.type="text/javascript";e.async=!0;e.src=("https:"===document.location.protocol?"https://":"http://")+"cdn.segment.com/analytics.js/v1/"+t+"/analytics.min.js";var n=document.getElementsByTagName("script")[0];n.parentNode.insertBefore(e,n)};analytics.SNIPPET_VERSION="4.0.0";
```
rlucioni: @clintonb a thought: what to we gain from using Segment here? What if we were to only include the GA snippet?

clintonb: My preference is to stick with Segment, since we use it everywhere else. The relatively minor bandwidth savings are outweighed by the desire to maintain development consistency.

PR #197, opened 2017-03-16
Removed jQuery

clintonb:
Facebook offers an integration method that does not rely on jQuery. This integration is now used, and jQuery has been removed.

ECOM-6578


clintonb:
@edx/ecommerce please review

clintonb:
Hat tip to @rlucioni for triggering me to investigate asset size and discover https://developers.facebook.com/docs/javascript/quickstart.

PR #198, opened 2017-03-17
Firing analytics events when share buttons clicked

clintonb:
A tracking event is now sent to Segment when a user clicks any of the share buttons. This will allow us to track usage of these buttons.

ECOM-6578

### DIFF
--- a/credentials/static/js/analytics.js
+++ b/credentials/static/js/analytics.js
@@ -1,34 +1,62 @@
+/**
+ * Click event handler.
+ *
+ * Emits a tracking event via window.analytics.track().
+ */
+function handleClick() {
+  'use strict';
+
+  const attributes = this.attributes;
+  const eventName = this.getAttribute('data-track-event');
+  const eventProperties = {};
+  const eventPropertyPrefix = 'data-track-event-property-';
+
+  // Build the properties hash by looking at all attributes with the specified prefix
+  for (let i = 0, len = attributes.length; i < len; i += 1) {
+    const attribute = attributes[i];
+    const attributeName = attribute.name;
+
+    if (attributeName.startsWith(eventPropertyPrefix)) {
+      eventProperties[attributeName.replace(eventPropertyPrefix, '')] = attribute.value;
+    }
+  }
+
+  window.analytics.track(eventName, eventProperties);
+}
+
+/**
+ * Sets up click handlers for all elements with the attribute data-track-type set to "click".
+ */
+function setupClickHandlers() {
+  'use strict';
+
+  const clickElements = document.querySelectorAll('[data-track-type="click"]');
+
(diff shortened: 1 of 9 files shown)


clintonb:
@edx/ecommerce please review

credentials/static/js/analytics.js
```diff
@@ -11,4 +12,23 @@
     }
 
     analytics.page();
+
+    // Setup click handlers
+    for (let clickElement of document.querySelectorAll('[data-track-type="click"]')) {
```
schenedx: We have typescript translation on credential service?

clintonb: Nope. We have ES2017.

schenedx:
No unit tests for those javascript functions?

credentials/static/js/analytics.js
rlucioni: IE 11 does not support ES2017 features (http://kangax.github.io/compat-table/es2016plus/), but it doesn't look like you're using anything here that's new to ES2017. The `let` statement [is supported](https://docs.microsoft.com/en-us/scripting/javascript/reference/let-statement-javascript).

credentials/templates/_actions.html
```diff
@@ -5,24 +5,37 @@
...
         </a>
       </button>
     {% endif %}
 
     {% if enable_linkedin_sharing %}
-      <button title="{% trans 'Add to LinkedIn profile' %}" class="action btn icon-only">
+      <button title="{% trans 'Add to LinkedIn profile' %}" class="action btn icon-only" data-track-type="click"
+              data-track-event="edx.bi.credentials.linkedin_share.attempted"
+              data-track-event-property-category="certificate"
+              data-track-event-property-credential-uuid="{{ user_credential.uuid }}"
```
rlucioni: @clintonb have you been able to confirm that these properties are actually visible to those looking at GA? Historically, attributes beyond label, category, and action haven't been visible (meaning all this info was basically lost).

credentials/static/js/analytics.js
clintonb: We use [Babel](https://babeljs.io/) to transpile ES2017 to browser-compatible JS.

clintonb:
I have pretty much spent the entire day working on JS tests. At this point, they are not running on Travis. Having sunk a day of time on this, I can honestly say it's not worth continuing with that work given the limited usage of JavaScript in this repository. If someone wants to pickup where I left off, he/she can take https://github.com/edx/credentials/tree/js-tests. However, I have no plans to continue attempting to add tests to this PR. The value is extremely low.

clintonb:
@edx/ecommerce please review
@AlasdairSwan please review the ESLint commit.

PR #199, opened 2017-03-17
Simplified Docker Image

clintonb:


clintonb:
FYI @gsong @jmbowman 

Jeremy and I had a discussion about building images without configuration or Ansible, and I decided to do a small prototype. The challenge to configuration-less images is determining how best to configure them! We have a couple options:

1. Environment variables. This is quite tedious. We run into issues with values that need to be structured as arrays/dicts (e.g. `JWT_ISSUERS` (array) and `MEDIA_STORAGE_BACKEND` (dict)).
2. Settings file via volume. I came across this on [a blog post](http://michal.karzynski.pl/blog/2015/04/19/packaging-django-applications-as-docker-container-images/). This is a neat idea for local development, but may pose issues when running in production and/or environments that are a bit more locked down.

Note that I used Alpine Linux as my base in an effort to decrease image size. My image is 308 MB (compressed). The current image is 580 MB (compressed). My image *does not* include an application server—gunicorn, uswsgi—but does include (unfortunately) a MariaDB server due to https://bugs.alpinelinux.org/issues/4768.

clintonb:
FYI @feanil @efagin @robrap @staubina @jbarciauskas

gsong:
@clintonb Re: configuration via envvars. I would strongly suggest looking at [django-configurations](https://django-configurations.readthedocs.io/en/stable/) which deals with array and dict types. And even do some validation, as well as provide defaults.

Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+    nodejs \
+    zlib-dev \
+    libjpeg-turbo-dev
+
+ADD Makefile /code/
+ADD requirements.txt /code/
+ADD package.json /code/
+ADD requirements/ /code/requirements/
+
+RUN make requirements
```
gsong: Maybe not right now, but can we be smarter about this? Perhaps have make targets that correspond to different environments, such as `dev-requirements`, `test-requirements`, `production-requirements`. That way we can pass a build time `ARG` such as `ENV=[dev | test | production]`, then we can just `RUN make ${ENV}-requirements`.

credentials/settings/docker.py
```diff
@@ -0,0 +1,30 @@
...
+        'debug_toolbar',
+    ]
+
+    MIDDLEWARE_CLASSES += (
+        'debug_toolbar.middleware.DebugToolbarMiddleware',
+    )
+
+    DEBUG_TOOLBAR_PATCH_SETTINGS = False
+
+INTERNAL_IPS = ('127.0.0.1',)
```
gsong: Unfortunately this won't work, since we'll be accessing the container using a different IP address. The way to handle this would be to specifically create a network with specified IP subnet in a docker-compose file, then we can use the gateway (`xxx.xxx.xxx.1`) address since that's where the browser traffic will be coming from.

gsong:
@clintonb I know you have limited time, but I wonder what the results might be with `python:2.7-slim` instead.

Dockerfile
clintonb: Makes sense. Today the `requirements` target is what you might consider `dev-requirements`. We `production-requirements` includes what would probably be considered infrastructure (more so than application) requirements such as `MySQL-python` and `newrelic`.

clintonb: Few, if any, of our IDAs actually have production-specific requirements. I can envision a future where the items in `production.txt` are installed as part of either Ansible or the Docker build.

credentials/settings/docker.py
clintonb: Simpler option (since I have yet to explore Docker Compose networks): set this to `['*']`.

clintonb:
@gsong 
https://hub.docker.com/r/clintonb/credentials/tags/

Sizes (compressed, uncompressed)
alpine: 304MB, 902MB
slim: 329MB, 825MB

credentials/settings/docker.py
gsong: That works, as long as we don't carry it forward for production.

gsong:
@clintonb Oh, the results for python slim looks promising as well, in case it's hard to get certain packages installed for Alpine. Thanks for exploring that.

clintonb:
@gsong I encountered no issues with alpine, aside from the oversized mariadb-dev lib.

.dockerignore
```diff
@@ -0,0 +1,95 @@
+*.py[cod]
```
gsong: One potential way of cutting down the docker ignore file is to ignore everything then selective include only stuff you want. I find that syntax much more clear as well. For example:

```
*
!Makefile
!bin
!config
!src
!manage.py
!npm-shrinkwrap.json
!package.json
!pytest.ini
!requirements
```

gsong:
@clintonb Re: Alpine, when I tried with NumPy/SciPy and friends it was a pain. Also pinning to specific versions of node seemed harder than necessary. 🙂

clintonb:
Fortunately, we don't need any complex libraries for the IDAs.

clintonb:
I split out the OS package installation to a separate base image: https://hub.docker.com/r/clintonb/edx-base/tags/. This allows me to solely focus on application requirements in the application's image: https://hub.docker.com/r/clintonb/edx-base/tags/.

This cuts down on build time, at the (slight) disadvantage of potentially having unnecessary packages installed in the final images. For example, not every service needs `libjpeg`. This is already a problem for these images, however, given that I am install `build-base` and `build-essential` rather than the explicit packages needed (e.g. `make`, `gcc`). Between our five IDAs the installed packages are pretty much all the same, so I am not too worried about this collection of packages in a base image getting crazy.

clintonb:
I also connected this branch to Docker Cloud to automatically build the images.

@rlucioni is also using Docker Cloud for https://github.com/rlucioni/courtbot. He builds and tests images for every PR. Merges to master are automatically deployed to Digital Ocean.

clintonb:
Oh, one more thing...our usage of the older version of edx-pattern-library is why the builds are so slow for this repo.

clintonb:
The latest commit extracts most of the settings to environment variables so that the image can be run in any environment. I was able to run a container on Docker Cloud. The remaining challenge is initial setup/provisioning—create database users/tables, run migrations.

At this point, I think I've provided a very good example of an image that can be used anywhere. Besides extracting the remaining settings, we will also want to put an application server (e.g. gunicorn or uwsgi) in front of Django before trying this in a production-like environment.

clintonb:
@gsong I want to move forward with getting this merged, but don't want to repeat your efforts. Are you actively working on any related projects?

gsong:
@clintonb Go ahead with this. I am actively experimenting on another branch. But my approach is so different that there won't be overlaps.

clintonb:
@gsong is your approach one we might consider adopting in the future? If so...the future can be now.

gsong:
@clintonb I definitely hope so. But I'm not sure how much time I would have before I leave for the conference next week. I'll keep you posted on my progress.

clintonb:
Okay. This is ready to go! What's going on, you ask?

1. The repo has its own `Dockerfile` that is built solely to serve via Django's `runserver` command.
2. The Docker image is built with every build, and that image is used for tests.
3. On master branch builds, the built image is pushed to Docker Hub as `edxops/credentials:devstack-slim`.

FYI @edx/learner @edx/docker-devstack-working-group @edx/devops

clintonb:
Builds take about 2 minutes longer (8 min. vs. 2 min.).

robrap:
8 - 2 = ?  ;)

clintonb:
Oh hush! 8 min. vs. 6 min.

.travis.yml
```diff
@@ -3,22 +3,42 @@ language: python
...
+script:
+  - make exec-validate-translations
+  - make exec-quality
+  - make exec-tests
 
 after_success:
   - pip install --upgrade codecov
   - make exec-coverage
   - codecov
+  - if [ "TRAVIS_PULL_REQUEST" == "false" ] && [ "$TRAVIS_BRANCH" == "master" ]; then
```
rlucioni: Consider echoing a message here that will appear in the build log so it's easy to tell if the push succeeded. For example, see [here](https://github.com/rlucioni/craigbot/blob/master/.travis/docker-push.sh). You may also want to leave a note explaining the rationale for building on Travis instead of Docker Hub/Cloud.

Makefile
```diff
@@ -53,30 +53,30 @@ up-dev: ## Bring up services for development
 	docker-compose -f docker-compose.yml -f docker-compose.dev.yml up -d
 
 up-test: ## Bring up services for testing
-	docker-compose -f docker-compose.yml -f docker-compose.test.yml up -d
+	docker-compose -f docker-compose.yml -f docker-compose.travis.yml up -d
 
 exec-requirements: ## Install requirements on a container
```
rlucioni: It looks like this target is no longer used. Can you get rid of it? If so, you can probably get rid of the `SOURCE_VENV` and `SOURCE_NODEENV ` vars, too.

credentials/settings/test.py
```diff
@@ -12,27 +12,15 @@
...
-
 DATABASES = {
     'default': {
-        'ENGINE': 'django.db.backends.sqlite3',
-        'NAME': ':memory:',
-        'USER': '',
-        'PASSWORD': '',
-        'HOST': '',
-        'PORT': '',
+        'ENGINE': os.environ.get('DB_ENGINE', 'django.db.backends.sqlite3'),
```
rlucioni: I tried making the jump from SQLite to MySQL for testing during the hackathon. I saw test failures when I did that (though I can't find a Travis build showing the failures I saw). I'm surprised you're not seeing the same failures now. Could you confirm that MySQL is actually being used when the tests run?

credentials/settings/test.py
```diff
@@ -12,27 +12,15 @@
...
-#         'NAME': os.environ.get('DB_NAME', ':memory:'),
-#         'USER': os.environ.get('DB_USER', ''),
-#         'PASSWORD': os.environ.get('DB_PASSWORD', ''),
-#         'HOST': os.environ.get('DB_HOST', ''),
-#         'PORT': os.environ.get('DB_PORT', ''),
-#         'CONN_MAX_AGE': int(os.environ.get('CONN_MAX_AGE', 0)),
-#     },
-# }
-
 DATABASES = {
```
rlucioni: While you're in here, consider adding the following to use Memcache when running tests:
```py
CACHES = {
    'default': {
        'BACKEND': os.environ.get('CACHE_BACKEND', 'django.core.cache.backends.locmem.LocMemCache'),
        'LOCATION': os.environ.get('CACHE_LOCATION', ''),
    }
}
```
Then you just need to export these environment variables to `docker-compose.travis.yml`:
```
CACHE_BACKEND: "django.core.cache.backends.memcached.MemcachedCache"
CACHE_LOCATION: "memcached:11211"
```

.travis.yml
```diff
@@ -3,22 +3,42 @@ language: python
...
-  - directories:
-    - node_modules
 
-before_install: make up-test
+before_install:
+  - docker --version
+  - sudo apt-get update
+  - sudo apt-get -y -o Dpkg::Options::="--force-confnew" install docker-engine
+  - docker --version
+  - docker-compose build
```
rlucioni: Just to confirm my understanding, this only builds services which define a `build` target?

Dockerfile
```diff
@@ -0,0 +1,24 @@
+# docker build . -t edxops/credentials:devstack-slim
+
+FROM clintonb/edx-base:python-3.5-slim
```
gsong: Is it OK to base an edX image on your image?

clintonb: Nope! I am awaiting new GitHub and Docker Hub repos.

.travis.yml
clintonb: Correct.

credentials/settings/test.py
clintonb: I have confirmed that the tests do run on MySQL.

PR #201, opened 2017-03-22
Updated code to be compatible with Python 3.5+

clintonb:
ECOM-7552


clintonb:
Related: https://github.com/edx/configuration/pull/3776

PR #204, opened 2017-03-30
Updated User API usage

clintonb:
We are now accessing the User API using the site's access token rather than a pre-made JWT. This decreases the number of settings we need to worry about, and decreases the security risk inherent in spreading symmetric keys across multiple services.

LEARNER-406


clintonb:
@edx/ecommerce please review

credentials/settings/base.py
```diff
@@ -214,12 +214,8 @@
 PROGRAMS_CACHE_TTL = 60 * 60
 
 # USER API CONFIGURATION
-# Absolute URL used to get user data from the user api in LMS
-USER_API_URL = 'http://127.0.0.1:8000/api/user/v1/'
```
schenedx: These needs to be removed from the https://github.com/edx/configuration/blob/master/playbooks/roles/credentials/defaults/main.yml#L74, right?

credentials/apps/core/models.py
```diff
@@ -182,6 +198,28 @@ def get_program(self, program_uuid, ignore_cache=False):
...
+        duration of USER_CACHE_TTL (in seconds). Failed API responses will NOT
+        be cached.
+
+        Arguments:
+            username (str): Unique identifier of the user for retrieval
+
+        Returns:
+            dict: Data returned from the User API
+        """
+        cache_key = 'user.api.data.{}'.format(hashlib.md5(username.encode('utf8')).hexdigest())
```
rlucioni: Credentials has no memcached cluster. This will be cached in-memory by each worker. That shouldn't be a problem, but it's worth keeping in mind.

credentials/settings/base.py
clintonb: Yes. That should happen in a second release.

credentials/apps/credentials/tests/test_models.py
```diff
@@ -55,16 +55,16 @@ def test_signatory_file_saving(self):
         """
         image = SimpleUploadedFile('image.jpg', b'file contents!')
         signatory = Signatory.objects.create(name='test name', title='Test Signatory', image=image)
-        self.assertEqual(
-            signatory.image, 'signatories/1/image.jpg'
-        )
+
+        assert signatory.image.name.startswith('signatories/1/image')
```
rlucioni: This will work fine, but keep in mind that we're not using pytest in this repo yet. If this fails, we won't get pytest's detailed introspection.

PR #206, opened 2017-04-05
Resolved most Django 1.10 deprecation warnings

clintonb:
- Updated urls.py to use functions and classes instead of strings to
represent views
- Upgraded django-extensions to a version that avoids SubfieldBase


clintonb:
FYI @edx/ecommerce

PR #210, opened 2017-04-10
Added Theming Support

clintonb:
Views can now be themed by site. Credential renderings are now themed 
by program type, program, course, course run, and seat type.

LEARNER-516


clintonb:
1. `credentials/base.html` now has three themable sections:
* Header
* Certificate
* Footer
2. Two new themes have been created for OpenEdX and edx.org.

clintonb:
@edx/learner @edx/helio please review

clintonb:
FYI @douglashall

credentials/apps/credentials/tests/test_views.py
```diff
@@ -70,41 +73,35 @@ def _render_user_credential(self, use_proper_logo_url=True):
...
         response = self._render_user_credential()
+        self.assertEqual(response.status_code, 200)
 
-        self._assert_user_credential_template_data(response, self.user_credential, certificate_title=certificate_title)
-        self._assert_signatory_data(response, self.signatory_1)
-        self._assert_signatory_data(response, self.signatory_2)
+        expected = {
+            'user_credential': self.user_credential,
+            'user_data': self.MOCK_USER_DATA,
+            # TODO Determine how to test these values
```
rlucioni: Do you intend to come back to resolve this now?

credentials/apps/credentials/views.py
```diff
@@ -34,67 +36,92 @@ def get_context_data(self, **kwargs):
...
+    @cached_property
+    def user_credential(self):
+        return get_object_or_404(
             UserCredential,
-            uuid=kwargs.get('uuid'),
+            uuid=self.kwargs.get('uuid'),
             status=UserCredential.AWARDED
         )
 
+    @cached_property
```
rlucioni: I see you introducing a lot of these cached properties. The "cache" on these only persists for the lifetime of the instantiated class.  Are these methods being called repeatedly within the lifetime of the class? If not, these don't really add anything.

credentials/templates/credentials/base.html
```diff
@@ -1,12 +1,52 @@
 {% extends 'base.html' %}
+{% load i18n %}
 {% load render_bundle from webpack_loader %}
 
 {% block title %}
+  {#  TODO: Determine title#}
```
rlucioni: Is this resolved?

credentials/apps/core/migrations/0012_remove_siteconfiguration_verified_certificate_url.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('core', '0011_auto_20170410_1646'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
```
rlucioni: This field is currently in use in prod. Shouldn't you be deleting the field after your changes to stop using the field are deployed?

credentials/apps/credentials/tests/test_views.py
clintonb: I would like to; but, I'm not holding out hope.

credentials/apps/credentials/views.py
clintonb: It was. Now it's not used. Removed.

credentials/templates/credentials/base.html
clintonb: Fixed.

credentials/apps/core/migrations/0012_remove_siteconfiguration_verified_certificate_url.py
clintonb: Yes. Great catch!

credentials/apps/credentials/tests/test_views.py
rlucioni: Okay. Maybe think twice about merging a commented out block to which we'll never return? I'd either fix it or just delete it and leave a TODO behind. Either way, not a big deal.

mjfrey:
We should probably remove template from AbstractCertificate as it was never implemented.  This implementation will replace any notion of that.  

The model CertificateTemplate probably has no use.

douglashall:
Thanks @clintonb! Can we get this up and running on a sandbox?

clintonb:
@douglashall if you want to run the code use Docker devstack.

douglashall:
@clintonb It would be great if people on the team who do not have the ability to set up devstack were able to see it in action.

douglashall:
@clintonb Could you also provide an overview of how this theming functionality is used?

clintonb:
@douglashall documentation is available at http://open-edx-credentials-guide.readthedocs.io/en/clintonb-theming-v2/theming.html. Everyone has the ability to run Docker devstack. If you really want see this, you're welcome to schedule some time on my calendar.

PR #217, opened 2017-04-13
Use pytest to run tests

rlucioni:
Having been in maintenance mode for the past several years, nose is effectively dead. edX and the Python community at large are moving towards pytest. pytest is actively developed and provides several benefits over nose, including detailed introspection of failing assert statements (no need to use unittest's assert methods), a rich plugin ecosystem, and modular fixtures (which may help move us away from relying on migrations for populating initial test data).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -82,14 +82,14 @@ private.py
 # pyenv
 .python-version
 
+# Visual Studio Code
+.vscode
+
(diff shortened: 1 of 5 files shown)


rlucioni:
Similar to https://github.com/edx/course-discovery/pull/608.

PR #218, opened 2017-04-13
Fix JS linting failure

rlucioni:
@clintonb this doesn't feel right, but it's causing linting failures on master.


rlucioni:
See this example [build failure](https://s3.amazonaws.com/archive.travis-ci.org/jobs/221818254/log.txt).

PR #222, opened 2017-04-14
Unify Docker Compose files used for development and testing

rlucioni:
We now use override files to extend a base Docker Compose file. Environment variables necessary to run tests against MySQL are now exported, too, but getting the tests to pass against MySQL requires more work.


docker-compose.dev.yml
```diff
@@ -0,0 +1,15 @@
+version: "2"
```
clintonb: Why not combine this with the default? I can't envision ever *not* wanting these values.

rlucioni: I was following the example set by https://docs.docker.com/compose/extends/#example-use-case. The idea is that the base Compose file defines just the canonical configuration for the services, while the override files take care of exposing ports, mounting volumes, etc.

We don't want port 18150 forwarded for credentials in all cases. For example, we don't need that exposed when running tests. I'm not sure it's possible to prevent the port from being forwarded via an override file if the base file already exposes it.

gsong: No, there isn't a way to "close" a port in override.

Makefile
```diff
@@ -1,119 +1,120 @@
...
+html_coverage: ## Generate and view HTML coverage report
+	coverage html && open htmlcov/index.html
+
+shell: ## Run a shell on a credentials container
 	docker exec -it credentials env TERM=$(TERM) /edx/app/credentials/devstack.sh open
 
-pkg-devstack:
-	docker build -t credentials:latest -f docker/build/credentials/Dockerfile git://github.com/edx/configuration
+tail: ## View logs from services running in detached mode
+	docker-compose logs --follow
```
rlucioni: @clintonb here's an example of a tail target that can get you logs at the command line while the containers run in detached mode.

docker-compose.dev.yml
rlucioni: Thanks @gsong. @clintonb in light of this, I'm going to leave these files separate for the time being.

PR #226, opened 2017-04-17
Added tests for JavaScript

clintonb:
ECOM-6578


clintonb:
@edx/learner please review

credentials/static/js/test/specs/analytics_spec.js
```diff
@@ -0,0 +1,20 @@
+describe('analytics module', () => {
+  let analyticsTestButton;
+
+  beforeEach(() => {
+    jasmine.getFixtures().fixturesPath = '/base/credentials/static/js/test/fixtures';
+    loadFixtures('analytics.html'); // eslint-disable-line no-undef
```
bjacobel: You can also add `loadFixtures` as a global in your .eslintrc, that's what we do in edx-platform.

PR #229, opened 2017-04-20
Removed unused template models

clintonb:
The template models were never used or implemented. Now that we have support for themes, the design of certificates should come from the theme rather than a database model.

LEARNER-620


clintonb:
@edx/helio please review

PR #236, opened 2017-05-11
Upgrade to Django 1.10.7

rlucioni:
@schenedx FYI. I started poking around to see what it would take to get this service to Django 1.10.7. There are still some test failures left I haven't figured out. I just wanted you to know this was available in case your team is able to pick up [LEARNER-480](https://openedx.atlassian.net/browse/LEARNER-480).


clintonb:
I ran into these same errors during my attempt. It might be worth taking a step back and determining if we actually need the groups being created. I know at least one, Learners, makes no sense.

ahsan-ul-haq:
@rlucioni could you please close this I merged https://github.com/edx/credentials/pull/248 for this functionality.

PR #243, opened 2017-05-21
Replaced npm with yarn

clintonb:
Yarn is faster than npm and, through its use of lock files, ensures that everyone gets the same version of dependencies.


Dockerfile
```diff
@@ -7,14 +7,20 @@ ENV ENABLE_DJANGO_TOOLBAR 1
 
 WORKDIR /edx/app/credentials/credentials
 
+# Yarn key/repo
```
clintonb: This will eventually be moved into the base images at https://github.com/edx/edx-docker-base.

clintonb:
Thoughts on this change, @edx/learner, @edx/fedx-team, @gsong?

Dockerfile
clintonb: https://github.com/edx/edx-docker-base/pull/4

arizzitano:
:+1: yay for yarn!

Dockerfile
clintonb: `yarn` is now in the base image.

PR #244, opened 2017-05-21
Cleanup

clintonb:
- Replaced karma-jasmine-jquery
- Removed unused constant in Makefile


clintonb:
These commits were merged as part of #243.

PR #247, opened 2017-05-23
Permanently enabled API site filtering

clintonb:
The switch supporting the disabling of site filtering has been removed. This service is now fully multi-tenant.

LEARNER-1102


clintonb:
@edx/helio please review

PR #248, opened 2017-05-25
Added migration for upgrading django

ahsan-ul-haq:


awais786:
Is it backward compatible ?

ahsan-ul-haq:
@awais786 this is part of Django upgrade to 1.10 as per written [here](https://docs.djangoproject.com/en/1.11/releases/1.10/#abstractuser-username-max-length-increased-to-150). This only increases the length of charfield.

PR #251, opened 2017-05-31
Upgrade Credentials service to Django 1.11

ahsan-ul-haq:
LEARNER-949
Upgrade credentials service to support django 1.11


credentials/apps/credentials/tests/test_views.py
```diff
@@ -26,7 +26,7 @@ class RenderCredentialViewTests(SiteMixin, TestCase):
     PROGRAM_TYPE = 'Professional Certificate'
 
     def setUp(self):
-        super().setUp()
+        super(RenderCredentialViewTests, self).setUp()
```
clintonb: Why is this necessary?

credentials/settings/test.py
```diff
@@ -11,6 +11,7 @@
 ]
 
 LOGGING = get_logger_config(debug=False, dev_env=True, local_loglevel='DEBUG')
+ALLOWED_HOSTS = ['testserver.fake']
```
clintonb: Why is this necessary?

ahsan-ul-haq: @clintonb according to documentation [here](https://docs.djangoproject.com/en/1.11/ref/settings/#allowed-hosts) unlike previous versions `ALLOWED_HOSTS` is now checked when running tests so if we don't provide this all the view test gives 400 response.

credentials/apps/credentials/tests/test_views.py
ahsan-ul-haq: While running tests I came across error for this which was resolved by this update.

schenedx: @ahsan-ul-haq What error is it? Can you paste the error message here? I wonder if this is the best solution for the error?

ahsan-ul-haq:
```
TypeError: super() takes at least 1 argument (0 given)
```

clintonb: Would you confirm you are using Python 3.5? No arguments should be necessary for Python 3.5.

credentials/settings/test.py
clintonb: Use '*' to avoid issues in the future.

requirements/base.txt
```diff
@@ -1,16 +1,16 @@
...
+django==1.11.1
+django-extensions==1.7.9
+django-filter==1.0.4
 django-rest-swagger[reST]==0.3.10
 django-storages==1.5.2
 django-waffle==0.11.1
 django-webpack-loader==0.4.1
-djangorestframework==3.4.7
-djangorestframework-jwt==1.9.0
+djangorestframework==3.6.3
```
clintonb: This is a jump in two versions of DRF. Is this necessary? If so, you will need to fix the Swagger docs at `/api-docs`. I would much rather see this change in a separate PR.

ahsan-ul-haq:
@schenedx @clintonb please review

clintonb:
Squash the commits.

ahsan-ul-haq:
@clintonb Done

schenedx:
Please verify you changes on staging as soon as possible

Simon Chen
Software Engineer
________________________________

ahsan-ul-haq:
Verified on stage.

PR #283, opened 2017-07-13
Upgrade to DRF 3.6.3

rlucioni:
3.6.3 is tested against Django 1.11 and includes built-in, interactive API documentation. We no longer need django-rest-swagger. The release announcement is at http://www.django-rest-framework.org/topics/3.6-announcement. Extended release notes are at http://www.django-rest-framework.org/topics/release-notes/#36x-series. We should do a more thorough job of documenting our views after this change lands: http://www.django-rest-framework.org/topics/documenting-your-api/#documenting-your-views.

LEARNER-1590

@edx/learner FYI.


schenedx:
@rlucioni 
We have a different ticket for this work in Spartans. It's LEARNER-1581
I am glad you are already working on this. 
@adeelkhan Can you sync with Renzo here to make sure you don't duplicate efforts?

PR #297, opened 2017-09-14
Fix user's credential identifier to None

uzairr:
**JIRA tickets**: [LEARNER-1778](https://openedx.atlassian.net/browse/LEARNER-1778)

Fix tracking id of the user to None

**Description**

If the user is not logged-in,credential service should not emit
a user identifier.In order to do this, tracking id of the user is
changed to 'None' instead of emitting 'credentials-None'

### DIFF
--- a/credentials/templates/_analytics.html
+++ b/credentials/templates/_analytics.html
@@ -6,16 +6,6 @@
         window.edx.segment = {
             key: '{{ segment_key }}'
         };
-
-        {% if user %}
-            window.edx.user = {
-                name: '{{ user.get_full_name }}',
-                email: '{{ user.email }}',
-                // TODO Set this from tracking context when we get around to
-                // adding that field to the Django user model.
-                trackingId: 'credentials-{{ user.id }}'
-            };
-        {% endif %}
     </script>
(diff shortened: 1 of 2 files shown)


credentials/templates/_analytics.html
```diff
@@ -13,7 +13,7 @@
                 email: '{{ user.email }}',
                 // TODO Set this from tracking context when we get around to
                 // adding that field to the Django user model.
-                trackingId: 'credentials-{{ user.id }}'
+                trackingId: 'None'
```
schenedx: I don't think we should have this line. Just remove this line is what we want I believe.

schenedx:
You should completely remove the `identify` call at https://github.com/edx/credentials/blob/master/credentials/static/js/analytics.js#L48

credentials/templates/_analytics.html
```diff
@@ -11,9 +11,6 @@
             window.edx.user = {
```
schenedx: I don't believe we should even define this variable if the only function need this is `analytics.identify` call. Please confirm and remove.

PR #299, opened 2017-09-20
Fixes XSS vulnerability on Credentials.

adeelkhan:
WIP - A translator could inject and execute Javascript code into
credential template via Django translation mechanism.
This fix escapes all template message to remove
this vulnerability.

LEARNER-1928


adeelkhan:
@simon Do i have to commit django.po file changes too?

credentials/apps/credentials/templatetags/scripttag_escape.py
```diff
@@ -0,0 +1,20 @@
+"""
+Template tags and helper functions for escaping script tag.
+"""
+from django import template
+from django.utils.html import escape
+register = template.Library()  # pylint: disable=invalid-name
+
+
+@register.filter(name="htmlescape")
+def html_escape(value):  # pylint: disable=unused-argument
```
schenedx: The file should be named more generally. Say`credentials/apps/credentials/utils/html.py`?

schenedx: If you have the file named `html` like I suggested, please update the function name to just be `escape`

credentials/apps/credentials/templatetags/scripttag_escape.py
```diff
@@ -0,0 +1,20 @@
...
+    escaped_msg = escape(value)
+
+    html_style = {
```
schenedx: is there a way to take in this dictionary as a function parameter?

adeelkhan:
@simon Initially i wrote view test to check if what it returned shouldn't have any script tag, but then i realized it was incorrect as i was manually injecting script tag as translated msg via django.po file and the test would inspect view response for escaped script tag. Ideally i should some how set this up in test setup but after some search I wasn't able to figure out how to set vulnerable translation message in test setup. So after some googling, i made the decision to write test for templatetag/filter instead. So now i test the filter does escaping correctly if used on any template. This indirectly test views as  we have used this filter for all translated messages.

credentials/apps/credentials/templatetags/scripttag_escape.py
adeelkhan: If you mean parameter to a filter function, i think so no.  Value parameter in filter function is the translated text, and we have no way to know what style or is there is a style around translated text. That is why moving style information from translation msg to a filter seems to be a 1-1 mapping.  I guess we can improve on this by making a module level dictionary keeping a mapping for all such styles for custom tags we introduced in translated messages.

schenedx: If that's the case. I think this is acceptable right now.

schenedx:
@adeelkhan Yes, I believe you should add the updated `django.po` files too

adeelkhan:
@schenedx Made suggested changes, please review again. I have not added .po and .mo files present at path credentials/conf/locale/en_US/LC_MESSAGES. I have used this path to test as it seems only changing them, i could inject and test javascript code. I assume the one at credentials/credentials/conf/locale/en/LC_MESSAGES/django.po by some way generates the above ones. Other than this I have added the ones generated by make targets like extract_compilation and fake_translation.

credentials/apps/credentials/templatetags/html.py
```diff
@@ -0,0 +1,20 @@
...
+@register.filter(name="htmlescape")
+def escape(value):
+    escaped_msg = html.escape(value)
+
+    html_style = {
+        'StartSpan': '<span class=\"accomplishment-statement-detail copy\">',
+        'EndSpan': '</span>',
+        'StrongStart': '<strong class=\"accomplishment-recipient\">',
+        'StrongEnd': '</strong>'
+    }
```
Ayub-Khan: I think this is going to apply unnecessary style tags on all the translation.

credentials/templates/_actions.html
```diff
@@ -1,55 +1,57 @@
+{% load html %}
 {% load i18n %}
 
 <div class="message-actions">
   <div class="message-actions-box">
-    <h3 class="sr-only">{% trans 'Print or share your certificate' %}</h3>
+    <h3 class="sr-only">{% trans 'Print or share your certificate' as tmsg %}{{ tmsg | htmlescape }}</h3>
```
Ayub-Khan: For example this after this is translated this is going to have span and strong .

PR #301, opened 2017-09-25
Show sharing bar to authenticated users only.

waheedahmed:
We should only show the top sharing bar and message when the authenticated learner is viewing their own certificate. For anonymous users, we should not display the sharing bar.

LEARNER-2563


waheedahmed:
@edx/learner-spartans Please review this PR.

PR #305, opened 2017-10-18
LEARNER-1934: Translate certificate views/templates

jlajoie:
### Description

[LEARNER-1934](https://openedx.atlassian.net/browse/LEARNER-1934)

I'm not convinced the way this is implemented is the best approach for detecting if a language specific SVG exists and will be experimenting w/ other options. In addition I'm still waiting for the translated assets to pair with this PR so this will go into blocked until those assets are added.

Just a note, what is implemented here for swapping between language specific SVGs will also be implemented in credentials-themes.

FYI: @edx/learner-heart


credentials/apps/credentials/templatetags/file.py
```diff
@@ -0,0 +1,16 @@
+"""
+Template tags and helper functions for escaping script tag.
+"""
+from django import template
+from django.template.loader import get_template
+from django.template import TemplateDoesNotExist
+register = template.Library()
```
amangano-edx: I'm not sure what the convention is, but I think it would be easier to read this if it were a couple lines below the imports.

credentials/apps/credentials_theme_openedx/templates/openedx/credentials/programs/professional-certificate/certificate.html
```diff
@@ -16,7 +17,13 @@
 {% endblock %}
 
 {% block background_logo %}
-  {% include 'openedx/images/example-logo.svg' %}
+    {% with image_url="openedx/images/example-logo-"|add:language_code|add:'.svg' %}
```
BenjiLee: Does this need an alt text?

jlajoie: @BenjiLee it does not, the a11y work is included in the .svg itself from what I can tell

credentials/apps/credentials/templatetags/file.py
jlajoie: Was following what was done in https://github.com/edx/credentials/blob/master/credentials/apps/credentials/templatetags/html.py though happy to change if you think that it'd be a valuable change

credentials/apps/credentials/templatetags/file.py
```diff
@@ -0,0 +1,50 @@
...
+def construct_svg_language_paths(filepath, language, default=settings.LANGUAGE_CODE):
+    """
+    Creates an array containing between two and five strings a guaranteed in order:
+        1. The requested language added before the svg suffix (e.g. filepath-es-419.svg)
+        1. (optional) The requested language two character code added before the svg suffix (e.g. filepath-es.svg)
+        1. (optional) The default language added before the svg suffix (e.g. filepath-en-US.svg)
+        1. (optional) The default language two character code added before the svg suffix (e.g. filepath-en.svg)
+        1. The unmodified requested filepath (e.g. filepath.svg)
+    """
+    paths = [filepath.replace('.svg', '-' + language + '.svg')]
```
amangano-edx: What if this is called with a file path that doesn't end in `.svg`? We should probably raise an error in that case if it is a hard requirement for this API.

credentials/apps/credentials/templatetags/file.py
```diff
@@ -0,0 +1,50 @@
...
+    or if neither the base path requested. Throws a TemplateNotFound exception if none are found.
+
+    The format necessary for filepath should match 'some/path/filename.svg'
+    This will search for 'some/path/filename-**.svg' where ** is the requested/default language tags.
+    """
+    paths = construct_svg_language_paths(filepath, get_language())
+    return select_template(paths).origin.template_name
+
+
+def construct_svg_language_paths(filepath, language, default=settings.LANGUAGE_CODE):
```
amangano-edx: There's a lot of logic in here, it might be good to have tests for some of the different cases.

credentials/apps/credentials/templatetags/file.py
```diff
@@ -0,0 +1,50 @@
...
+def translate_svg_path(filepath):
+    """
+    Uses the filepath parameter to construct a set of potential language suffixed file names, and return
+    the template_name of the first template matching the current language, default language if otherwise,
+    or if neither the base path requested. Throws a TemplateNotFound exception if none are found.
+
+    The format necessary for filepath should match 'some/path/filename.svg'
+    This will search for 'some/path/filename-**.svg' where ** is the requested/default language tags.
+    """
+    paths = construct_svg_language_paths(filepath, get_language())
```
amangano-edx: Is it possible for `get_language()` to ever return `None`? I think @BenjiLee ran into some issues with that in certain contexts, though I'm not sure if it would be a problem here. Should the `construct_svg_language_paths` method handle that case?

credentials/apps/credentials/templatetags/file.py
```diff
@@ -0,0 +1,50 @@
...
+        1. (optional) The requested language two character code added before the svg suffix (e.g. filepath-es.svg)
+        1. (optional) The default language added before the svg suffix (e.g. filepath-en-US.svg)
+        1. (optional) The default language two character code added before the svg suffix (e.g. filepath-en.svg)
+        1. The unmodified requested filepath (e.g. filepath.svg)
+    """
+    paths = [filepath.replace('.svg', '-' + language + '.svg')]
+
+    # If the requested language is longer than just the two character language code, add the 2 char substring as well
+    if len(language) > 2:
+        paths.append(filepath.replace('.svg', '-' + language[:2] + '.svg'))
```
amangano-edx: Would it be possible to take advantage of something like https://docs.python.org/2/library/os.path.html#os.path.splitext to make this method more generic?

credentials/apps/credentials_theme_openedx/templates/openedx/credentials/programs/professional-certificate/certificate.html
```diff
@@ -1,4 +1,5 @@
 {% extends 'credentials/programs/base.html' %}
+{% load file %}
```
amangano-edx: Nit: Could we name this something a little more descriptive. Maybe something like `i18nAssets` or something?

credentials/apps/credentials/templatetags/file.py
jlajoie: Great suggestion :+1: will add that in

credentials/apps/credentials/templatetags/file.py
jlajoie: I don't think it ever would, though it probably doesn't hurt to add the None check before adding the default case.

credentials/apps/credentials/templatetags/file.py
jlajoie: Hmm I don't see why not, definitely will swap to this

credentials/apps/credentials/templatetags/i18n_assets.py
```diff
@@ -0,0 +1,54 @@
...
+    or if neither the base path requested. Throws a TemplateNotFound exception if none are found.
+
+    The format necessary for filepath should match 'some/path/filename.svg'
+    This will search for 'some/path/filename-**.svg' where ** is the requested/default language tags.
+    """
+    paths = construct_file_language_paths(filepath, get_language())
+    return select_template(paths).origin.template_name
+
+
+def construct_file_language_paths(filepath, language, default=settings.LANGUAGE_CODE):
```
bderusha: NIT: unless django convention dictates otherwise I would prefer to have the method that does the "work" here be called `construct_file_language_name` or something.  Since that is what we are actually alter as opposed to other parts of the path.

credentials/apps/credentials_theme_openedx/templates/openedx/credentials/programs/professional-certificate/certificate.html
```diff
@@ -16,7 +17,8 @@
 {% endblock %}
 
 {% block background_logo %}
-  {% include 'openedx/images/example-logo.svg' %}
+    <div class="{{ 'openedx/images/example-logo.svg'|translate_file_path }}"></div>
```
bderusha: Why is the class name == the filepath name?

jlajoie: Oh whoops that snuck through and isn't supposed to be here, good catch!

credentials/apps/credentials/templatetags/i18n_assets.py
jlajoie: Works for me :+1:

credentials/apps/credentials/templatetags/i18n_assets.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+    paths = construct_file_language_names(filepath, get_language())
+    return select_template(paths).origin.template_name
+
+
+def construct_file_language_names(filepath, language, default=settings.LANGUAGE_CODE):
+    """
+    Creates an array containing between two and five strings a guaranteed in order:
+        1. The requested language added before the svg suffix (e.g. filepath-es-419.svg)
+        1. (optional) The requested language two character code added before the svg suffix (e.g. filepath-es.svg)
```
bderusha: prefer 1,2,3,4,5 over 1,1,1,1,1

jlajoie: :| yup, copy paste failure

credentials/apps/credentials/templatetags/i18n_assets.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+    Uses the filepath parameter to construct a set of potential language suffixed file names, and return
+    the template_name of the first template matching the current language, default language if otherwise,
+    or if neither the base path requested. Throws a TemplateNotFound exception if none are found.
+
+    The format necessary for filepath should match 'some/path/filename.svg'
+    This will search for 'some/path/filename-**.svg' where ** is the requested/default language tags.
+    """
+    paths = construct_file_language_names(filepath, get_language())
+    return select_template(paths).origin.template_name
```
amangano-edx: Can `.origin` ever return `None`?

credentials/apps/credentials/templatetags/i18n_assets.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def construct_file_language_names(filepath, language, default=settings.LANGUAGE_CODE):
+    """
+    Creates an array containing between two and five strings a guaranteed in order:
```
amangano-edx: "five strings in a guaranteed order"

credentials/apps/credentials/templatetags/i18n_assets.py
```diff
@@ -0,0 +1,56 @@
...
+        1. The requested language added before the svg suffix (e.g. filepath-es-419.svg)
```
amangano-edx: Does `get_language()` return 'es_419' or 'es-419'?

credentials/apps/credentials/templatetags/i18n_assets.py
```diff
@@ -0,0 +1,56 @@
...
+        2. (optional) The requested language two character code added before the svg suffix (e.g. filepath-es.svg)
+        3. (optional) The default language added before the svg suffix (e.g. filepath-en-US.svg)
```
amangano-edx: Is the default language code specified as 'en-US' or 'en_US'?

credentials/apps/credentials/templatetags/i18n_assets.py
```diff
@@ -0,0 +1,56 @@
...
+        # If the requested language is longer than just the two character language code, add the 2 char substring
+        if len(language) > 2:
+            paths.append(root + '-' + language[:2] + ext)
+
+    # If the requested language and the default are different, add the default language to the paths
+    if language != default:
+        paths.append(root + '-' + default + ext)
+
+        # If the default language is longer than just the two character language code, add the 2 char substring as well
+        if len(default) > 2:
```
amangano-edx: We should also check if `default[:2]` has already been added to `paths`. That might be the case if the default language is 'en_US' but the default language is 'en'.

credentials/apps/credentials/templatetags/i18n_assets.py
```diff
@@ -0,0 +1,56 @@
...
+def construct_file_language_names(filepath, language, default=settings.LANGUAGE_CODE):
+    """
+    Creates an array containing between two and five strings a guaranteed in order:
+        1. The requested language added before the svg suffix (e.g. filepath-es-419.svg)
+        2. (optional) The requested language two character code added before the svg suffix (e.g. filepath-es.svg)
+        3. (optional) The default language added before the svg suffix (e.g. filepath-en-US.svg)
+        4. (optional) The default language two character code added before the svg suffix (e.g. filepath-en.svg)
+        5. The unmodified requested filepath (e.g. filepath.svg)
+    """
+    root, ext = splitext(filepath)
```
amangano-edx: If filepath doesn't have an extension, what will be returned to `ext`?

credentials/apps/credentials/tests/test_views.py
```diff
@@ -185,3 +186,30 @@ def test_xss_escape_rendered(self):
         )
         rendered_template = template_to_render.render(context)
         self.assertEqual(rendered_template.find('<script>'), -1)
+
+
+class I18nAssetsTemplateTagTest(TestCase):
```
amangano-edx: Is it possible to test the filter itself, maybe using the example-logo?

jlajoie: Will see what I can do, it seemed a bit like just testing the way django itself works so that why I specified only the path methods, but it probably doesn't hurt to try

credentials/apps/credentials/templatetags/i18n_assets.py
jlajoie: It'll just be an empty string which should work fine with our implementation and file naming practices

credentials/apps/credentials/templatetags/i18n_assets.py
jlajoie: As far as I can tell no, it has to do w/ where the template is loaded from, and if a template is found (which will be always unless the exception is thrown, which would be a different problem)

credentials/apps/credentials/templatetags/i18n_assets.py
```diff
@@ -0,0 +1,56 @@
+"""
+Template tags and helper functions for escaping script tag.
```
amangano-edx: I think this may need to be updated.

credentials/apps/credentials/templatetags/i18n_assets.py
```diff
@@ -0,0 +1,56 @@
...
+    paths = construct_file_language_names(filepath, get_language())
+    return select_template(paths).origin.template_name
+
+
+def construct_file_language_names(filepath, language, default=settings.LANGUAGE_CODE):
+    """
+    Creates an array containing between two and five strings in a guaranteed order:
+        1. The requested language added before the svg suffix (e.g. filepath-es-419.svg)
+        2. (optional) The requested language two character code added before the svg suffix (e.g. filepath-es.svg)
+        3. (optional) The default language added before the svg suffix (e.g. filepath-en-US.svg)
```
amangano-edx: Just to confirm, but Django will always return the language code with dashes ('-') and never underscores ('_')? If not, it may be helpful to normalize the paths returned by this method to always use one or the other (ie, call `language.replace('-', '_')` before appending the language code to the paths). That way, we can set a reliable convention for how the language-specific svg files should be named.

jlajoie: Good suggestion, if following the IETF language tag standard it should always be hyphens, though it's such an easy fix I'll just add it in as I've personally seen `en_US` instead of `en-US`

PR #306, opened 2017-10-19
Enabling hours of effort for program certificates

McKenzieW:
Added a setting to selectively enable or disable hours of effort for a given program's certificates
Added hours of effort to the program details stored in the certificate rendering context

[LEARNER-1941](https://openedx.atlassian.net/browse/LEARNER-1941)


McKenzieW:
@edx/learner-heart take a look

credentials/apps/credentials/models.py
```diff
@@ -254,7 +261,8 @@ def program_details(self):
             title=data['title'],
             type=data['type'],
             course_count=len(data['courses']),
-            organizations=organizations
+            organizations=organizations,
+            hours_of_effort=data.get('hours_of_effort')
```
amangano-edx: Where is `hours_of_effort` coming from?

credentials/apps/credentials/models.py
```diff
@@ -223,6 +224,12 @@ class ProgramCertificate(AbstractCertificate):
                     "instead of its short name (e.g. ACMEx)"),
         verbose_name=_('Use organization name')
     )
+    include_hours_of_effort = models.BooleanField(
+        default=False,
+        help_text=_("Display the estimated total number of hours needed to complete all courses in the program. This "
+                    "feature will only be displayed in the certificate if the attribute 'Total hours of effort' has "
+                    "been set for the program in Discovery.")
```
amangano-edx: Please don't translate this.

credentials/apps/credentials/models.py
```diff
@@ -202,7 +202,8 @@ def course_key(self):
 
 OrganizationDetails = namedtuple('OrganizationDetails', ('uuid', 'key', 'name', 'display_name',
                                                          'certificate_logo_image_url'))
-ProgramDetails = namedtuple('ProgramDetails', ('uuid', 'title', 'type', 'course_count', 'organizations'))
+ProgramDetails = namedtuple('ProgramDetails', ('uuid', 'title', 'type', 'course_count',
+                                               'organizations', 'hours_of_effort'))
```
amangano-edx: What is this used for?

credentials/apps/credentials/models.py
McKenzieW: It's going to be added the the Program Model in Discovery

McKenzieW: Should it instead be a Program Certificate attribute?

credentials/apps/credentials/models.py
McKenzieW: I can remove the gettext wrapper, but the reason I translated this new field's help_text is that the preceding field - use_org_name - also has its help_text translated. Should neither of them be translated?

BenjiLee: Would the translation be useful for openedx?

credentials/apps/credentials/models.py
McKenzieW: This data model is used instead of a dictionary to store program details

credentials/apps/credentials/models.py
McKenzieW: I'd imagine that is why the existing field's help_text is translated, so presumably yes.

credentials/apps/credentials/models.py
amangano-edx: No, I think it's better for that data to live in discovery.

McKenzieW:
@amangano-edx is this good to merge?

PR #311, opened 2017-10-25
Render Certificates in Program content language

McKenzieW:
Language is explicitly configured for a particular program's certificates
Template rendering language is set to the configured content_language
[LEARNER-1936](https://openedx.atlassian.net/browse/LEARNER-1936)


McKenzieW:
@edx/learner-heart take a look

credentials/settings/base.py
```diff
@@ -172,6 +173,8 @@
         }
     },
 ]
+# I'm not entirely sure whether this should live here or in the configuration repo
```
jlajoie: I took a look at some of the models in edx-platform, it looks like they just include the settings in the model itself, might be best to just do that?

https://github.com/edx/edx-platform/blob/41862cf163bec5b6ac9cc2d7432ce887740e09f9/common/djangoapps/student/admin.py#L40 for example

credentials/templates/base.html
```diff
@@ -5,7 +5,8 @@
 
 <!DOCTYPE html>
 <html>
-<head lang="{{ language_code }}">
+{% language content_language %}
+<head lang="{% if content_language %}{{content_language}}{% else %}{{language_code}}{% endif %}">
```
jlajoie: I think you could shorten this to `content_language or language_code`

McKenzieW: Would that look like this? 
lang="{{content_language}}{% or %}{{language_code}}"

credentials/apps/credentials/views.py
```diff
@@ -87,6 +87,7 @@ def get_context_data(self, **kwargs):
             'user_credential': user_credential,
             'user_data': user_data,
             'child_templates': self.get_child_templates(),
+            'content_language': user_credential.credential.language,
```
Mjloturco: Can this return null if no language is defined?If so , content_language of None might behave weirdly at https://github.com/edx/credentials/pull/311/files#diff-0ed0d978b3addaff36a2abbdc8fa12cfR8?

credentials/settings/base.py
McKenzieW: Actually what I was wondering was whether the dictionary should be here, in the credentials repo settings folder, or if it should be here, https://github.com/edx/configuration/blob/master/playbooks/roles/credentials/defaults/main.yml

credentials/apps/credentials/views.py
McKenzieW: If no language is defined then the language tag just doesn't do anything

credentials/templates/base.html
Mjloturco: The firstof tag might help here https://github.com/edx/credentials/pull/311/files#diff-0ed0d978b3addaff36a2abbdc8fa12cfR8

credentials/settings/base.py
```diff
@@ -172,6 +173,8 @@
...
+CERTIFICATE_LANGUAGES = {'en': 'English', 'es_419': 'Español'}
```
bderusha: We should default to just english or nothing in the base settings file

credentials/settings/base.py
bderusha: We should add this to the configuration repo (both the generic one as well as our own internal configuration)

credentials/templates/base.html
```diff
@@ -5,7 +5,8 @@
 
 <!DOCTYPE html>
 <html>
-<head lang="{{ language_code }}">
+{% language content_language %}
+<head lang="{% firstof content_language language_code %}">
```
amangano-edx: Why does `<head>` have a lang attribute but `<html>` doesn't? I think `<html>` is likely the better place for it. If children of <html> use a different language, they can always define their own lang attribute.

credentials/templates/base.html
```diff
@@ -5,7 +5,8 @@
 
 <!DOCTYPE html>
 <html>
-<head lang="{{ language_code }}">
+{% language content_language %}
```
amangano-edx: This probably belongs above `<html>`.

credentials/templates/base.html
```diff
@@ -37,5 +38,6 @@
 {% include '_analytics.html' %}
 
 {% block javascript %}{% endblock %}
+{% endlanguage %}
```
amangano-edx: This probably belongs below `</html>`

credentials/templates/base.html
amangano-edx: Also, where does `language_code` come from?

amangano-edx: Will this still result in us showing certificates in the browser-prefered language if we haven't explicitly set a certificate language in the configuration? Is that desirable behavior?

credentials/apps/credentials/forms.py
```diff
@@ -18,6 +20,14 @@ class Meta:
 
 
 class ProgramCertificateAdminForm(forms.ModelForm):
+    def __init__(self, *args, **kwargs):
+        super(ProgramCertificateAdminForm, self).__init__(*args, **kwargs)
+        languages = settings.CERTIFICATE_LANGUAGES.items()
+        lang_choices = sorted(languages, key=itemgetter(1))
```
amangano-edx: What does `itemgetter` do?

credentials/apps/credentials/forms.py
```diff
@@ -18,6 +20,14 @@ class Meta:
 
 
 class ProgramCertificateAdminForm(forms.ModelForm):
+    def __init__(self, *args, **kwargs):
+        super(ProgramCertificateAdminForm, self).__init__(*args, **kwargs)
+        languages = settings.CERTIFICATE_LANGUAGES.items()
```
amangano-edx: Nit: It might be good to just define this as a list of tuples in the configuration rather than as a dictionary if we plan to use it like that.

credentials/apps/credentials/views.py
amangano-edx: Is the behavior of the language tag when no language is provided defined as part of the interface, or does it just happen to work that way? If it's not part of the defined interface, it may be good to guard against passing `None` or `''` to `{% language %}` in case something about the implementation changes in the next version of Django. Maybe we could move the `or` logic from https://github.com/edx/credentials/pull/311/files#diff-0ed0d978b3addaff36a2abbdc8fa12cfR9 to here and then update the template to always expect content_language to be set?

amangano-edx:
Is it possible to test the language rendering

credentials/apps/credentials/views.py
```diff
@@ -87,6 +87,7 @@ def get_context_data(self, **kwargs):
             'user_credential': user_credential,
```
amangano-edx: Would it be possible to add a test for this view to verify that content_language is provided in the template context and that the template renders correctlu?

amangano-edx:
Sorry, I must have mis-clicked something and closed this accidentally.

credentials/templates/base.html
McKenzieW: `language_code` is the browser-preferred language, which I would imagine would be the desired render language if the certificate language hasn't been explicitly set.
I can make this an attribute of `<html>` though instead of `<head>`. I only added the logic here because it was the existing location of the `lang` attribute.

credentials/apps/credentials/forms.py
McKenzieW: It gets the first element of the tuples so as to sort the list by the the language as opposed to the language code

credentials/templates/base.html
amangano-edx: I think for a11y compliance the lang attribute should be on the `<html>` element. 

You should check with Jasper about what we should do if certificate language is not set for a program. I vaguely remember talking about turning off browser-detection for consistency.

credentials/apps/credentials/forms.py
McKenzieW: For the configuration yaml file it is simpler to represent the data as a dictionary than as a list of tuples

McKenzieW:
@bderusha @amangano-edx what do you think?

credentials/apps/credentials/tests/test_views.py
```diff
@@ -164,6 +165,24 @@ def test_logo_missing_exception(self):
         with self.assertRaisesMessage(MissingCertificateLogoError, 'No certificate image logo defined for program'):
             self._render_user_credential(use_proper_logo_url=False)
 
+    @ddt.data(
+        (True, 'Felicitaciones, '),
+        (False, 'Congratulations, ')
+    )
+    @ddt.unpack
+    @responses.activate
+    def test_render_language(self, langauge_set, expected_text):
```
amangano-edx: Will this pass if you do not have a Spanish translation for 'Congratulations'? It might be safer to use the built-in dummy language for testing, `eo`.

amangano-edx: Actually, it might be better to just check that the HTML lang element is set correctly.

credentials/templates/base.html
```diff
@@ -4,8 +4,9 @@
 {% load render_bundle from webpack_loader %}
 
 <!DOCTYPE html>
-<html>
-<head lang="{{ language_code }}">
+{% language language %}
```
amangano-edx: Are there any namespace issues here?

McKenzieW: Because the variable name is the same as the method name? I didn't experience any issues with it in testing, but I can change it to be safe.

credentials/apps/credentials/models.py
```diff
@@ -230,6 +230,11 @@ class ProgramCertificate(AbstractCertificate):
                   "feature will only be displayed in the certificate if the attribute 'Total hours of effort' has "
                   "been set for the program in Discovery."
     )
+    language = models.CharField(
+        max_length=8,
+        null=True,
+        help_text=u'Language in which certificates for this program will be rendered'
```
clintonb: In Python 3.x, everything is Unicode. There is no need for the `u` prefix.

PR #315, opened 2017-10-30
read config file strings as unicode

McKenzieW:
[LEARNER-3052](https://openedx.atlassian.net/browse/LEARNER-3052)


feanil:
@McKenzieW you should be able to bring up a sandbox from this version of credentials to test that it resolves the issue, did you already do that?

clintonb:
@feanil I manually made the change on int-nightly to confirm the fix.

clintonb:
FYI @edx/analytics you may want to make the same change for Insights and the data API.

PR #325, opened 2017-12-20
bump credentials themes version to 0.1.11

bderusha:
related to https://github.com/edx/credentials-themes/pull/26


bderusha:
FYI @edx/learner

PR #327, opened 2017-12-20
Help travis succeed

BenjiLee:
Ruling out the if composer is out of date.


BenjiLee:
Still the same error.

PR #334, opened 2018-01-11
Add docs for i18n/theming

amangano-edx:
JIRA: https://openedx.atlassian.net/browse/LEARNER-3834

### DIFF
--- a/docs/theming.rst
+++ b/docs/theming.rst
@@ -32,7 +32,6 @@ Follow the steps below to add a new theme.
 
 Creating a Theme
 ----------------
-.. _directory tree example:
 .. Generate this with tree (http://mama.indstate.edu/users/ice/tree/), which is available via Homebrew.
 .. code-block:: text
 
@@ -110,8 +109,8 @@ Refer to the `Django i18n docs`_ for more details.
 Translations for custom strings may be provided by including a top-level directory named ``'locale'`` within the theme
 application. The ``'locale'`` directory should contain the compiled translation (.mo) files (produced by running the
 ``django-admin.py compilemessages`` command), and should be structured according to the conventions described in `How
-Django discovers translations`_. The `directory tree example`_ provided above includes a properly structured
-``'locale'`` directory.
+Django discovers translations`_. The directory tree example provided above in the `Creating a Theme`_ section includes
+a properly structured ``'locale'`` directory.
 
 Translations included with the theme application are available to the including application by default. Any conflicts
 between translations provided by the theme application and the including application are resolved according to the
--- a/docs/theming.rst
+++ b/docs/theming.rst
@@ -73,10 +73,10 @@ The tree structure above shows the Open EdX theme with overrides for courses and
 
 There are a few notes that theme creators should keep in mind:
 
-1. All directory names, except ``'LC_MESSAGES'``, should be lower-cased to avoid issues that might occur when
-   developing on case-insensitive environments (e.g. macOS) and deploying to case-sensitive (e.g. Linux). The
-   ``'LC_MESSAGES'`` directory is an exception to this rule, as Django convention requires it to be capitalized. See
-   `How Django discovers translations`_ for more information.
+1. All directory names, except ``LC_MESSAGES``, should be lowercase to avoid issues that might occur when you are
+   developing on case-insensitive environments (e.g. macOS) and deploying to case-sensitive environments (e.g. Linux).
+   The ``LC_MESSAGES`` directory is an exception to this rule, as Django convention requires it to be all uppercase.
+   See `How Django discovers translations`_ for more information.
 2. The rendering view is specifically looking for files named `certificate.html`, `_footer.html`, and `_header.html`.
    If you use different file names for these components, they *will not* be rendered.
 3. Feel free to import your custom CSS in `certificate.html`.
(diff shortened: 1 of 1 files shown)


docs/theming.rst
```diff
@@ -92,3 +101,21 @@ Course certificates can be overridden at the following levels:
...
+application. The ``'locale'`` directory should contain the compiled translation (.mo) files (produced by running the
+``django-admin.py compilemessages`` command), and should be structured according to the conventions described in `How
+Django discovers translations`_. The `directory tree example`_ provided above includes a properly structured
+``'locale'`` directory.
+
+Translations included with the theme application are available to the including application by default. Any conflicts
+between translations provided by the theme application and the including application are resolved according to the
+precedence rules described in `How Django discovers translations`_.
+
+.. _Django i18n docs: https://docs.djangoproject.com/en/1.11/topics/i18n/translation/#internationalization-in-template-code
```
mikix: Oof, a version number in the URL.  It is more correct, but means we have to clean these up as we upgrade.  Is there a non-versioned URL and would we want to use it?  We must do this elsewhere?

docs/theming.rst
```diff
@@ -32,12 +32,19 @@ Follow the steps below to add a new theme.
 
 Creating a Theme
 ----------------
-
+.. _directory tree example:
```
mikix: This isn't working in the github viewer.

docs/theming.rst
amangano-edx: I talked with @catong and we decided it makes sense to include these links as-is. Unfortunately, all Django docs seem to include the version number, so there isn't an ideal solution here.

docs/theming.rst
amangano-edx: Fixed.

docs/theming.rst
```diff
@@ -32,12 +32,18 @@ Follow the steps below to add a new theme.
 
 Creating a Theme
 ----------------
-
 .. Generate this with tree (http://mama.indstate.edu/users/ice/tree/), which is available via Homebrew.
 .. code-block:: text
 
     ../credentials/apps/credentials_theme_openedx/
```
catong: @amangano-edx It looks as if there is an underscore under the word "init" in the filename ```__init__.py``` that is making it less legible than the other items in the tree. Is it intended?

docs/theming.rst
```diff
@@ -67,8 +73,10 @@ The tree structure above shows the Open EdX theme with overrides for courses and
 
 There are a few notes that theme creators should keep in mind:
 
-1. All directory names should be lower-cased to avoid issues that might occur when developing on case-insensitive
-   environments (e.g. macOS) and deploying to case-sensitive (e.g. Linux).
+1. All directory names, except ``'LC_MESSAGES'``, should be lower-cased to avoid issues that might occur when
+   developing on case-insensitive environments (e.g. macOS) and deploying to case-sensitive (e.g. Linux). The
```
catong: Add "environments" after "case-sensitive"

docs/theming.rst
```diff
@@ -67,8 +73,10 @@ The tree structure above shows the Open EdX theme with overrides for courses and
 
 There are a few notes that theme creators should keep in mind:
 
-1. All directory names should be lower-cased to avoid issues that might occur when developing on case-insensitive
-   environments (e.g. macOS) and deploying to case-sensitive (e.g. Linux).
+1. All directory names, except ``'LC_MESSAGES'``, should be lower-cased to avoid issues that might occur when
```
catong: I would make "lowercase" non hyphenated (and use "lowercase" (adjective) rather than "lowercased" (verb)

docs/theming.rst
catong: "when developing on..." -> "when you are developing on..."

docs/theming.rst
```diff
@@ -67,8 +73,10 @@ The tree structure above shows the Open EdX theme with overrides for courses and
...
+   developing on case-insensitive environments (e.g. macOS) and deploying to case-sensitive (e.g. Linux). The
+   ``'LC_MESSAGES'`` directory is an exception to this rule, as Django convention requires it to be capitalized. See
```
catong: "capitalized" -> "all uppercase".

docs/theming.rst
```diff
@@ -92,3 +100,21 @@ Course certificates can be overridden at the following levels:
 * Seat type (e.g. honor, professional, verified)
 * Course (e.g. edX+DemoX)
 * Course run (e.g. course-v1:edX+DemoX+4T2017) + seat type
+
+Internationalization
+~~~~~~~~~~~~~~~~~~~~
+Strings appearing in overridden files may be marked for translation by wrapping them in Django translation functions.
+Refer to the `Django i18n docs`_ for more details.
```
catong: Suggest using something more fully written out for the link text, e.g. "Django internationalization documentation"

docs/theming.rst
```diff
@@ -92,3 +100,21 @@ Course certificates can be overridden at the following levels:
 * Seat type (e.g. honor, professional, verified)
 * Course (e.g. edX+DemoX)
 * Course run (e.g. course-v1:edX+DemoX+4T2017) + seat type
+
+Internationalization
+~~~~~~~~~~~~~~~~~~~~
+Strings appearing in overridden files may be marked for translation by wrapping them in Django translation functions.
```
catong: "appearing" -> "that appear"
"may" -> "can"

docs/theming.rst
```diff
@@ -92,3 +100,21 @@ Course certificates can be overridden at the following levels:
...
+Refer to the `Django i18n docs`_ for more details.
+
+Translations for custom strings may be provided by including a top-level directory named ``'locale'`` within the theme
```
catong: "may" -> "can"
No need for hyphen in "top level"

catong: Is it intended that the straight quotes be part of the directory name? If no, only need the double back quotes for monospace formatting: ``` ``locale`` ```

docs/theming.rst
```diff
@@ -92,3 +100,21 @@ Course certificates can be overridden at the following levels:
...
+application. The ``'locale'`` directory should contain the compiled translation (.mo) files (produced by running the
```
catong: Same comment as above re monospace formatting

docs/theming.rst
amangano-edx: I'm not sure I understand what you're referring to. Are you referring to the fact that the filename begins and ends with two underscores? That is actually part of the python convention that the file is required to follow.

amangano-edx: I think I see what you mean now, in github's diff viewer it looks like the entire word `init` is underlined. That's a bit strange, and I'm not sure why that's happening. If you look at the raw code though, you will see that there are no underscores under `init`.

amangano-edx:
@catong I think I've addressed all of your feedback. Please take another look when you have a chance.

catong:
@amangano-edx 👍

PR #335, opened 2018-01-12
Fix Spanish date translations

mikix:
Django's Spanish date formatting uses uppercase months, but Spanish speakers expect lowercase months. Workaround Django's current behavior and also make sure that translators have more control of how the certificate issued-on phrase is formatted.

https://openedx.atlassian.net/browse/LEARNER-3121

I considered several alternative ways of ending up with the right translation, but I think this is the least invasive and the easiest to undo when/if django fixes itself.  (See LEARNER-3859 for a techdebt bug about removing this lowercase workaround and following django's side of things.)

The capfirst filter is in case some language like Spanish uses non-capital months but reorders the string to put the month first, in which case we still want sentence capitalization.


Makefile
```diff
@@ -19,7 +19,7 @@ production-requirements: ## Install requirements for production
 
 requirements: ## Install requirements for local development
 	npm install
-	pip install -r requirements/local.txt
+	pip install -r requirements/local.txt --user
```
mikix: Unrelated to main change.  Just something to make Linux users' lives easier in future.

credentials/conf/locale/eo/LC_MESSAGES/django.po
```diff
@@ -6,14 +6,14 @@ msgid ""
...
-"POT-Creation-Date: 2017-10-25 16:59+0000\n"
+"POT-Creation-Date: 2018-01-12 19:23+0000\n"
 "PO-Revision-Date: \n"
 "Last-Translator: \n"
 "Language-Team: openedx-translation <openedx-translation@googlegroups.com>\n"
-"Language: eo\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
+"Language: en\n"
```
amangano-edx: For some reason, when this file gets generated it sets Language to 'en'. That is wrong though, it should be 'eo'.

mikix: Hmmmph.  I'mma look at fixing that then.

credentials/templates/credentials/programs/base.html
```diff
@@ -98,7 +98,18 @@ <h4 class="signatory-name">{{ signatory.name }}</h4>
           </li>
           <li class="accomplishment-stamp-date">
             <span class="title">{% block accomplishment_stamp_title %}{% endblock %}</span>
-            <span class="copy-micro emphasized">{% trans "Issued" as tmsg %}{{ tmsg | htmlescape }} {{ user_credential.modified|date:"F Y" }}</span>
+            <span class="copy-micro emphasized">
+              {# Translators: This phrase appears on certificates to show the issue date #}
+              {% trans "Issued {month} {year}" as tmsg %}
```
amangano-edx: `tmsg` should be passed through the `{{ tmsg | htmlescape }}` filter. See https://github.com/edx/credentials/blob/491d97424998373c2b4aa42344ef1fddc71cbf26/credentials/templates/credentials/programs/base.html#L28 for an example.

credentials/templates/credentials/programs/base.html
```diff
@@ -98,7 +98,18 @@ <h4 class="signatory-name">{{ signatory.name }}</h4>
...
+              {# Django's Spanish support has a bug with inappropriately-uppercase months. #}
+              {# We work around it here, but see LEARNER-3859 for more details. #}
+              {% if render_language|slice:':2' == 'es' %}
+                {% interpolate_html tmsg month=user_credential.modified|date:"E"|lower year=user_credential.modified|date:"Y" as tmsg %}
+              {% else %}
+                {% interpolate_html tmsg month=user_credential.modified|date:"E" year=user_credential.modified|date:"Y" as tmsg %}
+              {% endif %}
+              {{ tmsg | capfirst }}
```
amangano-edx: Could we move this formatting logic into the `View` so that we can keep the templates simple?

credentials/templates/credentials/programs/base.html
mikix: interpolate_html does it for us.

credentials/conf/locale/eo/LC_MESSAGES/django.po
mikix: https://github.com/edx/i18n-tools/pull/65

mikix:
Ready for another look I think.

credentials/apps/credentials/templatetags/html.py
```diff
@@ -29,3 +31,25 @@ def interpolate_html(value, **kwargs):
     escaped_msg = html.escape(value)
     safe_interpolated_msg = html.format_html(escaped_msg, **kwargs)
     return safe_interpolated_msg
+
+
+@register.filter(expects_localtime=True, is_safe=False)
+def edxdate(value, fmt):
+    """
+    Provides date formatting with some special edX behavior.
```
amangano-edx: Please clarify the special edX behavior.

credentials/apps/credentials/templatetags/html.py
```diff
@@ -29,3 +31,25 @@ def interpolate_html(value, **kwargs):
...
+
+    Arguments:
+        value (datetime): date to format
+        fmt (string): Django-style date format string
```
amangano-edx: Is it Django-specific, or part of the python standard library?

credentials/apps/credentials/templatetags/html.py
```diff
@@ -29,3 +31,25 @@ def interpolate_html(value, **kwargs):
     escaped_msg = html.escape(value)
     safe_interpolated_msg = html.format_html(escaped_msg, **kwargs)
     return safe_interpolated_msg
+
+
+@register.filter(expects_localtime=True, is_safe=False)
+def edxdate(value, fmt):
```
amangano-edx: Do we have tests for this file anywhere? If so, could we add a couple for this new filter?

credentials/apps/credentials/templatetags/html.py
```diff
@@ -29,3 +31,25 @@ def interpolate_html(value, **kwargs):
...
+        fmt (string): Django-style date format string
+
+    Returns:
+        string: A formatted version of the date
+    """
+    formatted = date(value, fmt)
+
+    # Django's Spanish support has a bug with inappropriately-uppercase months.
+    # We work around it here, but see LEARNER-3859 for more details.
+    if get_language().split('-')[0].lower() == 'es':
```
amangano-edx: Alternatively, you could do:
```
language = get_language()
if language and language[:2] == 'es'
```
This will work if/when get_language returns `None`, or if/when the language code includes an underscore instead of a dash.

credentials/templates/credentials/programs/base.html
```diff
@@ -98,7 +98,12 @@ <h4 class="signatory-name">{{ signatory.name }}</h4>
           </li>
           <li class="accomplishment-stamp-date">
             <span class="title">{% block accomplishment_stamp_title %}{% endblock %}</span>
-            <span class="copy-micro emphasized">{% trans "Issued" as tmsg %}{{ tmsg | htmlescape }} {{ user_credential.modified|date:"F Y" }}</span>
+            <span class="copy-micro emphasized">
+              {# Translators: This phrase appears on certificates to show the issue date #}
+              {% trans "Issued {month} {year}" as tmsg %}
+              {% interpolate_html tmsg month=user_credential.modified|edxdate:"E" year=user_credential.modified|edxdate:"Y" as tmsg %}
+              {{ tmsg | capfirst }}
```
amangano-edx: Why do we need to capfirst filter now?

mikix: If the Spanish translators decide that "Issued {month} {year}" should translate as "{month} de {year} issuo" or something -- we still want that first word to be capitalized, even though it's not a proper noun.

credentials/apps/credentials/templatetags/html.py
mikix: Django specific.  They use different format codes.

credentials/apps/credentials/templatetags/html.py
mikix: Some locale/language codes [1] use three characters at the beginning, not just two.  I don't think any start with 'es' but I didn't want to encode assumptions about the length of the locale code (i.e. I didn't want to match a theoretical 'ese-US' code).

As for dash vs underscore, it seems to be part of Django's get_language API that languages use dashes and locales use underscores (c.f. django.util.translation.to_locale which has a docstring of "Turn a language name (en-us) into a locale name (en_US)").

But worrying about None makes sense, yes. I'll add that check.

[1] See /usr/share/i18n/SUPPORTED on a linux system for a long list of options. And django uses three for some codes (e.g. dsb, udm, ast).

credentials/templates/credentials/programs/base.html
amangano-edx: I think in that case, we can rely on the translators to capitalize the appropriate words, much like we rely on them for month name capitalization/non-capitalization.

credentials/apps/credentials/templatetags/html.py
mikix: Done.

credentials/apps/credentials/templatetags/html.py
mikix: Done.

mikix:
Ready for another look I think. I think I got all the remaining nits.

PR #345, opened 2018-01-29
Update pylint

BenjiLee:


credentials/apps/api/v2/views.py
```diff
@@ -55,27 +55,27 @@ def destroy(self, request, *args, **kwargs):
         serializer = self.get_serializer(instance)
         return Response(serializer.data)
 
-    def list(self, request, *args, **kwargs):
+    def list(self, request, *args, **kwargs):  # pylint: disable=useless-super-delegation
```
mikix: Were these supposed to be disable=arguments-differ instead of super delegation?

PR #351, opened 2018-03-07
Updated travis validate translations

Ayub-Khan:
Updated travis to use the i18n-tools validate
to validate translations in pullrequests.
LEARNER-2556

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,6 +56,9 @@ up-test: ## Bring up services for testing
 exec-validate-translations: ## Check translations on a container
 	docker exec -t credentials bash -c 'make validate_translations'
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -108,4 +111,7 @@ pull_translations: ## Pull translations from Transifex
 detect_changed_source_translations: ## Check if translation files are up-to-date
 	cd credentials && i18n_tool changed
 
+test_translations: ## Test translations files
+	cd credentials && i18n_tool validate
```
amangano-edx: Why not just add this to the `validate_translations` target?

It's a little confusing that we have two separate validation targets. Also, the name `test_translations` doesn't really describe much about what this make target does.

PR #360, opened 2018-04-23
Add initial My Records view

mikix:
This is just a skeleton view, with fake data. Good for testing initial workflow and building upon later.

View requires a logged in user and a 'records' waffle switch to be enabled. This is an in-progress feature.

A11y was tested manually. No bokchoy framework exists currently. Might need to spin up a separate ticket to add that.

You can test this locally by:
* running devstack
* visiting http://localhost:18150/admin/waffle/switch/add/ to add a 'records' active switch
* visiting http://localhost:18150/records/ to see this view


mikix:
Hmm, I forgot to do the responsive mobile version. Let me hack on this a bit more. But it should still be reviewable as-is.

Makefile
```diff
@@ -32,7 +32,7 @@ tests: ## Run tests and generate coverage report
 	coverage report
 	$(NODE_BIN)/gulp test
 
-static: ## Gather all static assets for production (mimized)
+static: ## Gather all static assets for production (minimized)
```
jlajoie: ♥️

credentials/apps/records/tests/test_views.py
```diff
@@ -0,0 +1,78 @@
...
+        return response
+
+    def assert_matching_template_origin(self, actual, expected_template_name):
+        expected = select_template([expected_template_name])
+        self.assertEqual(actual.origin, expected.origin)
+
+    def test_no_anonymous_access(self):
+        """ Verify that the view rejects non-logged-in users. """
+        self.client.logout()
+        response = self._render_records(status_code=302)
```
jlajoie: Should this be a 401 rather than a 302, in that you're unauthorized and need to supply credentials for this page?

credentials/apps/records/views.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class RecordsView(LoginRequiredMixin, TemplateView, ThemeViewMixin):
+    template_name = 'records.html'
+
+    def _get_programs(self):
+        # FIXME: Stop using fake data here, obviously
+        return [
+            {
+                'name': 'Dog Mind Reading',
```
jlajoie: <3

gulpfile.js
```diff
@@ -2,7 +2,7 @@
 
 const eslint = require('gulp-eslint');
 const gulp = require('gulp');
-const jsPath = 'credentials/static/js/**/*.js';
+const jsPath = 'credentials/static/{components,js}/**/*.{js,jsx}';
```
jlajoie: Is this really valid syntax? That's amazing

credentials/apps/records/tests/test_views.py
mikix: Eh, this is the default behavior of Credentials when you go to a page that requires a logged in account. It redirects you to the /login page.

gulpfile.js
mikix: Yeah. You can also use syntax on the command line. Like "ls *.{txt,png}"

PR #364, opened 2018-05-02
Add i18n support for javascript

mikix:
Use i18n_tool to extract the strings. I've tried to consolidate on using it more in a few places as well, with some light hacks around assumptions-that-don't-hold-for-credentials I saw. I've tried to note them with comments here, and links to github issues I created.

Use statici18n to bundle them into js files used in production.

I created a [new resource](https://www.transifex.com/open-edx/edx-platform/credentials-js/) in Transifex (and it has reviewed fr_CA translations already!)

I updated python dependencies, but I did not adopt [OEP-18](http://open-edx-proposals.readthedocs.io/en/latest/oep-0018-bp-python-dependencies.html) in this branch. It wasn't finalized at the time I was doing the work, and I didn't feel like being prematurely ambitious.

https://openedx.atlassian.net/browse/LEARNER-4865


credentials/conf/locale/config.yaml
```diff
@@ -1,5 +1,10 @@
 # Configuration for i18n workflow.
 
+# This will copy each source language to a new directory at the end of the i18n generate step
+# which allows us to migrate to a new locale code without re-creating the Transifex project.
+edx_lang_map:
+    zh_TW: zh_HANT
```
mikix: This is what we do in edx-platform for these locales. Left of zh_HANS for now, since we don't have any translations for it.

Makefile
```diff
@@ -94,24 +97,30 @@ down: ## Bring down all services and remove associated resources
...
-compile_translations: ## Compile translation files, outputting .po files for each supported language
-	python manage.py compilemessages
+compile_translations: ## Compile translation files, outputting .mo files for each supported language
+	cd credentials && PYTHONPATH=.. i18n_tool generate
 
 fake_translations: extract_translations dummy_translations compile_translations ## Generate and compile dummy translation files
 
 pull_translations: ## Pull translations from Transifex
-	tx pull -af --mode reviewed --minimum-perc=1
+	cd credentials && i18n_tool transifex pull
```
mikix: i18n_tool has a couple differences from the old manual tx pull line. One is that it doesn't pass -a, so we only pull the languages listed in config.yaml (which is why I've added a few that were missing).

And it doesn't pass a --minimum-perc at all, so we get 0%-translated files. Which causes problems when we try to validate po files, so I have a slight hack in the next couple lines to compensate. As noted in the comments, i18n_tool should get smarter for us, but this is a decent hack in the meantime.

credentials/static/components/FoldingTable.jsx
```diff
@@ -50,7 +50,7 @@ FoldingTable.propTypes = {
   foldedColumns: PropTypes.arrayOf(PropTypes.shape({
     key: PropTypes.string.isRequired,
     format: PropTypes.string,
-    className: PropTypes.arrayOf(PropTypes.string),
+    className: PropTypes.string,
```
mikix: This is just a minor cleanup -- I had specified the wrong type for this field before. We are passing it a string and using it as a string, so no functional change. Just quieting a run-time warning here.

mikix:
@Ayub-Khan I figured you might have opinions in this space? You don't have to review, but heads up in case you were interested.

.gitignore
```diff
@@ -49,6 +49,7 @@ unittests.xml
 credentials/conf/locale/fake*/LC_MESSAGES/*.po
 credentials/conf/locale/fake*/LC_MESSAGES/*.mo
 credentials/conf/locale/messages.mo
+credentials/conf/locale/zh_HANT/
```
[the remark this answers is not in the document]
mikix: It's generated by i18n_tool. See the bit in config.yaml that maps TW to HANT.

PR #369, opened 2018-05-14
Set HTML lang tag for records pages

mikix:
Make sure that the <html lang=""> tag is set for new Student Records work.

I also cleaned up a few instances of using en-us as the fallback language code. Our locale generated for English is 'en' not 'en-us' and in production, we already configure 'en' as the fallback. So devstack should be in sync too.

### DIFF
--- a/credentials/apps/records/views.py
+++ b/credentials/apps/records/views.py
@@ -36,6 +36,7 @@ class RecordsView(LoginRequiredMixin, TemplateView, ThemeViewMixin):
                 'header': self.select_theme_template(['_header.html']),
             },
             'programs': json.dumps(self._get_programs(), sort_keys=True),
+            'render_language': self.request.LANGUAGE_CODE,
         })
         return context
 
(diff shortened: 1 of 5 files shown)


credentials/templates/base.html
```diff
@@ -18,7 +18,7 @@
       {% render_bundle 'base.style-ltr' 'css' %}
     {% endif %}
 
-    <script src="{% statici18n LANGUAGE_CODE %}"></script>
+    <script src="{% statici18n render_language %}"></script>
```
mikix: To match rest of file that already uses render_language.

credentials/apps/records/views.py
```diff
@@ -36,6 +36,7 @@ def get_context_data(self, **kwargs):
                 'header': self.select_theme_template(['_header.html']),
             },
             'programs': json.dumps(self._get_programs(), sort_keys=True),
+            'render_language': self.request.LANGUAGE_CODE,
```
mikix: Actual change, to make the base template pick up the language to set for the lang attribute.

PR #370, opened 2018-05-14
Convert records waffle switch to a flag

mikix:


credentials/apps/records/constants.py
```diff
@@ -1 +1 @@
-WAFFLE_SWITCH_RECORDS = 'records'
+WAFFLE_FLAG_RECORDS = 'student_records'
```
mikix: Name switch to match LMS a bit closer.

PR #378, opened 2018-05-17
Add rtl fake language

mikix:


mikix:
Wait... no that doesn't make sense. The conversion is still happening on the source strings, not the translated strings... Why are n's and u's not getting flipped right???

PR #382, opened 2018-05-22
Add catalog app to copy Discovery data

mikix:
Adds a catalog django app and accompanying 'copy_catalog' management command. This will query Discovery and add the results to a local copy of the catalog in the local database.

Currently, we're only interested in programs and their courses, and only a small subset of their data. Can be expanded as needed.

I intentionally avoided the word "cache" to try to reduce confusion about what this does.

Our models match what the endpoint gives us, not discovery's internal modeling (a couple differences: course orgs are "owners" not "authoring organizations" and partners aren't exposed). Though I did reconvert course run titles to title_overrides, just to save space (the endpoint always gives full run titles, even when it's the same as the course).

I haven't run this against production to see how long it takes / how performant it is, since production gives me a 403 and I'm not sure how to set that up correctly... (I'm using existing "ask discovery" infrastructure, so it should work in stage/prod, but doesn't locally. not sure how to do the oauth correctly). But regardless, we're not doing anything too insane, and we're paginating and all that. I don't envision terrible problems? We can test in stage once we set up a jenkins job.

https://openedx.atlassian.net/browse/LEARNER-5049


.gitignore
```diff
@@ -39,6 +39,7 @@ nosetests.xml
 unittests.xml
 
 ### Internationalization artifacts
+.transifexrc
```
mikix: Unrelated minor cleanup.

Makefile
```diff
@@ -30,7 +30,7 @@ quality: ## Run linters
 	$(NODE_BIN)/gulp lint
 
 tests: ## Run tests and generate coverage report
-	coverage run -m pytest --durations=25
+	coverage run -m pytest --ds credentials.settings.test --durations=25
```
mikix: Unrelated, but lets this command work in docker, and is a good hint for devs wondering how to make direclty running pytest work.

credentials/apps/catalog/management/commands/copy_catalog.py
```diff
@@ -0,0 +1,53 @@
...
+        for site in Site.objects.all():
+            site_configs = SiteConfiguration.objects.filter(site=site)
+            site_config = site_configs.get() if site_configs.exists() else None
+            if not site_config or not site_config.catalog_api_url:
+                logger.info('Skipping site {}. No configuration.'.format(site.domain))
+                continue
+
+            logger.info('Copying catalog data for site {}'.format(site.domain))
+            client = site_config.catalog_api_client
+            Command.fetch_programs(site, client, page_size=page_size)
```
jlajoie: Maybe a dumb python question but should this be self.fetch_programs?

credentials/apps/catalog/management/commands/copy_catalog.py
```diff
@@ -0,0 +1,53 @@
...
+
+    @staticmethod
+    @transaction.atomic
```
jlajoie: Does this actually need the atomic transaction? parse_program already has it and we shouldn't need it with just the `.programs.get` call

credentials/apps/catalog/management/commands/copy_catalog.py
mikix: Naw cause that method is static. It doesn't need/use self.

credentials/apps/catalog/management/commands/copy_catalog.py
mikix: :+1: dropped

PR #385, opened 2018-05-30
Serialize course creds too

mikix:
Our API previously only serialized program credentials (or more accurately, threw an exception if we had any stored course creds).

This adds support for such credentials if they exist.

It also adjusts the program_uuid filter to return any course credentials for course runs in that program and adds a new type filter to filter on credentials for programs vs course runs.

I've made some choices that deserve an impartial look:
* I've added a 'type' field to the serialization of a credential, so that consumers don't need to test for field existance.
* I've chosen 'course-run' as the course run value for 'type'. Should that be an underscore? The keys for the JSON use underscores...
* I've not included the 'id' for the course-run serialization. I don't know why we'd expose that? I don't know why we would want to expose it for the program either, but we did. I left it in the program serialization, since it's part of the API now. But for new types, seems like a crazily-internal value to expose.
* The 'certificate_url' field for course-run certificates is going to be bogus. But I couldn't for the life of me figure out how to get the correct model instance from inside a field serializer in order to only return a serialization for that for program certs. I figured it wouldn't hurt for now to keep including it.
* I removed translations for the filter labels -- those are developer-facing.

I also need to clean up one spot in edx-platform that assumes we only return program certs. But this can land before/after since we don't have any course certs now anyway. https://github.com/edx/edx-platform/pull/18297

credentials-api


Makefile
```diff
@@ -34,15 +34,17 @@ tests: ## Run tests and generate coverage report
...
 
-static.dev: ## Gather all static assets for development (not minimized)
+static: static-assets## Gather all static assets for production (minimized)
+	$(NODE_BIN)/webpack --config webpack.config.js --display-error-details --progress --optimize-minimize
+
+static.dev: static-assets ## Gather all static assets for development (not minimized)
 	$(NODE_BIN)/webpack --config webpack.config.js --display-error-details --progress
 
-static.watch: ## Gather static assets when they change (not minimized)
+static.watch: static-assets ## Gather static assets when they change (not minimized)
```
mikix: This set of changes is just unrelated cleanup -- all static commands should get the assets going. You need them for the site to work.

credentials/apps/catalog/utils.py
```diff
@@ -77,8 +77,9 @@ def parse_program(site, data):
         org = parse_organization(site, org_data)
         program.authoring_organizations.add(org)
 
+    program.course_runs.clear()
     for course_data in data['courses']:
         _, course_runs = parse_course(site, course_data)
-        program.course_runs = course_runs
+        program.course_runs.add(*course_runs)
```
mikix: This is a patch for the catalog command -- it wasn't setting the course runs for a program correctly, whoops.

credentials/apps/api/v2/serializers.py
```diff
@@ -19,23 +20,39 @@ class CredentialField(serializers.Field):
...
-        if not program_uuid:
-            raise ValidationError('Credential identifier is missing.')
-
-        try:
-            return ProgramCertificate.objects.get(program_uuid=program_uuid, is_active=True)
-        except ObjectDoesNotExist:
-            msg = 'No active ProgramCertificate exists for program [{}]'.format(program_uuid)
-            logger.exception(msg)
-            raise ValidationError({'program_uuid': msg})
+        if program_uuid:
```
jlajoie: Maybe a dumb question, is there ever a case where we'd send along both a program_uuid AND a course_run_id?

credentials/apps/api/v2/tests/test_views.py
```diff
@@ -247,18 +248,52 @@ def test_list_username_filtering(self):
         self.assert_list_username_filter_request_succeeds(username, expected)
 
     def test_list_program_uuid_filtering(self):
-        """ Verify the endpoint returns data for all UserCredentials awarded for the given program. """
+        """ Verify the endpoint returns data for all UserCredentials in the given program. """
+
+        # Course run 1 is in a program, course run 2 is not
+        courserun1 = CourseRunFactory()
```
jlajoie: style nit: 🐍 case maybe?

credentials/apps/api/v2/serializers.py
mikix: Naw, as we discussed in person, this is just deserializing. Not sure we even really use that internally.

mikix:
OK, rebased and fixed the python case nit.  Will merge once tests pass again.

PR #387, opened 2018-05-31
Add a UserGrade table

mikix:
Add a UserGrade table that stores a grade for a unique user&course combo. This is needed so that we can keep a record of verified attempts in a given course.


credentials/apps/core/models.py
```diff
@@ -247,7 +247,7 @@ class Meta(object):
         get_latest_by = 'date_joined'
 
     def __str__(self):
-        return self.get_full_name()
+        return self.username
```
mikix: Otherwise the user field doesn't show up well in the admin list for grades (will be non-unique, plus will be blank for users without full names).

credentials/apps/records/models.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+class UserGrade(TimeStampedModel):
+    """
+    A grade for a specific user and course run
+    """
+    user = models.ForeignKey(User)
+    course_run = models.ForeignKey(CourseRun)
+    letter_grade = models.CharField(max_length=255, blank=False)
+    percent_grade = models.DecimalField(max_digits=5, decimal_places=4, null=False)
```
mikix: In the LMS, the PersistentCourseGrade table uses a FloatField for grades while the CourseOverview uses a DecimalField to store a grade threshold.

DecimalField feels like the more appropriate usage here. But it does mean we have to choose values for decimal_places and such.

I figured might as well follow PersistentCourseGrade's example and store values from 0 to 1 (a true percent). And despite not thinking we need it, allow grades as fine-tuned as 83.21%. The range allowed by this field as written is 0% to 999.99%, which seems fine.

credentials/apps/records/models.py
```diff
@@ -0,0 +1,38 @@
...
+    return [(value,) * 2 for value in values]
+
+
+class UserGrade(TimeStampedModel):
+    """
+    A grade for a specific user and course run
+    """
+    user = models.ForeignKey(User)
+    course_run = models.ForeignKey(CourseRun)
+    letter_grade = models.CharField(max_length=255, blank=False)
```
mikix: In the LMS, the PersistentCourseGrade table allows 255 for this, so we do too here.

credentials/apps/records/models.py
```diff
@@ -0,0 +1,38 @@
...
+    percent_grade = models.DecimalField(max_digits=5, decimal_places=4, null=False)
+    mode = models.CharField(
+        max_length=255,
+        blank=False,
+        choices=_choices(
+            constants.CertificateType.HONOR,
+            constants.CertificateType.PROFESSIONAL,
+            constants.CertificateType.VERIFIED
+        )
+    )
```
mikix: I figure LMS will only send us grades for verified/professional modes today. But I wanted to future-proof this a little bit and not have us write logic on the Credentials side that assumes all data is pre-screened.

That might be a bad impulse though? Maybe we should assume everything is filtered for us and we don't need to store this here? I welcome thoughts.

credentials/apps/core/models.py
jlajoie: Is this used elsewhere for displaying the model by any chance, where it'll now come up as username rather than the user's name? I'm assuming no, since not all users even have this data populated in prod but just wanted to voice the question

mikix: I don't believe in a user-facing way. The other tables in the admin interface mostly store a username already, so they didn't bump against this. I think it's safe.

credentials/apps/records/models.py
jlajoie: I'm on board with this, better to assume mistakes upstream and be explicit than just rely on everyone doing this right thing

PR #392, opened 2018-06-04
Alasdair/learner 5462 move to jest and enzyme for react tests

AlasdairSwan:


package.json
```diff
@@ -45,5 +53,32 @@
     "karma-jasmine-jquery-2": "^0.1.1",
     "karma-spec-reporter": "0.0.30",
     "karma-webpack": "^2.0.9"
+  },
+  "jest": {
```
arizzitano: @AlasdairSwan I looked at this last week and wasn't able to reproduce the path issue locally either. I would guess `make-exec tests` runs differently in the test environment's docker container vs. the local one. One thing you might try is splitting this config out into a `jest.config.js` and using `path` to set the root directory in `setupFiles`, since it seems like `roots` is pretty much getting ignored.

dianakhuang:
The build is green! I repeat: the build is green!

PR #394, opened 2018-06-07
Use bootstrap a little bit

mikix:
Start using bootstrap instead of edx-pattern-library in a few places. We eventually want to get rid of the pattern library altogether. But baby steps.

This branch mostly just changes some base styling and bits of the openedx them to use less pattern-library code. I didn't touch the edx theme in credentials-themes.

I've tested this on mobile (firefox and safari) and desktop (firefox and chromium so far).
I've tested the openedx and edx.org themes. The edx.org one needed no changes, so I don't think we'll break any other custom themes.
I've tested by looking at the example certificate and also confirmed that we didn't break the early-stages records work we have.

I am using an old version of bootstrap (v4 alpha 6, which paragon used to use) because newer versions caused some conflicts with edx-pattern-library that I was having trouble resolving. As we remove more and more of it, hopefully when we're done we can bump the bootstrap version to be more modern. Baby steps.

You can compare and contrast on this sandbox: https://credentials-mikix.sandbox.edx.org/credentials/example/

Or these basic images:
Old:
old
New:
new

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ before_install:
   - docker --version
 
   # Build the image.
-  - docker-compose build > /dev/null
+  - docker-compose build
 
   # Ensure we have a place to store coverage output
   - mkdir -p coverage
--- a/credentials/static/sass/_lib.scss
+++ b/credentials/static/sass/_lib.scss
@@ -7,3 +7,5 @@
 // TODO (CCB): Update edx-pattern-library to get rid of susy.
 @import '~susy/sass/susy';
 @import '~breakpoint-sass/stylesheets/breakpoint';
+
+@import '~bootstrap/scss/bootstrap';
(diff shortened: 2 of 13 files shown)


.travis.yml
```diff
@@ -27,7 +27,7 @@ before_install:
   - docker --version
 
   # Build the image.
-  - docker-compose build > /dev/null
+  - docker-compose build
```
mikix: We both *can* and *need* to make this change here. Now with more work for webpack, this build was taking longer than 10min, which since it didn't output anything, travis kills and fails the build.

I believe we were redirecting output because of all the noise of bourbon deprecations. I've fixed that elsewhere in this branch, so we can see real output here now.

credentials/static/sass/_lib.scss
```diff
@@ -7,3 +7,5 @@
 // TODO (CCB): Update edx-pattern-library to get rid of susy.
 @import '~susy/sass/susy';
```
jlajoie: Can we get rid of susy now?

mikix: *Maybe*, but I'm not confident. Susy has some other stuff in it besides the breakpoint work. I didn't have time to dig into whether we could drop it altogether.

PR #395, opened 2018-06-08
LEARNER-5438 Added ProgramCertRecord table

ColeLRogers:
Added new table to associate program certificate with user and added its migrations.
@mikix I am also adding a migration from the change you made earlier, does this look correct?


credentials/apps/credentials/models.py
```diff
@@ -287,3 +287,19 @@ class UserCredentialAttribute(TimeStampedModel):
...
     class Meta(object):
         unique_together = (('user_credential', 'name'),)
+
+
+class ProgramCertRecord(TimeStampedModel):
+    """
+    Connects a User with a Program Certificate
+    """
+    certificate = models.ForeignKey(ProgramCertificate)
+    user_credential = models.ForeignKey(UserCredential)
```
dianakhuang: As discussed, this should point to the user instead of user credential.

credentials/apps/credentials/models.py
```diff
@@ -287,3 +287,19 @@ class UserCredentialAttribute(TimeStampedModel):
...
+    uuid = models.UUIDField(default=uuid.uuid4, editable=False, unique=True)
+
+    def __str__(self):
+        return 'ProgramCertificateRecord: {uuid}'.format(uuid=self.uuid)
+
+    class Meta(object):
+        unique_together = (('certificate', 'user_credential'),)
+        verbose_name = "Program certificate record"
```
dianakhuang: This verbose name for this could probably be more descriptive. "A viewable record of a program certificate." or something like that.

.isort.cfg
```diff
@@ -1,2 +1,3 @@
 [settings]
 line_length=119
+skip=migrations
```
mikedikan: what is this file?

credentials/apps/core/migrations/0014_auto_20180608_1352.py
```diff
@@ -0,0 +1,21 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('core', '0013_auto_20170525_0947'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='user',
```
mikedikan: why was the user model altered?

.isort.cfg
ColeLRogers: This is the configuration file for isort, which the quality tests use for import statement formatting. The migration files do not use this standard when being auto-generated, so this change allows the tests to ignore those.

credentials/apps/core/migrations/0014_auto_20180608_1352.py
ColeLRogers: This was from a previous change that new migrations weren't made for. I OK'd with @mikix.

.isort.cfg
mikedikan: cool, thanks

credentials/apps/core/migrations/0014_auto_20180608_1352.py
mikedikan: 👍

credentials/apps/credentials/migrations/0013_auto_20180608_1352.py
```diff
@@ -0,0 +1,49 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.11 on 2018-06-08 13:52
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import django.db.models.deletion
+import django_extensions.db.fields
+import uuid
```
dianakhuang: I would recommend deleting these migrations since they're no longer relevant.

credentials/apps/credentials/migrations/0014_auto_20180611_1419.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.11 on 2018-06-11 14:19
+from __future__ import unicode_literals
+
+from django.conf import settings
+from django.db import migrations, models
+import django.db.models.deletion
```
dianakhuang: This one, too.

credentials/apps/records/models.py
```diff
@@ -36,3 +39,19 @@ class UserGrade(TimeStampedModel):
...
+    """
+    certificate = models.ForeignKey(ProgramCertificate)
+    user_credential = models.ForeignKey(User)
+    uuid = models.UUIDField(default=uuid.uuid4, editable=False, unique=True)
+
+    def __str__(self):
+        return 'ProgramCertificateRecord: {uuid}'.format(uuid=self.uuid)
+
+    class Meta(object):
+        unique_together = (('certificate', 'user_credential'),)
```
dianakhuang: I believe `user_credential` needs to be switched to `user` here.

credentials/apps/credentials/migrations/0015_auto_20180611_1511.py
```diff
@@ -0,0 +1,30 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.11 on 2018-06-11 15:11
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
```
dianakhuang: I think this one is an extra one, too.

credentials/apps/records/migrations/0002_auto_20180611_1511.py
```diff
@@ -0,0 +1,39 @@
...
+                ('certificate', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, to='credentials.ProgramCertificate')),
+                ('user_credential', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, to=settings.AUTH_USER_MODEL)),
+            ],
+            options={
+                'verbose_name': 'A viewable record of a program certificate',
+            },
+        ),
+        migrations.AlterUniqueTogether(
+            name='programcertrecord',
+            unique_together=set([('certificate', 'user_credential')]),
```
dianakhuang: It looks like this didn't pick up your fix. I would recommend deleting this migration as well and then regenerating them all.

PR #399, opened 2018-06-12
ProgramRecordCreationView WIP

ColeLRogers:


credentials/apps/records/tests/test_views.py
```diff
@@ -144,3 +146,57 @@ def test_xss(self):
...
+
+    def test_user_creation(self):
+        """Verify successful creation of a ProgramCertRecord and return of a uuid"""
+        rev = reverse('records:cert_creation')
+        data = {'username': self.MOCK_USERNAME, 'program_cert_uuid': self.pc.program_uuid}
+        response = self.client.post(rev, data)
+        json_data = response.json()
+
+        self.assertEqual(response.status_code, 201)
+        self.assertRegex(json_data['uuid'], UUID_PATTERN)  # pylint: disable=deprecated-method
```
dianakhuang: Do we use this `assertRegex` pattern elsewhere? I'm leery of using it if it's being deprecated.

credentials/apps/records/views.py
```diff
@@ -129,3 +134,39 @@ def dispatch(self, request, *args, **kwargs):
...
+        cert_uuid = request.POST.get('program_cert_uuid')
+        certificate = get_object_or_404(ProgramCertificate, program_uuid=cert_uuid)
+
+        # verify that the User has the User Credentials for the Program Certificate
+        try:
+            UserCredential.objects.get(username=username, program_credentials__program_uuid=cert_uuid)
+        except UserCredential.DoesNotExist:
+            return JsonResponse({'error': 'User does not have credentials'}, status=404)
+
+        pcr = ProgramCertRecord.objects.create(user=user, certificate=certificate)
```
dianakhuang: This should fail due to the uniqueness constraints if a `ProgramCertRecord` that already exists attempts to be recreated, but I think it makes sense to simply return the uuid of the existing record instead of throwing an error when this happens. The Django ORM has `get_or_create` which should work here.

Also, a test for calling it multiple times and ensuring that the same uuid gets returned without failure would be useful.

credentials/apps/records/tests/test_views.py
ColeLRogers: It is used in the other two classes in this file, but nowhere else. Attempting to use assertRegexpMatches() (used in edx-platform) results in the same pylint warning. It seems to be an issue with the combination of Python and Django versions we use for this repo, resources online tend to say to switch to the other when this warning arises, but both functions give it here.

PR #400, opened 2018-06-13
Learner 5358

spencerhance:
Ticket: [Finalize FAQ sections](https://openedx.atlassian.net/browse/LEARNER-5358)

Makes the FAQ section match the mockup [here](https://invis.io/XVKN9913HYQ#/300683942_Records-3account-Desktop)

Adds in (optional) site configuration for the records FAQ url in admin.  Doesn't render the FAQ footer if no URL is provided.

Adds additional StringUtils functionality. 

**From PR:**
faq-render

**Mock:**
mock


credentials/apps/core/management/commands/create_or_update_site.py
```diff
@@ -100,6 +100,14 @@ def add_arguments(self, parser):
             required=True,
             help='Certificate Help URL'
         )
+        parser.add_argument(
+            '--records-help-url',
+            action='store',
+            dest='records_help_url',
+            type=str,
+            required=True,
```
mikix: New required attribute, might need special care to avoid breaking any scripts out there. I'm guessing none, but might be worth checking at least our edx/configuration repo?

credentials/apps/core/models.py
```diff
@@ -81,6 +81,12 @@ class SiteConfiguration(models.Model):
         blank=False,
         null=True,
     )
+    records_help_url = models.URLField(
+        verbose_name=_('Student Records Help URL'),
+        help_text=_('URL of page for questions about Student Records'),
```
mikix: I know it seemed obvious to do this, and the surrounding code does this (though it shouldn't), but don't use translated strings for stuff that ends up in django admin - django doesn't escape it.

credentials/apps/records/views.py
```diff
@@ -30,13 +30,16 @@ def _get_programs(self):
...
+        request = self.request
+        site_configuration = request.site.siteconfiguration
         context.update({
             'child_templates': {
                 'footer': self.select_theme_template(['_footer.html']),
                 'header': self.select_theme_template(['_header.html']),
             },
             'programs': json.dumps(self._get_programs(), sort_keys=True),
             'render_language': self.request.LANGUAGE_CODE,
+            'records_help_url': site_configuration.records_help_url,
```
mikix: Might want some fallback if site_configuration is None here (nothing bound to the Site object).

credentials/conf/locale/en/LC_MESSAGES/django.po
```diff
@@ -7,15 +7,15 @@ msgid ""
...
+"POT-Creation-Date: 2018-06-13 20:21+0000\n"
+"PO-Revision-Date: 2018-06-13 20:21:54.957081\n"
 "Last-Translator: \n"
 "Language-Team: openedx-translation <openedx-translation@googlegroups.com>\n"
 "Language: en\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
-"Plural-Forms: nplurals=2; plural=(n != 1);\n"
+"Plural-Forms: nplurals=INTEGER; plural=EXPRESSION;\n"
```
mikix: This strikes me as odd. Have you done a "make pull" in devstack recently. I fixed a bug to avoid this change recently, looks like you might not have it in your container yet.

credentials/static/components/RecordsList.jsx
```diff
@@ -20,22 +22,13 @@ class RecordsList extends React.Component {
     );
   }
 
-  static renderFAQ() {
-    // TODO: not final wording, not gettext'ed
+  static renderFAQ(helpURL) {
+    // TODO: translations
```
mikix: Why still have this todo? We're marking it for translation, yeah?

credentials/static/components/RecordsList.jsx
```diff
@@ -79,6 +72,7 @@ class RecordsList extends React.Component {
     super(props);
     this.state = {
       programs: RecordsList.convertPropsDataToTableData(this.props.programs),
+      helpURL: this.props.helpURL,
```
mikix: You don't need to save it in state. It's not going to change. You can just refer to it below as this.props.helpURL.

(Small nit, you don't have to do this, but helpUrl might look better to my eyes at least. Acronyms that are capitalized start looking weird once you have another word in there, like helpURLTitle. So I generally don't capitalize the full acronym.)

credentials/static/components/RecordsList.jsx
```diff
@@ -105,18 +99,20 @@ class RecordsList extends React.Component {
...
 }
 
 RecordsList.propTypes = {
   programs: PropTypes.arrayOf(PropTypes.object),
+  helpURL: PropTypes.string,
 };
 
 RecordsList.defaultProps = {
   programs: [],
+  helpURL: '',
```
mikix: If React isn't given a URL, maybe don't show the FAQ section? (better than showing a broken link) Like, it's possible that some partner site doesn't bother creating a help center. And they just don't care about the FAQ part.

credentials/static/components/Utils.jsx
```diff
@@ -0,0 +1,31 @@
...
+  static interpolate(formatString, parameters) {
+    return formatString.replace(/{\w+}/g, (parameter) => {
+      const parameterName = parameter.slice(1, -1);
+      return String(parameters[parameterName]);
+    });
+  }
+
+}
+
+// FIXME: Make sure that HTML is sanitized for XSS first
```
mikix: Maybe link to the ticket:
https://openedx.atlassian.net/browse/LEARNER-5537

credentials/static/components/Utils.jsx
```diff
@@ -0,0 +1,31 @@
...
+    htmlString = formatString;
+  } else {
+    htmlString = StringUtils.interpolate(formatString, parameters);
+  }
+
+  return (
+    /* eslint-disable react/no-danger */
+    <span dangerouslySetInnerHTML={{ __html: htmlString }} />
+  );
+}
```
mikix: Nice function.

credentials/static/components/Utils.jsx
```diff
@@ -0,0 +1,31 @@
...
+
+
+export { StringUtils, renderDangerousHTML };
```
mikix: Feels a little odd to be returning a namespace and a function as peers. Maybe we don't need the wrapper StringUtils (maybe rename interpolate as interpolateString then)? Or put renderDangerousHTML underneath StringUtils...

credentials/apps/core/management/commands/create_or_update_site.py
spencerhance: Good call, looks like I need to add this to some more existing tests and also submit a PR in devstack for provision-credentials.sh

credentials/static/components/RecordsList.jsx
spencerhance: Found the styleguide [here](http://edx.readthedocs.io/projects/edx-developer-guide/en/latest/style_guides/python-guidelines.html?highlight=acronym) and it turns out I totally had it wrong 😄

credentials/apps/records/views.py
```diff
@@ -30,13 +30,23 @@ def _get_programs(self):
...
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
+        request = self.request
+
+        site_configuration = request.site.siteconfiguration
+
+        if not site_configuration:
+            records_help_url = ''
+        else:
+            records_help_url = site_configuration.records_help_url
```
mikix: A slightly more Pythonic way to do this would be:
records_help_url = site_configuration.records_help_url if site_configuration else ''
But it's just a matter of taste.

credentials/static/components/RecordsList.jsx
```diff
@@ -105,18 +98,20 @@ class RecordsList extends React.Component {
         </header>
         {hasPrograms && this.renderPrograms()}
         {hasContent || RecordsList.renderEmpty()}
-        {hasContent && RecordsList.renderFAQ()}
+        {hasHelpUrl && RecordsList.renderFAQ(this.props.helpUrl)}
```
mikix: I think you want hasContent && hasHelpUrl here. We don't want to render the FAQ on an empty page.

credentials/apps/core/models.py
```diff
@@ -81,6 +81,12 @@ class SiteConfiguration(models.Model):
         blank=False,
         null=True,
     )
+    records_help_url = models.URLField(
+        verbose_name='Student Records Help URL',
+        help_text='URL of page for questions about Student Records',
+        blank=False,
+        null=True,
```
mikix: My SQL-foo is not strong, but you might want blank=True, null=False? Since that's how we treat it in practice? But looks like the surrounding code does what you did, so leave it as is I guess.

credentials/static/components/RecordsList.jsx
mikix: Or... Maybe I'm wrong there?

mikix: I guess that's a Marco question. But I can see how that would be useful to someone coming at this page and it's empty.

spencerhance: paging @marcotuts

credentials/apps/core/models.py
spencerhance: Hmm, good catch.  I'm assuming that change would default to an empty string?  Which is what I'd also expect/want

mikix: That's what I'd expect. But you might want to verify with docs or testing.

spencerhance: Ah so it looks like `blank=False` makes the form required in /admin which _is not_ what we want.  `null=True` sets it to an empty string by default which _is_ what we want.  If I understand it correctly.

credentials/apps/records/views.py
spencerhance: So now that the records are not required, apparently django sets the records_help_url to None and it just gets cast to a string in the URL.  I'm not sure why this is happening, but in the meantime I added an additional check to make sure it exists (although maybe I should use "is not None"), which makes doing it in one line kind of unwieldy.  Still not sure why it's not just an empty string though.

mikix: Maybe specify default="" in the model?

spencerhance: Good call!  Just tested it and it looks like it works.  I'll push that in.

PR #403, opened 2018-06-19
Alasdair/learner 5444 hook up api to share modal

AlasdairSwan:
Ready for review, but I've still to verify that the successful API call resolves as expected.


credentials/static/components/ProgramRecord.jsx
```diff
@@ -35,120 +36,95 @@ class ProgramRecord extends React.Component {
...
+    const { shareModelOpen } = this.state;
+    const { learner, program, platform_name, grades, uuid } = this.props;
     const recordWrapperClass = 'program-record';
 
     return (
       <main className={recordWrapperClass}>
-        {record &&
+        {learner &&
           <section id="program-record-actions" className={['program-record-row']}>
             <a href="/records/" className="top-bar-link flex-4">{gettext('< Back to My Records')}</a>
```
jlajoie: I know this isn't something you did, but thinking about it this may cause problems when we set up translation linting for this repo, having this as `&lt;` might make it easier to not detect this as XML.

Separate thought, is this an a11y concern because it's sort of a nonsensical piece of text that only makes sense visually (specifically the `<` symbol)

credentials/static/components/ShareProgramRecordModal.jsx
```diff
@@ -20,14 +21,28 @@ class ShareProgramRecordModal extends React.Component {
...
+    const { username, uuid } = this.props;
+
+    const headers = {
+      withCredentials: true,
+      headers: {
+        'X-CSRFToken': Cookies.get('credentials_csrftoken'),
+      },
+    };
+
+    axios.post('/records/new/', {
```
jlajoie: Not sure if this is a placeholder or if we're hoping to go full blown REST best practices here, but if we are this probably shouldn't be /new, POST implies creation so just posting to /records should be enough, if we don't care feel free to ignore 👍

PR #416, opened 2018-06-21
LEARNER-5577 Management command to seed db with fake data

spencerhance:
**Ticket:** https://openedx.atlassian.net/browse/LEARNER-5577

### DIFF
--- a/credentials/apps/records/management/commands/seed-records.py
+++ b/credentials/apps/records/management/commands/seed-records.py
@@ -1,0 +1,282 @@
+import logging
+from datetime import datetime
+
+from django.contrib.contenttypes.models import ContentType
+from django.contrib.sites.models import Site
+from django.core.management import BaseCommand
+from django.db import transaction
+from faker import Faker
+
+from credentials.apps.catalog.models import Course, CourseRun, Organization, Program
+from credentials.apps.core.models import User
+from credentials.apps.credentials.models import CourseCertificate, ProgramCertificate, Signatory, UserCredential
+from credentials.apps.records.models import ProgramCertRecord, UserGrade
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    """ Seed all the data needed to display fake student records """
+    help = 'Seed catalog with fake data'
+
+    def add_arguments(self, parser):
+        """ Add arguments to the command parser """
+        parser.add_argument(
+            '--site-name',
+            action='store',
+            default='Open edx',
+            required=False,
+            help="The site to attach all records to"
+        )
+
+        parser.add_argument(
+            '--user-name',
+            action='store',
+            type=str,
+            default='edx',
+            required=False,
+            help="The user to attach all certs to"
+        )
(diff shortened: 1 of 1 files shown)


credentials/apps/records/management/commands/seed-records.py
```diff
@@ -0,0 +1,247 @@
...
+
+
+class Command(BaseCommand):
+    """ Seed all the data needed to display fake student records """
+    help = 'Seed catalog with fake data'
+
+    def handle(self, *args, **kwargs):  # pylint: disable=unused-argument
+        Command.seed_all()
+
+    @staticmethod
```
jlajoie: nit: So these don't _need_ to be static, but there's no real harm one way or another. We're not going to import from here to call `Command.log_action`

credentials/apps/records/management/commands/seed-records.py
```diff
@@ -0,0 +1,247 @@
...
+    @transaction.atomic
+    def seed_all():
+        """ Seed all catalog data """
+        # Make predictable UUIDs using faker
+        faker = Faker()
+        faker.seed(1234)
+
+        site = Site.objects.get(name='Open edx')
```
jlajoie: I think having these as not site specific, but configurable via `add_arguments` could be helpful here. Not wrong, but serving these as a default that can be overridden via the command line call will be helpful. Same for site, user, etc.

credentials/apps/records/management/commands/seed-records.py
spencerhance: Fair point, i'll switch em over!

credentials/apps/records/management/commands/seed-records.py
```diff
@@ -0,0 +1,247 @@
...
+
+    @staticmethod
+    def seed_courses(site, organizations, faker):
+        """ Seed two courses per organization """
+        courses = []
+        course_id = 1
+
+        for organization in organizations:
+
+            course1, _ = Course.objects.get_or_create(
```
jlajoie: If you're using the `_` downstream should use a real variable name like `created` or something. Python convention has it that _ is typically a throwaway variable

credentials/apps/records/management/commands/seed-records.py
jlajoie: Second thought, I like that they're static because they encourage nothing in here depends on the Command state.

credentials/apps/records/management/commands/seed-records.py
```diff
@@ -0,0 +1,247 @@
...
+        """ Seed  program cert records for user"""
+        program_cert_records = []
+
+        for program_certificate in program_certificates:
+            program_cert_record, _ = ProgramCertRecord.objects.get_or_create(
+                certificate=program_certificate,
+                user=user,
+                uuid=faker.uuid4())
+
+            Command.log_action("", program_certificate, _)
```
jlajoie: Might want to add in a log message here instead of `""`

PR #421, opened 2018-06-25
Added check for public/private view

ColeLRogers:
Added a boolean field to indicate whether the public or private view should be visible. It is private if the program record is associated with the current logged in user (checked via cookie), and public otherwise. This has been commented out to pass code coverage for now, will be used for LEARNER-5550.

Also added a utility function for parsing the 'edx-user-info' cookie as its JSON string contains escaped quotes and encoded commas which break JSON.parse().

### DIFF
--- a/credentials/static/components/Utils.jsx
+++ b/credentials/static/components/Utils.jsx
@@ -25,6 +25,11 @@ class StringUtils {
       <span dangerouslySetInnerHTML={{ __html: htmlString }} />
     );
   }
+  // Parses a JSON string that contains encoded commas and escaped quotes around keys/values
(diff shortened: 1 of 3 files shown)


credentials/static/components/Utils.jsx
```diff
@@ -25,6 +25,10 @@ class StringUtils {
       <span dangerouslySetInnerHTML={{ __html: htmlString }} />
     );
   }
+
+  static parseDirtyJSON(jsonString) {
```
AlasdairSwan: What do you mean by 'dirty' JSON? A comment describing what this function does would be useful.

PR #422, opened 2018-06-25
LEARNER-5624 Implements _get_record() for ProgramRecordView

spencerhance:


credentials/apps/records/views.py
```diff
@@ -61,71 +64,74 @@ def dispatch(self, request, *args, **kwargs):
...
-                    'school': 'TestX',
-                    'attempts': 1,
-                    'course_id': 'course5x',
-                    'issue_date': '2018-01-01',
-                    'percent_grade': '98%',
-                    'letter_grade': 'A'
-                }
-            ]
-        }
+    def _get_record(self, program_uuid):
```
jlajoie: @mikix wouldn't hurt to have a second set of eyes on here for the queries

credentials/apps/records/views.py
```diff
@@ -61,71 +64,74 @@ def dispatch(self, request, *args, **kwargs):
...
+        user = self.request.user
+        platform_name = self.request.site.siteconfiguration.platform_name
+
+        program = Program.objects.get(uuid=program_uuid)
+        program_course_runs = program.course_runs.all()
```
jlajoie: I wonder if you'd be able to do a .select_related('course_runs') here

credentials/apps/records/views.py
```diff
@@ -61,71 +64,74 @@ def dispatch(self, request, *args, **kwargs):
...
+                                      status=UserCredential.AWARDED))
+        course_certificates = CourseCertificate.objects.filter(course_id__in=program_course_run_keys,
+                                                               user_credentials__in=course_user_credentials)
+        # Maps course run 'key' to the cert
+        course_certificate_dict = {
+            course_certificate.course_id: course_certificate for course_certificate in course_certificates}
+
+        # Get all user grades relevant to this program
+        course_grades = UserGrade.objects.filter(
+            username=str(user), course_run__in=program_course_runs, verified=True)  # Only verified courses count here
```
jlajoie: Can you user.username I think

jlajoie: Also might be able to select_related for the course_runs here as well so we're not fetching them downstream in the loop

credentials/apps/records/views.py
```diff
@@ -61,71 +64,74 @@ def dispatch(self, request, *args, **kwargs):
...
+                        (course in highest_attempt_dict and
+                            percent_grade > highest_attempt_dict[course].percent_grade)):
+                    highest_attempt_dict[course] = course_grade
+
+        learner_data = {'full_name': user.get_full_name(),
+                        'username': str(user),
+                        'email': user.email}
+
+        program_data = {'name': program.title,
+                        'school': ','.join(sorted(program.authoring_organizations.values_list('name', flat=True)))}
```
jlajoie: Will probably want a space after the comma, `', '.join`

credentials/apps/records/views.py
spencerhance: Is that more efficient once you already have the program though? I think I still need the program data in the final json

credentials/apps/records/views.py
spencerhance: I saw that's what you were doing with the records list but I'm not sure what the expected behavior is.  Is the front end just gonna do a split on the comma?  Or just use the entire string?

credentials/apps/records/views.py
```diff
@@ -61,71 +64,74 @@ def dispatch(self, request, *args, **kwargs):
...
+
+        course_data = [{'name': course.title,
+                        'school': ','.join(sorted(course.owners.values_list('name', flat=True))),
```
spencerhance: Should the organizations be sorted alphabetically?  Or just stay in the order that they were added (which the model does already)?

credentials/apps/records/views.py
spencerhance: turns out prefetch_related works here since it's ManyToMany

credentials/apps/records/views.py
spencerhance: You can also chain select_related's apparently and follow the relationship further apparently, using .select_related("course_run__course") shaves some time off

credentials/apps/records/views.py
jlajoie: I guess it really doesn't matter, since it's a label I imagined the UI would dumbly render... could be either way though

credentials/apps/records/views.py
jlajoie: I think just leave it at what the model currently does for now

credentials/apps/records/views.py
spencerhance: _actually_ I think I can just that one select_related since it implies the first select_related

spencerhance:
A bunch of our tests are failing randomly due to duplicate sites from faker.  Currently looking into it

PR #426, opened 2018-06-28
LEARNER-5548 Clean up ProgramRecord.jsx

AlasdairSwan:
https://openedx.atlassian.net/browse/LEARNER-5548


credentials/static/components/ProgramIcon.jsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import MicroMastersIcon from '../images/icons/micromasters.svg';
```
jlajoie: Do any of these icons need to be internationalized? Or will that be out of scope for this ticket?

AlasdairSwan: The icons aren't currently internationalized anywhere else (and I don't think we actually internationalize our program type names - i.e. 'MicroMasters' is always referred to as 'MicroMastesr' regardless of language) so this is out of scope.

jlajoie: Ah sorry didn't actually look at what these icons are, they definitely don't need to be.

Other banners we have internationalized as part of the credentials-themes for both Xseries and Professional Certificates, the MicroMasters one looks the same.

https://github.com/edx/credentials-themes/blob/master/edx_credentials_themes/templates/edx.org/images/professional-certificate-logo-es.svg

https://github.com/edx/credentials-themes/blob/master/edx_credentials_themes/templates/edx.org/images/xseries-logo-es.svg

AlasdairSwan: Good catch, but yeah these are small image-based icons.

credentials/static/components/ProgramRecord.jsx
```diff
@@ -61,106 +65,133 @@ class ProgramRecord extends React.Component {
...
-          // Convert the data to an array despite being a single
-          // object to use the FoldingTable styles
-          data={[learner]}
-          dataKey="username"
-        />
-      </section>
-    );
+    const date = new Date(isoDate);
+
+    return `${date.getMonth() + 1}/${date.getDate()}/${date.getFullYear().toString().slice(2)}`;
```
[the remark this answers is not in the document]
AlasdairSwan: @jlajoie we would need the locale code for that, do you know whether that is currently available on the page?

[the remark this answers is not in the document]
AlasdairSwan: It looks like we only have language preference from cookies. I could map it on the Front End but it would be great if the back end could pass it through.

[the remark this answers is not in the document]
AlasdairSwan: @jlajoie I'm going to merge this as is for know and talk to @marcotuts about i18n dates for this repo because it's a decent amount of work to add i18n dates for React to this repo and I don't want to block the other work in this PR.

PR #427, opened 2018-06-29
Updated records log to include all cert statuses

spencerhance:
Fix for [LEARNER-5624](https://openedx.atlassian.net/browse/LEARNER-5624)

Updates records logic to include all cert statuses (AWARDED or REVOKED), only verified grades are used.

Updates course name and key to the course _run_ name and key
Updates issue date to refer to the user_credential modified date
Updates Program ordering on the _get_programs() to sort programs by name
Updates course-run ordering to use the ordering from Programs SortedManytoMany field


spencerhance:
still working on an optimal way for the records ordering, but @jlajoie feel free to take a look at the ProgramView change I made in case it disagrees with what you wanted

PR #431, opened 2018-07-02
Switch Docker Container to the configuration one.

dianakhuang:
Switching from using (and building) `devstack-slim` to `latest`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,25 +17,8 @@ cache:
   - pip
 
 before_install:
-  # Confirm Docker version and login credentials
-  - docker --version
-  - echo -n "$DOCKER_PASSWORD" | docker login -u="$DOCKER_USERNAME" --password-stdin
+  - docker-compose -f docker-compose.yml -f docker-compose.travis.yml up -d
 
-  # Upgrade Docker
-  - sudo apt-get -y update
-  - sudo apt-get -y install -o Dpkg::Options::="--force-confnew" docker-ce
-  - docker --version
-
-  # Build the image.
-  - docker-compose build
-
-  # Ensure we have a place to store coverage output
-  - mkdir -p coverage
-
-  # Bring everything online
-  - make up-test
-
-# Do NOT install Python requirements.
 # Doing so is a waste of time since they won't be used.
 install: true
 
@@ -51,7 +34,3 @@ after_success:
   - pip install --upgrade codecov
   - make exec-coverage
   - codecov
-  - if [ "$TRAVIS_PULL_REQUEST" == "false" ] && [ "$TRAVIS_BRANCH" == "master" ]; then
-    echo -n "$DOCKER_PASSWORD" | docker login -u="$DOCKER_USERNAME" --password-stdin;
-    docker push edxops/credentials:devstack-slim;
-    fi
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,7 @@
-# docker build . -t edxops/credentials:devstack-slim
+FROM edxops/credentials:latest
 
-FROM edxops/python:3.5
-ENV PYTHONUNBUFFERED 1
-ENV DJANGO_SETTINGS_MODULE credentials.settings.devstack
-ENV ENABLE_DJANGO_TOOLBAR 1
-
-WORKDIR /edx/app/credentials/credentials
-
-# Iceweasel is the Debian name for Firefox
-RUN apt-get update && apt-get install -y \
-    iceweasel \
-    xvfb
+RUN apt-get update && \
+    apt-get install --no-install-recommends -y \
+        firefox \
+        xvfb \
+    && rm -rf /var/lib/apt/lists/*
--- a/Makefile
+++ b/Makefile
@@ -68,25 +68,25 @@ up-test: ## Bring up services for testing
 	docker-compose -f docker-compose.yml -f docker-compose.travis.yml up -d
 
 exec-validate-translations: ## Check translations on a container
-	docker exec -t credentials bash -c 'make validate_translations'
+	docker exec -t credentials bash -c 'source /edx/app/credentials/credentials_env && cd /edx/app/credentials/credentials/ && make validate_translations'
 
 exec-check_translations_up_to_date: ## test translations on a container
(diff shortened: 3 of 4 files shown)


dianakhuang:
The other half of https://github.com/edx/devstack/pull/324 . Moving Travis to use the same configuration-built docker image.

.travis.yml
```diff
@@ -17,30 +17,13 @@ cache:
   - pip
 
 before_install:
-  # Confirm Docker version and login credentials
-  - docker --version
-  - echo -n "$DOCKER_PASSWORD" | docker login -u="$DOCKER_USERNAME" --password-stdin
+  - docker-compose -f docker-compose.yml -f docker-compose.travis.yml up -d
```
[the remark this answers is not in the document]
dianakhuang: I figured that it was okay to not build a docker container anymore to keep it in line with the devstack changes. Does this mean I should delete the `Dockerfile` now?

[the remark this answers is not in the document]
jmbowman: This will probably work most of the time, but it does mean that there could be problems in cases where the code is changed in ways that would alter how the container is built (different dependencies, etc.)  For most reliable testing, Travis should probably check out the configuration repo and use the Dockerfile there to build the image with the local credentials checkout (I don't think there's a good way to specify this right now).

If Firefox is still needed for testing here, the Dockerfile is still needed (although it strikes me as odd that it doesn't also install geckodriver; that's normally needed for Selenium testing).  That could be eliminated by using the standard Selenium Docker container instead, like we use in devstack.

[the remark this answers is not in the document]
dianakhuang: Okay, that does seem like a longer term fix for things. I'm mostly imitating what the `ecommerce` repo does in terms of pulling the pre-built container, so at least we can switch all the IDAs over when things change.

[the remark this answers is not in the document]
jmbowman: Ok, if you're just reusing the latest official container build, you'll probably want to keep `make exec-requirements` below to deal with any dependency changes made in the PR.  Other than that, this is probably good to go assuming everything still runs and passes.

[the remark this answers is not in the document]
dianakhuang: When you say dependency changes, you mean running `apt-get update && apt-get upgrade`, right?

Makefile
```diff
@@ -68,25 +68,25 @@ up-test: ## Bring up services for testing
...
-	docker exec -t credentials bash -c 'make check_translations_up_to_date'
+	docker exec -t credentials bash -c 'source /edx/app/credentials/credentials_env && cd /edx/app/credentials/credentials/ && make check_translations_up_to_date'
 
 exec-clean: ## Remove all generated files from a container
-	docker exec -t credentials bash -c 'make clean'
+	docker exec -t credentials bash -c 'source /edx/app/credentials/credentials_env && cd /edx/app/credentials/credentials/ && make clean'
 
 exec-requirements:
-	docker exec -t credentials bash -c 'make requirements production-requirements'
+	docker exec -t credentials bash -c 'source /edx/app/credentials/credentials_env && cd /edx/app/credentials/credentials/ && make requirements'
```
mikix: It's intentional that we don't want production-requirements here? (Is it because it should be a strict subset of requirements?)

Dockerfile
```diff
@@ -1,13 +1,7 @@
-# docker build . -t edxops/credentials:devstack-slim
+FROM edxops/credentials:latest
 
-FROM edxops/python:3.5
-ENV PYTHONUNBUFFERED 1
-ENV DJANGO_SETTINGS_MODULE credentials.settings.devstack
-ENV ENABLE_DJANGO_TOOLBAR 1
```
mikix: We don't want these envs?

.travis.yml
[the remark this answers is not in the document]
jmbowman: No, I was thinking more of updated Python package dependencies.  A PR could upgrade a dependency or add a new one, and the virtualenv in the container would still be using the old set of requirements; `make exec-requirements` would make sure that everything needed by the current requirements files is installed before running tests.

Dockerfile
jmbowman: These were originally here for devstack usage; now that devstack uses images built in the `configuration` repo and this one is just an extension of that which adds Firefox and xvfb for testing, they shouldn't be needed.

Makefile
jmbowman: Good catch, we probably still need both of these until the requirements files get overhauled for OEP-18 compliance.

dianakhuang:
Changes have been addressed @jmbowman . Thank you so much for all your help!

.travis.yml
```diff
@@ -17,27 +17,11 @@ cache:
...
-  # Ensure we have a place to store coverage output
-  - mkdir -p coverage
-
-  # Bring everything online
-  - make up-test
-
-# Do NOT install Python requirements.
 # Doing so is a waste of time since they won't be used.
 install: true
+  - docker exec -t credentials bash -c 'source /edx/app/credentials/credentials_env && cd /edx/app/credentials/credentials/ && make requirements'
```
jmbowman: Do you still need this with `make exec-requirements` below in `script`?  If so, you should probably remove the `true` after `install`; that's normally how you skip the step when there's nothing to do in it.

dianakhuang: This is probably why I removed the line below, but I agree that it makes sense to remove this one instead. 👍

.travis.yml
```diff
@@ -17,27 +17,7 @@ cache:
...
-
-  # Ensure we have a place to store coverage output
-  - mkdir -p coverage
-
-  # Bring everything online
-  - make up-test
-
-# Do NOT install Python requirements.
-# Doing so is a waste of time since they won't be used.
-install: true
```
jmbowman: One last thing; you still want `install: true` because by default Travis will run `pip install -r requirements.txt` (hence the original comment, which could have been better worded).

https://docs.travis-ci.com/user/languages/python/#pip
https://docs.travis-ci.com/user/customizing-the-build/#Skipping-the-Installation-Step

dianakhuang: Wow, that isn't confusing at all, travis. Fixed now.

PR #435, opened 2018-07-04
Fix cannot resolve svg-inline-loader.

waheedahmed:
LEARNER-5722


waheedahmed:
@edx/learner-spartans Please review.

Ayub-Khan:
@waheedahmed  and @hamzamunir7300  just for my learning if a dependency is not properly in order in package file it wont be installed ?

PR #437, opened 2018-07-05
Adds records analytics

spencerhance:
[LEARNER-5650](https://github.com/edx/credentials/tree/spencerhance/LEARNER-5650-add-records-analytics)

Implements all of the [new GA metrics](https://openedx.atlassian.net/wiki/spaces/AN/pages/779944554/Student+Records?focusedCommentId=789217308#comment-789217308) except for download_started and download_finished which need to be done in the Django API (via [LEARNER-5513](https://openedx.atlassian.net/browse/LEARNER-5513))

### DIFF
--- a/credentials/static/components/ProgramRecord.jsx
+++ b/credentials/static/components/ProgramRecord.jsx
@@ -8,6 +8,7 @@ import ProgramIcon from './ProgramIcon';
 import SendLearnerRecordModal from './SendLearnerRecordModal';
 import ShareProgramRecordModal from './ShareProgramRecordModal';
 import StringUtils from './Utils';
+import trackEvent from './Analytics';
 
 class ProgramRecord extends React.Component {
   constructor(props) {
@@ -54,6 +55,10 @@ class ProgramRecord extends React.Component {
       sendRecordModalOpen: true,
     });
     this.setActiveButton(event.target);
+    trackEvent('edx.bi.credentials.program_record.send_started', {
+      category: 'records',
+      'program-uuid': this.props.uuid,
+    });
   }
 
   loadShareModel(event) {
@@ -61,6 +66,10 @@ class ProgramRecord extends React.Component {
       shareModelOpen: true,
     });
     this.setActiveButton(event.target);
+    trackEvent('edx.bi.credentials.program_record.share_started', {
+      category: 'records',
+      'program-uuid': this.props.uuid,
+    });
   }
 
   closeSendRecordModal() {
@@ -224,6 +233,7 @@ class ProgramRecord extends React.Component {
           <SendLearnerRecordModal
             {...defaultModalProps}
             onClose={this.closeSendRecordModal}
+            uuid={uuid}
           />
         }
         {shareModelOpen &&
--- a/credentials/static/components/SendLearnerRecordModal.jsx
+++ b/credentials/static/components/SendLearnerRecordModal.jsx
@@ -2,8 +2,9 @@ import 'babel-polyfill'; // Needed to support Promises on legacy browsers
 import React from 'react';
 import PropTypes from 'prop-types';
 import { Button, CheckBoxGroup, CheckBox, Modal } from '@edx/paragon';
+import trackEvent from './Analytics';
 
-class ShareProgramRecordModal extends React.Component {
+class SendLearnerRecordModal extends React.Component {
   constructor(props) {
     super(props);
     this.sendRecord = this.sendRecord.bind(this);
@@ -14,15 +15,30 @@ class ShareProgramRecordModal extends React.Component {
     };
   }
 
-  sendRecord() {
-    this.setState({
-      recordSent: true,
-    });
+  // Get the organizations we are sharing for the analytics event
+  // TODO: remove hardcoded values once the state is no longer hardcoded
+  getCheckedOrganizations() {
+    const organizations = [];
+
+    if (this.state.RIT) { organizations.push('RIT'); }
+    if (this.state.MIT) { organizations.push('MIT'); }
+    return organizations;
   }
 
   updateState(checked, name) {
     this.setState({
       [name]: !this.state[name],
+    });
+  }
+
+  sendRecord() {
+    this.setState({
+      recordSent: true,
+    });
+    trackEvent('edx.bi.credentials.program_record.send_finished', {
+      category: 'records',
+      'program-uuid': this.props.uuid,
+      organizations: this.getCheckedOrganizations(),
     });
   }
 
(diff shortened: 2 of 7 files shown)


credentials/static/components/SendLearnerRecordModal.jsx
```diff
@@ -14,10 +14,12 @@ class ShareProgramRecordModal extends React.Component {
...
 
-  sendRecord() {
-    this.setState({
-      recordSent: true,
-    });
+  getCheckedOrganizations() {
+    const organizations = [];
+
+    if (this.state.RIT) { organizations.push('RIT'); }
+    if (this.state.MIT) { organizations.push('MIT'); }
```
AlasdairSwan: What are these lines for?

credentials/static/js/analytics.js
```diff
@@ -40,6 +40,9 @@ function setupClickHandlers() {
 !function(){var analytics=window.analytics=window.analytics||[];if(!analytics.initialize)if(analytics.invoked)window.console&&console.error&&console.error("Segment snippet included twice.");else{analytics.invoked=!0;analytics.methods=["trackSubmit","trackClick","trackLink","trackForm","pageview","identify","reset","group","track","ready","alias","debug","page","once","off","on"];analytics.factory=function(t){return function(){var e=Array.prototype.slice.call(arguments);e.unshift(t);analytics.push(e);return analytics}};for(var t=0;t<analytics.methods.length;t++){var e=analytics.methods[t];analytics[e]=analytics.factory(e)}analytics.load=function(t){var e=document.createElement("script");e.type="text/javascript";e.async=!0;e.src=("https:"===document.location.protocol?"https://":"http://")+"cdn.segment.com/analytics.js/v1/"+t+"/analytics.min.js";var n=document.getElementsByTagName("script")[0];n.parentNode.insertBefore(e,n)};analytics.SNIPPET_VERSION="4.0.0";
 /*eslint-enable*/
 
+  // Export functions to window so they can be picked up by react
+  window.edx.analytics = { handleClick, setupClickHandlers };
```
AlasdairSwan: This is an anti-pattern in modern JavaScript. It would be preferable to export the functions in this file and any React code that needs to use the functions can import them from this file.

credentials/static/components/ProgramRecord.jsx
```diff
@@ -141,13 +141,25 @@ class ProgramRecord extends React.Component {
             <Button
               label={gettext('Send Learner Record')}
               className={['btn-primary']}
-              onClick={this.loadSendRecordModal}
+              onClick={(event) => {
+                this.loadSendRecordModal(event); window.edx.analytics.handleClick(event);
+              }}
```
AlasdairSwan: It would be cleaner and more performant to have a wrapper function that calls these 2 nested functions, rather than creating an anonymous function inline.

credentials/static/components/ProgramRecord.jsx
```diff
@@ -141,13 +141,25 @@ class ProgramRecord extends React.Component {
...
+              data-track-event-property-category="records"
+              data-track-event-property-program-uuid={uuid}
             />
             <Button
               label={gettext('Share')}
               className={['btn-secondary']}
-              onClick={this.loadShareModel}
+              onClick={(event) => {
+                this.loadShareModel(event); window.edx.analytics.handleClick(event);
+              }}
```
AlasdairSwan: It would be cleaner and more performant to have a wrapper function that calls these 2 nested functions, rather than creating an anonymous function inline.

credentials/static/components/SendLearnerRecordModal.jsx
```diff
@@ -61,7 +69,14 @@ class ShareProgramRecordModal extends React.Component {
           <Button
             label={gettext('Send')}
             buttonType="primary"
-            onClick={this.sendRecord}
+            onClick={(event) => {
+              this.sendRecord(event); window.edx.analytics.handleClick(event);
+            }}
```
AlasdairSwan: It would be cleaner and more performant to have a wrapper function that calls these 2 nested functions, rather than creating an anonymous function inline.

credentials/static/components/ProgramRecord.jsx
AlasdairSwan: Or, if this is the only way ```loadSendRecordModal``` is called just add the analytics call to that.

credentials/static/components/ProgramRecord.jsx
```diff
@@ -141,13 +141,25 @@ class ProgramRecord extends React.Component {
...
               label={gettext('Send Learner Record')}
               className={['btn-primary']}
-              onClick={this.loadSendRecordModal}
+              onClick={(event) => {
+                this.loadSendRecordModal(event); window.edx.analytics.handleClick(event);
+              }}
+              data-track-type="click"
+              data-track-event="edx.bi.credentials.program_record.send_started"
+              data-track-event-property-category="records"
+              data-track-event-property-program-uuid={uuid}
```
AlasdairSwan: I'm curious as to the approach here of adding the data to the DOM as custom attributes, just to have ```handleClick``` scrape the DOM to get them. Why not just have ```onClick``` call a new function that passes the values straight to ```handleClick```?.

credentials/static/components/SendLearnerRecordModal.jsx
AlasdairSwan: If ```sendRecord``` is only triggered by this ```onClick``` you can just add the analytics call there.

credentials/static/components/SendLearnerRecordModal.jsx
spencerhance: The analytics event is supposed to grab a list of orgs, so since the two orgs are currently hardcoded this is just adding them to a list.  I may be missing something about JS here, is there an obviously better way to do this?

AlasdairSwan: That's fine, but it would be great to add a comment explaining that this is temporary and will be updated once the data stops being hard-coded.

credentials/static/components/ProgramRecord.jsx
spencerhance: This is mainly to keep the same behavior that our older analytics events use. (e.g. [_actions.html](https://github.com/edx/credentials/blob/master/credentials/templates/_actions.html))

I see your point though, I could create another function for the newer react stuff that would take an object of properties.  Is that what you mean?

AlasdairSwan: Yes. You could even update the old function to scrape the data from the DOM and then call the new function with the scraped data.

spencerhance: I ran into a lot of issues exporting a new function from analytics.js while keeping the old behavior due to firefox errors.  For now I added a new jsx file for the react stuff, but I'd appreciate if you can take another look when you get a chance.

credentials/static/components/SendLearnerRecordModal.jsx
```diff
@@ -75,11 +91,13 @@ ShareProgramRecordModal.propTypes = {
     PropTypes.string,
     PropTypes.bool,
   ]),
+  uuid: PropTypes.string,
```
AlasdairSwan: I think this should be set to ```isRequired```. It is in ```ProgramRecord.jsx``` already and if it wasn't passed the analytics wouldn't be very useful.

credentials/static/components/SendLearnerRecordModal.jsx
```diff
@@ -69,17 +85,19 @@ class ShareProgramRecordModal extends React.Component {
...
     PropTypes.bool,
   ]),
+  uuid: PropTypes.string.isRequired,
 };
 
-ShareProgramRecordModal.defaultProps = {
+SendLearnerRecordModal.defaultProps = {
   onClose: () => {},
   parentSelector: false,
+  uuid: '',
```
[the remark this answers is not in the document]
spencerhance: Sounds good, thanks for bearing with me 😁

PR #444, opened 2018-07-09
Updated backend to support public views at separate URL

ColeLRogers:
- Created new url for public views
- Updated get_record() to work with public views (there's an extra layer)
- Updated the context data to pass along the is_public boolean to the front end for button rendering
- Updated tests, added ProgramCertRecord factory


mikix:
@marcotuts - quick product opinion on /records/programs/shared/{uuid} as the public URL? I don't have a strong opinion myself, but maybe you do?

marcotuts:
url suggestion looks good to me. 👍

ColeLRogers:
Added a new commit to allow non-logged in users to access public view. Could a reviewer take a look at this latest commit?

dianakhuang:
👍 Nice job with the new mixin!

PR #446, opened 2018-07-10
Updated front end to use public/private prop

ColeLRogers:
Previously the ProgramRecord view used a cookie to decide whether it should be public or private. This PR updates it to instead use a prop.

### DIFF
--- a/credentials/static/components/ProgramRecord.jsx
+++ b/credentials/static/components/ProgramRecord.jsx
@@ -1,7 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { Button } from '@edx/paragon';
-import Cookies from 'js-cookie';
 
 import FoldingTable from './FoldingTable';
 import ProgramIcon from './ProgramIcon';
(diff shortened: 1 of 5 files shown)


credentials/static/components/specs/ProgramRecord.test.jsx
```diff
@@ -135,20 +131,14 @@ describe('<ProgramRecord />', () => {
...
   });
 
   describe('Public view of record', () => {
+    const publicProps = { ...defaultProps };
+    publicProps.isPublic = true;
+
     beforeEach(() => {
-      Cookies.set('prod-edx-user-info', cookieJSON, { path: '' });
-      wrapper = mount(<ProgramRecord {...defaultProps} />);
+      wrapper = mount(<ProgramRecord {...publicProps} />);
```
AlasdairSwan: Instead of creating a new const ```publicProps```, you can just do the following;
```
wrapper = mount(<ProgramRecord {...defaultProps} isPublic />);
```

credentials/static/components/ProgramRecord.jsx
```diff
@@ -263,6 +249,7 @@ ProgramRecord.propTypes = {
     last_updated: PropTypes.string,
   }).isRequired,
   grades: PropTypes.arrayOf(PropTypes.object).isRequired,
+  isPublic: PropTypes.bool.isRequired,
```
AlasdairSwan: It may make sense to default to the public view.

PR #447, opened 2018-07-11
Updated request/response format for share modal

ColeLRogers:
- Updated backend to get parameters from body
- Updated backend to return full url instead of just uuid
- Updated front end to read this url

### DIFF
--- a/credentials/static/components/specs/ShareProgramRecordModal.test.jsx
+++ b/credentials/static/components/specs/ShareProgramRecordModal.test.jsx
@@ -16,7 +16,7 @@ const defaultProps = {
 describe('<ShareProgramRecordModal />', () => {
   describe('Happy Path', () => {
     beforeEach(() => {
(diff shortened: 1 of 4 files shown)


credentials/static/components/specs/ShareProgramRecordModal.test.jsx
```diff
@@ -16,7 +16,7 @@ const defaultProps = {
 describe('<ShareProgramRecordModal />', () => {
   describe('Happy Path', () => {
     beforeEach(() => {
-      promise = Promise.resolve({ uuid: '21366aa129514333a4a9f32161ad3a69' });
+      promise = Promise.resolve({ data: { url: '21366aa129514333a4a9f32161ad3a69' } });
```
AlasdairSwan: It could be useful to have the url string look more like an actual url.

ColeLRogers: Ah yes, forgot to update that

PR #448, opened 2018-07-11
Added change to views to require the program credentials to match the…

calistenson:
… sites that they land on. Added test to verify.

https://openedx.atlassian.net/browse/LEARNER-5589

### DIFF
--- a/credentials/apps/credentials/tests/test_views.py
+++ b/credentials/apps/credentials/tests/test_views.py
@@ -13,7 +13,7 @@ from django.utils.text import slugify
 from faker import Faker
 from mock import patch
 
-from credentials.apps.core.tests.factories import USER_PASSWORD, UserFactory
+from credentials.apps.core.tests.factories import USER_PASSWORD, SiteConfigurationFactory, UserFactory
 from credentials.apps.core.tests.mixins import SiteMixin
 from credentials.apps.credentials.exceptions import MissingCertificateLogoError
 from credentials.apps.credentials.models import ProgramCertificate, UserCredential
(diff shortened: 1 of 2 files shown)


credentials/apps/credentials/tests/test_views.py
```diff
@@ -144,6 +144,25 @@ def test_invalid_uuid(self):
         response = self.client.get(path)
         self.assertEqual(response.status_code, 404)
 
+    def test_invalid_site(self):
+        """ Verify that the view returns a 404 if user_credentials are displayed on a site
+        they are not associated with.
+        """
+        domain = 'testserver.fakesite'
```
mikix: Small nit (feel free to ignore). This is very close to the SiteMixin's domain of testserver.fake -- it might not be obvious to future archaeologists what your intent is here. Something like "other.site" or "second.unused" or something along those lines might lend this test some clarity.

credentials/apps/credentials/tests/test_views.py
```diff
@@ -144,6 +144,25 @@ def test_invalid_uuid(self):
...
+            site__domain=domain,
+        )
+        test_site = site_configuration.site
+        test_program_certificate = factories.ProgramCertificateFactory(site=test_site)
+        test_signatory_1 = factories.SignatoryFactory()
+        test_signatory_2 = factories.SignatoryFactory()
+        test_program_certificate.signatories.add(test_signatory_1, test_signatory_2)
+        test_user_credential = factories.UserCredentialFactory(
+            username=self.MOCK_USER_DATA['username'], credential=test_program_certificate
+        )
```
mikix: nit: This is all ripped from the SiteMixin. Maybe worth removing some duplicated code by moving the SiteMixin's code from setUp to a method that you could call here, giving a different domain (and not setting siteid)? But it's all test code, probably doesn't matter much.

credentials/apps/credentials/tests/test_views.py
```diff
@@ -144,6 +144,25 @@ def test_invalid_uuid(self):
...
+        response = self.client.get(test_user_credential.get_absolute_url())
+        self.assertEqual(response.status_code, 404)
```
mikix: Small test nit: It might be nice to add a second stanza to this test where you change the programcertificates's site back to self.site, then re-GET the url and see a success (so that you know you weren't seeing this 404 because of a mistake in the code above, but rather because the site was mismatched).

credentials/apps/credentials/views.py
```diff
@@ -50,7 +50,8 @@ def user_credential(self):
         return get_object_or_404(
             UserCredential,
             uuid=self.kwargs.get('uuid'),
-            status=UserCredential.AWARDED
+            status=UserCredential.AWARDED,
+            program_credentials__site=self.request.site,
```
mikix: :+1:

mikix: When we eventually show certificates for courses too in credentials, this is going to prevent us (because this query requires a program cert). We will eventually have to make this a more complicated query when that happens. Not something you need to solve today, but something to be aware of.

credentials/apps/credentials/tests/test_views.py
calistenson: Not going to change this since some objects reference self and some don't and I didn't want to deal with the difference in a utility function for the two instances of this so far.

PR #450, opened 2018-07-12
Adds UI for sending email

spencerhance:
[LEARNER-5751 Hit endpoint from UI](https://openedx.atlassian.net/browse/LEARNER-5751)

**All 3 alerts, minus the org names:**
image

**Mock:**
image


credentials/static/components/ProgramRecord.jsx
```diff
@@ -111,6 +125,101 @@ class ProgramRecord extends React.Component {
...
+    trackEvent('edx.bi.credentials.program_record.send_finished', {
+      category: 'records',
+      'program-uuid': this.props.uuid,
+      organizations: orgs,
+    });
+
+    // Show the loading "info" alert
+    this.setState({ sendRecordLoadingAlertOpen: true });
+
+    /* eslint-disable prefer-const */
```
AlasdairSwan: Why disable these lint errors instead of just changing to use ```const```?

credentials/static/sass/views/_program-record.scss
```diff
@@ -56,6 +56,14 @@ $record-border-color: #C2C2C2;
     background: #f0cdcf;
   }
 
+  // Needed to override edx-pattern-library default .alert styles
```
AlasdairSwan: I think it's actually the default Paragon alert styles isn't it?

credentials/static/components/ProgramRecord.jsx
```diff
@@ -22,12 +24,24 @@ class ProgramRecord extends React.Component {
...
+    this.renderSendRecordLoadingStatusAlert = this.renderSendRecordLoadingStatusAlert.bind(this);
     this.state = {
       shareModelOpen: false,
       sendRecordModalOpen: false,
       isPublic: true,
       recordDownloaded: false,
+      sendRecordSuccessOrgs: [],
+      sendRecordFailureOrgs: [],
+      sendRecordSuccessAlertOpen: false,
+      sendRecordFailureAlertOpen: false,
```
AlasdairSwan: Can both the success alert and failure alert be open at the same time?

credentials/static/components/ProgramRecord.jsx
```diff
@@ -111,6 +125,101 @@ class ProgramRecord extends React.Component {
...
+    this.setState({ sendRecordSuccessOrgs: [], sendRecordFailureAlertOpen: false });
+  }
+
+  closeSendRecordLoadingStatusAlert() {
+    this.setState({ sendRecordLoadingAlertOpen: false });
+  }
+
+  renderSendRecordLoadingStatusAlert() {
+    const headerMessage = gettext('We are sending your program record.');
+    const dialog = <div> <span className="hd-5"> {headerMessage} </span> <br /> <span className=""> <Icon className={['fa', 'fa-spinner', 'fa-spin']} /> </span> </div>;
```
AlasdairSwan: This can be reformatted to be easier to read, using;
```
const dialog = (
 <div>
    <span className="hd-5"> {headerMessage} </span>
    <br />
    <span className=""> <Icon className={['fa', 'fa-spinner', 'fa-spin']} /> </span>
  </div>
);
```

AlasdairSwan: CSS should be used instead of ```<br />```.
Why is the ```Icon``` wrapped in a span?

credentials/static/components/ProgramRecord.jsx
```diff
@@ -111,6 +125,101 @@ class ProgramRecord extends React.Component {
...
+    return (
+      <StatusAlert dialog={dialog} open={this.state.sendRecordLoadingAlertOpen} alertType="info" onClose={this.closeSendRecordLoadingStatusAlert} />
+    );
+  }
+
+  renderSendRecordFailureStatusAlert() {
+    const orgs = this.state.sendRecordFailureOrgs;
```
AlasdairSwan: Using [object destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Object_destructuring) this can be written more cleanly as;
```
const { sendRecordFailureOrgs } = this.state;
```
(then use ```sendRecordFailureOrgs``` instead of ```orgs``` in later code - which makes it clearer where the value comes from).

credentials/static/components/ProgramRecord.jsx
```diff
@@ -111,6 +125,101 @@ class ProgramRecord extends React.Component {
...
+    const headerMessage = gettext('We were unable to send your program record.');
+    const bodyMessage = StringUtils.interpolate(gettext('We were unable to send your record to {orgs}.  You can attempt to send this record again.  Contact support if this issue persists.'), { orgs: StringUtils.formatStringList(orgs) });
+    const dialog = <div> <span className="hd-5"> {headerMessage} </span> <br /> <span className="text-muted"> {bodyMessage} </span> </div>;
```
AlasdairSwan: Please update the formatting as per my previous comment.

credentials/static/components/ProgramRecord.jsx
```diff
@@ -111,6 +125,101 @@ class ProgramRecord extends React.Component {
...
+    return (
+      <StatusAlert dialog={dialog} open={this.state.sendRecordFailureAlertOpen} alertType="danger" onClose={this.closeSendRecordSuccessStatusAlert} />
+    );
+  }
+
+  renderSendRecordSuccessStatusAlert() {
+    const orgs = this.state.sendRecordSuccessOrgs;
+    const headerMessage = gettext('You have successfully shared your Learner Record');
+    const bodyMessage = StringUtils.interpolate(gettext('You have sent your record to {orgs}.  Next, ensure you understand their application process.'), { orgs: StringUtils.formatStringList(orgs) });
+    const dialog = <div> <span className="hd-5"> {headerMessage} </span> <br /> <span className="text-muted"> {bodyMessage} </span> </div>;
```
AlasdairSwan: Please update the formatting as per my previous comment.

credentials/static/components/ProgramRecord.jsx
```diff
@@ -111,6 +125,101 @@ class ProgramRecord extends React.Component {
...
+    const headerMessage = gettext('We were unable to send your program record.');
+    const bodyMessage = StringUtils.interpolate(gettext('We were unable to send your record to {orgs}.  You can attempt to send this record again.  Contact support if this issue persists.'), { orgs: StringUtils.formatStringList(orgs) });
+    const dialog = <div> <span className="hd-5"> {headerMessage} </span> <br /> <span className="text-muted"> {bodyMessage} </span> </div>;
+    return (
+      <StatusAlert dialog={dialog} open={this.state.sendRecordFailureAlertOpen} alertType="danger" onClose={this.closeSendRecordSuccessStatusAlert} />
+    );
+  }
+
+  renderSendRecordSuccessStatusAlert() {
+    const orgs = this.state.sendRecordSuccessOrgs;
```
AlasdairSwan: See previous comment around object destructuring.

credentials/static/components/ProgramRecord.jsx
```diff
@@ -155,7 +264,9 @@ class ProgramRecord extends React.Component {
             />
           </section>
         }
-
+        {this.renderSendRecordLoadingStatusAlert()}
+        {this.renderSendRecordFailureStatusAlert()}
+        {this.renderSendRecordSuccessStatusAlert()}
```
AlasdairSwan: I'm not really a fan of these wrapper functions for fairly simple HTML. I think it's clearer when the bulk of the HTML is contained within the render function. The loading status alert could be added within the render return function as;
```
<StatusAlert
  alertType="info"
  open={this.state.sendRecordLoadingAlertOpen}
  onClose={this.closeSendRecordLoadingStatusAlert} />
  dialog={{(
    <div>
      <span className="hd-5">{ gettext('We are sending your program record.') }</span>
      <br />
      <span className=""> <Icon className={['fa', 'fa-spinner', 'fa-spin']} /> </span>
    </div>
  )}}
/>
```

credentials/static/components/ProgramRecord.jsx
```diff
@@ -111,6 +125,101 @@ class ProgramRecord extends React.Component {
     }));
   }
 
+  sendRecords(orgs) {
```
AlasdairSwan: Why is this code here, instead of in ```SendLearnerRecordModal.jsx```?

credentials/static/components/specs/Utils.test.jsx
```diff
@@ -0,0 +1,35 @@
+import StringUtils from '../Utils';
+
+
+describe('String Utils formatStringList()', () => {
+  beforeEach(() => {
+  });
```
AlasdairSwan: If it's not being used there's no reason to include a ```beforeEach```.

credentials/static/components/ProgramRecord.jsx
spencerhance: It seems somewhat counterintuitive to have const values that are expected to change, but I don't mind changing it.

credentials/static/components/ProgramRecord.jsx
spencerhance: Yes, right now the expected behavior is that if you send to multiple orgs, and a few fail and a few succeed, the success alert would show those that succeeded, and the failure alert would show those that failed.

credentials/static/components/ProgramRecord.jsx
AlasdairSwan: This gives a good overview of the preconception that ```const``` is a constant https://mathiasbynens.be/notes/es6-const

credentials/static/components/ProgramRecord.jsx
AlasdairSwan: Cool, just wanting to clarify.

credentials/static/components/ProgramRecord.jsx
spencerhance: This was only because the ProgramRecord state depends on what happens in this function, I wasn't sure if there was a way to put it in the modal and still be able to set the state variables.

credentials/static/components/ProgramRecord.jsx
spencerhance: Whoops, looks like a leftover from a copy and paste 🙃

credentials/static/components/ProgramRecord.jsx
spencerhance: Sounds good, I'll just move all of the alerts

credentials/static/components/ProgramRecord.jsx
spencerhance: @AlasdairSwan Mind taking another look at this PR?

PR #452, opened 2018-07-13
Send credit-request emails

mikix:
Adds a new endpoint to send credit-request emails to partners. Text for those emails is just temporary.

Also renames the old create-share endpoint to be consistent.

The ace config probably doesn't work on prod because we don't have the devops support there.


mikix:
@ColeLRogers you might want to check that I haven't mucked with your endpoint in ways you don't approve of. (I renamed it from /new to /programs/UUID/share and added a requirement that you be logged in -- it already checked username, so this was more of a just-in-case thing)

ColeLRogers:
Those changes look good

PR #454, opened 2018-07-13
Drop hardcoded data

mikix:
Some small fixes I noticed could be done:

1. Drops hardcoded "Course 6" data, we pull from db these days
1. Supplies "last_updated" field up to UI
1. Pass a "completed" bool to the UI and **it** decides which string to show and which color. (and actually show a different color depending)
1. Fix some fiddliness with the program type slug.
1. Fix the My Records page title


credentials/apps/records/views.py
```diff
@@ -156,19 +158,23 @@ def _get_record(self, uuid, is_public):
...
 
+        last_updated = last_updated or datetime.datetime.today()
+
         learner_data = {'full_name': user.get_full_name(),
                         'username': user.username,
                         'email': user.email, }
 
         program_data = {'name': program.title,
                         'type': slugify(program.type),
+                        'last_updated': last_updated.strftime('%x'),
```
mikix: Hmm, have to find out a per-request-locale way of doing this.

credentials/static/components/RecordsList.jsx
```diff
@@ -32,7 +32,9 @@ class RecordsList extends React.Component {
                 <ProgramIcon type={row.type} className="inline-data certificate-icon" />
                 <span className="record-partner inline-data">{row.partner}</span>
                 <span className="inline-data"> {' | '} </span>
-                <span className="font-weight-bold inline-data">{row.progress}</span>
+                <span className="font-weight-bold inline-data">{
+                  row.completed ? gettext('Completed') : gettext('In Progress')
```
jlajoie: Should this also match the "Earned/Partially Completed" change?

mikix: I was following the mocks...

PR #460, opened 2018-07-16
LEARNER-5513: Adds in CSV download and endpoint

jlajoie:


credentials/apps/records/tests/test_views.py
```diff
@@ -593,3 +593,66 @@ def test_feature_toggle(self):
...
+        """ Verify that the view 404s if a program cert isn't found"""
+        self.program_cert.delete()
+        response = self.client.get(
+            reverse('records:program_record_csv', kwargs={'uuid': self.program_cert_record.uuid.hex})
+        )
+        self.assertEqual(404, response.status_code)
+
+    @override_flag(WAFFLE_FLAG_RECORDS, active=True)
+    def test_404s_with_no_program_cert(self):
+        """ Verify that the view 404s if a program cert isn't found"""
```
mikix: Name and docstring seem wrong. And they hide the previous method.

credentials/apps/records/urls.py
```diff
@@ -10,6 +10,9 @@
         name='private_programs'),
     url(r'^programs/shared/{uuid}/$'.format(uuid=UUID_PATTERN), views.ProgramRecordView.as_view(), {'is_public': True},
         name='public_programs'),
+    url(r'^programs/shared/csvs/{uuid}/$'.format(uuid=UUID_PATTERN),
```
mikix: I'd feel much more comfortable with:
'^programs/shared/{uuid}/csv$'

credentials/apps/records/views.py
```diff
@@ -299,3 +305,28 @@ def dispatch(self, request, *args, **kwargs):
...
+    """
+    Returns a csv view of the Progam Record for a Learner from a username and program_uuid
+    """
+
+    def get(self, request, *args, **kwargs):
+        if not waffle.flag_is_active(request, WAFFLE_FLAG_RECORDS):
+            raise http.Http404()
+        program_cert_record = get_object_or_404(ProgramCertRecord, uuid=kwargs.get('uuid'))
+        program_certificate = get_object_or_404(ProgramCertificate, id=program_cert_record.certificate_id)
+        record = get_record_data(program_cert_record.user, program_certificate.program_uuid, request.site)
```
mikix: This leaves platform_name=None in the csv right? Maybe take that argument out of get_record_data and just add platform_name manually in the other view?

credentials/apps/records/views.py
```diff
@@ -299,3 +305,28 @@ def dispatch(self, request, *args, **kwargs):
...
+
+        string_io = io.StringIO()
+        writer = csv.DictWriter(string_io, record['grades'][0].keys(), quoting=csv.QUOTE_ALL)
+        writer.writeheader()
+        writer.writerows(record['grades'])
+        string_io.seek(0)
+        filename = '{username}_{program_name}_grades.csv'.format(
+            username=record['learner']['username'],
+            program_name=record['program']['name']
```
mikix: Nit: There's a marketing slug for programs that discovery provides us. We could keep that when we query discovery in copy-catalog, and use it here.

credentials/static/components/ProgramRecordFactory.jsx
```diff
@@ -5,16 +5,7 @@ import ProgramRecord from './ProgramRecord';
...
-  let uuid = url.substring(url.indexOf(urlPrefix) + urlPrefix.length);
-
-  const queryIndex = uuid.indexOf('?');
-  if (queryIndex > -1) {
-    uuid = uuid.substring(0, queryIndex);
-  }
-
-  // Remove any trailing slashes
-  return uuid.replace(/\//g, '');
+  return /records\/programs\/(?:shared\/)?([a-zA-Z0-9]{32})/.exec(url)[1];
```
mikix: Nit: This whole method could be removed if we just passed the uuid up from the view, right? That seems cleaner. I know this isn't your mess.

package.json
```diff
@@ -29,11 +29,12 @@
     "font-awesome": "^4.7.0",
     "imports-loader": "^0.7.1",
     "js-cookie": "2.2.0",
+    "js-file-download": "^0.4.1",
```
mikix: Not updating package-lock.json?

js-file-download isn't used anywhere else in edx and doesn't seem so widely used. Is it buying us much? (Just thinking it may go away or be poorly maintained.)

credentials/static/components/ProgramRecord.jsx
```diff
@@ -71,9 +73,15 @@ class ProgramRecord extends React.Component {
...
-    // TODO: Add functionality as part of LEARNER-5513
+  downloadRecord(uuid) {
     this.setState({ recordDownloaded: true });
+    axios({
+      url: `/records/programs/shared/csvs/${uuid}`,
+      method: 'GET',
+      responseType: 'blob',
+    }).then((response) => {
+      FileDownload(response.data, /filename=(.*)/.exec(response.headers['content-disposition'])[1]);
+    });
```
mikix: Are errors logged to the console automatically? If not, might be nice to do so, to help debugging. Or throw up a user-visible error even. Like a status alert that says "nope can't do it"

credentials/apps/records/tests/test_views.py
```diff
@@ -593,3 +593,66 @@ def test_feature_toggle(self):
...
+        response = self.client.get(
+            reverse('records:program_record_csv', kwargs={'uuid': self.program_cert_record.uuid.hex})
+        )
+        self.assertEqual(200, response.status_code)
+        # Check that the header is present in the response bytestring
+        headers = ['course_id','percent_grade','attempts','school','issue_date','letter_grade','name']
+        response_content = response.content.decode('utf-8')
+        for header in headers:
+            self.assertIn(header, response_content)
```
mikix: Nit: You could also maybe use the csv module to parse the output here and validate that it's well-formed csv. (I know we use csv to create this output, so it better be...)

credentials/static/components/ProgramRecord.jsx
```diff
@@ -71,9 +73,15 @@ class ProgramRecord extends React.Component {
...
-  downloadRecord() {
-    // TODO: Add functionality as part of LEARNER-5513
+  downloadRecord(uuid) {
     this.setState({ recordDownloaded: true });
+    axios({
+      url: `/records/programs/shared/csvs/${uuid}`,
+      method: 'GET',
+      responseType: 'blob',
+    }).then((response) => {
+      FileDownload(response.data, /filename=(.*)/.exec(response.headers['content-disposition'])[1]);
```
mikix: Maybe just set window.location? I dunno if that works or not.

credentials/static/components/ProgramRecordFactory.jsx
jlajoie: 1. Regex is bae
2. Absolutely, makes more sense to include it in the context and the provide it to the props

package.json
jlajoie: Swapping this to window.location to avoid this

credentials/static/components/ProgramRecord.jsx
mikix: Huh. Why would it fail? Bad uuid? That shouldn't happen from the page link I wouldn't think

mikix: Like, not a realistic error case I guess is what I'm saying. So maybe 404 is fine?

mikix: And at least those would be logged in splunk

credentials/apps/records/views.py
jlajoie: So it looks like the only unique index is based on the `title` in Disco (name in Credentials) and it seems like a non-trivial amount of work for a nit?

mikix: I was using nit as in optional comment, not necessarily scope of work, sorry.

marketing_slug isn't unique I don't think, no. But it might result in a more traditional filename (mterry_mit-data-science_grades.csv rather than mterry_MIT Data Science_grades.csv). That's all.

It's already exposed by disco, but you'd need to add a colum to our Program table, parse it in copy_catalog, and reference it here. Non trivial work, but might buy us filenames without spaces. No need to do it.

jlajoie: Alternative... I just regex replace spaces with `-` and lower case it?

mikix: Yeah, fair.

credentials/apps/records/tests/test_views.py
```diff
@@ -593,3 +595,75 @@ def test_feature_toggle(self):
...
+        response = self.client.get(
+            reverse('records:program_record_csv', kwargs={'uuid': self.program_cert_record.uuid.hex})
+        )
+        self.assertEqual(404, response.status_code)
+
+    @override_flag(WAFFLE_FLAG_RECORDS, active=True)
+    @patch('credentials.apps.records.views.SegmentClient', autospec=True)
+    @patch('credentials.apps.records.views.SegmentClient.track', autospec=True)
+    def test_404s_with_no_program_cert(self, client, track):
+        """ Verify that the view 404s if a program cert isn't found"""
```
mikix: This and the previous test have the same name and docstring. That hides the previous test right?

credentials/apps/records/urls.py
```diff
@@ -10,6 +10,9 @@
         name='private_programs'),
     url(r'^programs/shared/{uuid}/$'.format(uuid=UUID_PATTERN), views.ProgramRecordView.as_view(), {'is_public': True},
         name='public_programs'),
+    url(r'^programs/shared/{uuid}/csvs$'.format(uuid=UUID_PATTERN),
```
mikix: I don't love the plural? Like, this is the "csv" version of the record. Much like we might someday add a "pdf" version or something. But plural feels weird in those cases.

credentials/static/components/ProgramRecord.jsx
```diff
@@ -71,11 +71,12 @@ class ProgramRecord extends React.Component {
     this.activeButton.focus();
   }
 
-  downloadRecord() {
-    // TODO: Add functionality as part of LEARNER-5513
-    this.setState({ recordDownloaded: true });
+  downloadRecord(uuid) {
+    this.setState({recordDownloaded: true});
```
mikix: Do we use this state field at all?

credentials/apps/records/tests/test_views.py
jlajoie: Oh wow thanks for catching this, fixed

credentials/apps/records/urls.py
jlajoie: Donezo

credentials/static/components/ProgramRecord.jsx
jlajoie: As far as I can tell no outside of tests... and I wasn't expecting to remove things as part of this PR, I can, but there may be value in just leaving this?

mikix: Oh sure, if we use it in tests, alright. I just didn't see it in the file and thought maybe it was cruft.

PR #461, opened 2018-07-16
Added a breadcrumb to the student records page to link back to studen…

calistenson:
…t profile. Also added some css changes so that the background on the records list and program records pages are white but the certificate page background stays gray.

https://openedx.atlassian.net/browse/LEARNER-5585


credentials/static/components/RecordsList.jsx
```diff
@@ -64,12 +64,15 @@ class RecordsList extends React.Component {
...
 
   render() {
-    const { programs, helpUrl } = this.props;
+    const { programs, helpUrl, profileUrl } = this.props;
     const hasPrograms = programs.length > 0;
     const hasHelpUrl = helpUrl !== '';
     const hasContent = hasPrograms; // will check for courses when we show those too
     return (
       <main className="record">
+         <a href= {profileUrl} className="top-bar-link flex-4">
```
spencerhance: Will this still render if profileUrl is an empty string?

credentials/apps/records/views.py
```diff
@@ -83,6 +84,7 @@ def get_context_data(self, **kwargs):
 
         site_configuration = request.site.siteconfiguration
 
+        profile_url = urllib.parse.urljoin(site_configuration.homepage_url, 'u/' + request.user.username) if site_configuration else ''
```
spencerhance: Seems good, I would just double check that homepage_url is a required config and/or can't be set to None in MySQL if you haven't already.

calistenson: I added a check in RecordsList.jsx to see if the url was an empty string and then wrote a rendering function to only show the url if it exists. Thanks for reminding me to double check that it may not be required.

spencerhance: Thanks and nice work!  It's confusing how some of the models interact and the way that our DB is set up.  IMO site_config values should never be None and should always exist as an empty string when not set, but unless you're very deliberate about your models, our current MySQL will just insert Null strings.

PR #463, opened 2018-07-16
Update program icons to reference those in credentials-themes

ColeLRogers:
Program icons will now come from the credentials-themes repo via Django templates. The icons get passed in as a dictionary prop, and if no icons are provided, or one doesn't exist for a certain program type, it just won't render one and won't give any errors. The base templates in the credentials repo do not have this icon dictionary, but if a corresponding template exists in credentials-themes ({THEME_NAME}/programs/programs.html), then they will get included and override the block which contains the factory call.


credentials/static/components/ProgramRecord.jsx
```diff
@@ -161,7 +162,7 @@ class ProgramRecord extends React.Component {
             <div className="program-overview">
               <h1 className="program-title h2">{ StringUtils.interpolate(gettext('{name} Record'), { name: program.name }) }</h1>
               <div className="text-muted program-type">
-                <ProgramIcon type={program.type} className="program-icon" />
+                <ProgramIcon icon={icons[program.type.replace('-', '_')]} className="program-icon" />
```
AlasdairSwan: Can we centralize this so we don't need to string parse the type every time we try to get an icon? (e.g. add the parsing in the ```ProgramIcon``` component)?

credentials/static/components/ProgramIcon.jsx
```diff
@@ -1,45 +1,30 @@
...
-    const html = { __html: icon };
-    return React.createElement(element, {
-      'aria-hidden': true,
-      className,
-      dangerouslySetInnerHTML: html,
-    });
-  }
-
-  return null;
+  const html = { __html: icon };
```
AlasdairSwan: I don't think you need this variable. Can't you just do this on line 14;
```
  dangerouslySetInnerHTML: { __html: icon },
```

credentials/static/components/ProgramRecord.jsx
ColeLRogers: We could, that would just require us to pass in the icons dictionary to ProgramIcon as well. Would that be preferred?

credentials/static/components/ProgramIcon.jsx
AlasdairSwan: Cool, I can't remember why I did that :D

credentials/static/components/ProgramRecord.jsx
AlasdairSwan: Ah, now I see. You need to update the program type to access the correct entry in the dictionary. So the problem appears to be that the dictionary uses ```_``` and the API uses ```-```. Why doesn't the dictionary match the API convention?

credentials/static/components/ProgramIcon.jsx
```diff
@@ -1,44 +1,33 @@
...
+    dangerouslySetInnerHTML: { __html: icon },
+  });
 };
 
 ProgramIcon.propTypes = {
   className: PropTypes.string,
   element: PropTypes.string,
-  type: PropTypes.string.isRequired,
+  type: PropTypes.string,
+  iconDict: PropTypes.shape(),
```
AlasdairSwan: Both of these props are required aren't they? Without either nothing will be rendered.

credentials/static/components/ProgramIcon.jsx
```diff
@@ -1,44 +1,33 @@
...
 };
 
 ProgramIcon.defaultProps = {
   className: '',
   element: 'span',
+  type: '',
+  iconDict: {},
```
AlasdairSwan: I don't think ```type``` and ```iconDict``` should have default values.

credentials/static/components/ProgramIcon.jsx
AlasdairSwan: In that case I would advocate for iconDict having a default value at the page level. It seems too loose otherwise.

PR #465, opened 2018-07-17
LEARNER-5812

bbaker6225:
Add program key to ProgramCertRecord in anticipation of deleting certificate from ProgramCertRecord. Stage 1 of 3 of the process defined below:

1.  Add program foreign key to ProgramCertRecord (migration and release)
2. Start using program in place of certificate (release)
3. Remove certificate (migration and release)


credentials/apps/records/models.py
```diff
@@ -29,13 +29,13 @@ class ProgramCertRecord(TimeStampedModel):
     """
     Connects a User with a Program Certificate
     """
-    certificate = models.ForeignKey(ProgramCertificate)
+    certificate = models.ForeignKey(ProgramCertificate, null=True)
```
dianakhuang: It would be good to verify that this change is backwards compatible with the old version of the code that assumes this field isn't nullable.

PR #466, opened 2018-07-17
WIP: LEARNER-5741: Adds in segment tracking for csv download

jlajoie:


credentials/apps/records/views.py
```diff
@@ -327,6 +369,24 @@ def get(self, request, *args, **kwargs):
...
+            properties={
+                'category': 'records',
+                'program-uuid': program_certificate.program_uuid,
+                'record-uuid': program_cert_record.uuid,
+            },
+            segment_client=segment_client
+        )
         response['Content-Disposition'] = 'attachment; filename={filename}'.format(filename=filename)
-        return response
+        return response
```
mikix: Missing newline :(

requirements/production.txt
```diff
@@ -1,6 +1,7 @@
 # Packages required in a production environment
 -r base.txt
 
+analytics-python==1.2.9
```
mikix: We probably want this in base.txt instead -- since it's always imported.

credentials/apps/records/views.py
```diff
@@ -311,13 +314,52 @@ class ProgramRecordCsvView(View):
...
+                'record_uuid': str(program_cert_record.uuid),
+            },
+            context={
+                'page': {
+                    'path': request.path,
+                    'referrer': request.META['HTTP_REFERER'],
+                },
+                'userAgent': request.META['HTTP_USER_AGENT'],
+            },
+        )
```
mikix: nit: feels like there's a bit of duplication in constructing these arguments with the below code. Could save context and properties as variables then pass them as parameters?

credentials/apps/records/views.py
```diff
@@ -311,13 +314,52 @@ class ProgramRecordCsvView(View):
...
         program_cert_record = get_object_or_404(ProgramCertRecord, uuid=kwargs.get('uuid'))
         program_certificate = get_object_or_404(ProgramCertificate, id=program_cert_record.certificate_id)
         record = get_record_data(program_cert_record.user, program_certificate.program_uuid, request.site)
 
+        segment_client.track(request.COOKIES['ajs_anonymous_id'],
+            event='edx.bi.credentials.program_record.download_started',
+            properties={
+                'category': 'records',
+                'program_uuid': str(program_certificate.program_uuid),
+                'record_uuid': str(program_cert_record.uuid),
```
mikix: I often see .hex instead of str() for this. I don't know if there's a difference.

When you pass those uuids below, you don't convert to a string, FYI.

credentials/apps/records/views.py
```diff
@@ -311,13 +314,52 @@ class ProgramRecordCsvView(View):
...
             raise http.Http404()
+
+        site_configuration = request.site.siteconfiguration
+        segment_client = SegmentClient(write_key=site_configuration.segment_key)
+
         program_cert_record = get_object_or_404(ProgramCertRecord, uuid=kwargs.get('uuid'))
         program_certificate = get_object_or_404(ProgramCertificate, id=program_cert_record.certificate_id)
         record = get_record_data(program_cert_record.user, program_certificate.program_uuid, request.site)
 
+        segment_client.track(request.COOKIES['ajs_anonymous_id'],
```
mikix: This is a blocking call I assume? And I bet they don't give us another, async option? Can this throw an exception we might want to catch?

jlajoie: It's non-blocking as per the docs!

> This library uses an internal queue to make identify and track calls non-blocking and fast.

Per https://segment.com/docs/sources/server/python/#track

mikix: Aw, that's a friendly design. Nice. Which also means there probably aren't any exceptions that get thrown.

(How does one notice an error?)

PR #467, opened 2018-07-17
LEARNER-5753: Allow sharing of records of partially completed programs

bbaker6225:
Allows sharing of partially completed program records by linking the model ProgramCertRecord to a program, rather than a program certificate.

Note:  Will not be doing the originally discussed 3 stage migration process due to the absence of production data.


bbaker6225:
@jlajoie 

@dianakhuang  talked to Fineal and I believe they came to the conclusion that it was fine to switch everything to 1 migration, since there isn't old data there, but @dianakhuang may be able to say more about that.

credentials/apps/records/models.py
```diff
@@ -27,15 +26,14 @@ class Meta(object):
 
 class ProgramCertRecord(TimeStampedModel):
     """
-    Connects a User with a Program Certificate
+    Connects a User with a Program
     """
-    certificate = models.ForeignKey(ProgramCertificate)
+    program = models.ForeignKey(Program, null=True)
```
mikix: I don't like allowing null Programs going forward. We can set some meaningless default for the migration maybe?

bbaker6225: I think that makes sense

PR #468, opened 2018-07-17
Show all program courses in program record view

spencerhance:
Updates the ProgramRecordView get request to return all program courses, even if they don't have a grade or cert for any associated course_runs.

If they don't have a verified cert, all table values except for Course name and org are left blank.

Additionally, the _Course_ title is used instead of the Course _run_ title when there are no certs for the associated course runs.

### DIFF
--- a/credentials/apps/records/views.py
+++ b/credentials/apps/records/views.py
@@ -3,10 +3,10 @@ import datetime
 import io
 import json
 import urllib.parse
-from analytics.client import Client as SegmentClient
 from collections import defaultdict
-import waffle
 
+import waffle
+from analytics.client import Client as SegmentClient
 from django import http
 from django.contrib.auth.mixins import AccessMixin, LoginRequiredMixin
 from django.contrib.contenttypes.models import ContentType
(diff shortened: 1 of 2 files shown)


credentials/apps/records/views.py
```diff
@@ -186,10 +186,23 @@ def _get_record(self, uuid, is_public):
...
         # Add course-run data to the response in the order that is maintained by the Program's sorted field
         course_data = []
+        added_courses = set()
         for course_run in program_course_runs:
             course = course_run.course
             grade = highest_attempt_dict.get(course)
-            if grade is not None and grade.course_run == course_run:
+
+            # If user hasn't taken this course yet, or doesn't have a cert, we want to show empty values
+            # The front end logic ignores course_id, letter_grade, and attempts if no issue_date is provided
```
mikix: optional nit: I would still pass those values along with empty strings or whatever, since the backend shouldn't assume which precise variables the frontend is going to inspect.

spencerhance: Ah good call!

spencerhance: Ah good call!

PR #469, opened 2018-07-17
LEARNER-5812

bbaker6225:
Add program key to ProgramCertRecord in preparation for removing the certificate field in 3 stages.


credentials/apps/records/models.py
```diff
@@ -29,13 +29,13 @@ class ProgramCertRecord(TimeStampedModel):
     """
     Connects a User with a Program Certificate
     """
-    certificate = models.ForeignKey(ProgramCertificate)
+    certificate = models.ForeignKey(ProgramCertificate, null=True)
+    program = models.ForeignKey(Program, null=True)
```
dianakhuang: I know we were talking about not making this nullable and just including a simple default. Did that not work out?

dianakhuang:
👍

PR #470, opened 2018-07-18
LEARNER-5813

bbaker6225:
Allows sharing of partially completed programs.  Also adds data migration for new program field in anticipation of preventing future programs from being nullable.  2nd stage of the migration process after [LEARNER-5812](https://github.com/edx/credentials/commit/a5c7c3048915681c2b2aac5fbbdc11eec327159a)


credentials/apps/records/migrations/0006_auto_20180718_1256.py
```diff
@@ -0,0 +1,27 @@
...
+from django.db import migrations
+
+from credentials.apps.catalog.models import Program
+from credentials.apps.credentials.models import ProgramCertificate
+from credentials.apps.records.models import ProgramCertRecord
+
+
+def seed_program_cert_records(apps, schema_editor):
+    for pcr in ProgramCertRecord.objects.all():
+        certificate = ProgramCertificate.objects.get(certificate=pcr.certificate)
```
bbaker6225: I'm worried about this line causing backwards compatibility issues, since we're trying to get rid of certificate eventually

dianakhuang: The migrations are meant to be run in order, so when this data migration is run, the table should still have the `certificate` available. The 3rd stage migrations that are created will be meant to be run after this one, so removing the field won't be an issue. (Happy to explain this more in person if it doesn't make sense.)

credentials/apps/records/migrations/0006_auto_20180718_1256.py
```diff
@@ -0,0 +1,27 @@
...
+        pcr.program = Program.objects.get(uuid=certificate.program_uuid)
```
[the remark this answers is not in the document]
bbaker6225: good catch, I'll fix this

PR #481, opened 2018-07-20
Prevent sending emails multiple times

spencerhance:
[LEARNER-5750](https://openedx.atlassian.net/browse/LEARNER-5750)

- Wires in the send to pathway modal
  - Appends 'Sent' to the end of sent pathways in the modal
- Adds new model for storing a record that an email was previously sent
- Updates ProgramSend logic to prevent an email from being sent more than once


credentials/apps/core/utils.py
```diff
@@ -10,3 +10,10 @@ def update_full_name(strategy, details, user=None, *_args, **_kwargs):
         if full_name and user.full_name != full_name:
             user.full_name = full_name
             strategy.storage.user.changed(user)
+
+
+def _choices(*values):
```
mikix: "Everybody's got choices"
https://www.youtube.com/watch?v=-zbw3KzgHBU (nsfw)

credentials/apps/records/views.py
```diff
@@ -33,6 +33,14 @@ def get_record_data(user, program_uuid, site, platform_name=None):
         program_course_runs = program.course_runs.all()
         program_course_runs_set = set(program_course_runs)
 
+        # Get all credit pathway organizations and their statuses
+        credit_pathways = program.pathways.all()
+        credit_pathways_set = set(credit_pathways)
```
mikix: This is a super minor thing, so just FYI. frozenset() is a slightly optimized version of set() when you only need a readonly one. And prevents accidentally editing the variable if you know it shouldn't change.

credentials/apps/records/models.py
```diff
@@ -44,3 +44,16 @@ def __str__(self):
 
     class Meta(object):
         verbose_name = "A viewable record of a program"
+
+
+class UserCreditPathway(TimeStampedModel):
```
mikix: How do you feel about an is_sent/status field or something? Rather than testing for existance? (To avoid us having to delete a row if we want to undo this state.)

I don't know proper database theory around this, but it seems to be a pattern that we never really delete rows. So if we ever wanted to disable this - sending failed for some reason (even though we don't monitor that now) or Support wants to turn this off so the learner can try again or something - Support can just toggle a checkbox rather than deleting a row.

I dunno. @jlajoie might be able to provide more clarity on whether deleting a row is an anti-pattern or just fine.

credentials/apps/records/views.py
```diff
@@ -289,6 +302,10 @@ def post(self, request, **kwargs):
         user = get_object_or_404(User, username=username)
         public_record, _ = ProgramCertRecord.objects.get_or_create(user=user, program=program)
 
+        # Make sure we haven't already sent an email
+        if UserCreditPathway.objects.filter(user=user, credit_pathway=pathway).exists():
+            return JsonResponse({'error': 'Pathway email already sent'}, status=403)
```
mikix: 400 might make more sense in this context?

credentials/static/components/ProgramRecord.jsx
```diff
@@ -26,10 +26,15 @@ class ProgramRecord extends React.Component {
     this.formatDate = this.formatDate.bind(this);
     this.formatGradeData = this.formatGradeData.bind(this);
     this.sendRecords = this.sendRecords.bind(this);
+    this.updateCreditPathwaysSent = this.updateCreditPathwaysSent.bind(this);
+    this.sendRecords = this.sendRecords.bind(this);
```
mikix: I hate all this binding nonsense.

credentials/static/components/ProgramRecord.jsx
```diff
@@ -127,6 +132,30 @@ class ProgramRecord extends React.Component {
...
+    }
+  }
+
+  // Parse the list of credit pathways into an object
+  parseCreditPathways() {
+    const creditPathways = {};
+
+    for (let i = 0; i < this.props.credit_pathways.length; i += 1) {
+      const pathway = this.props.credit_pathways[i];
+      const sent = (pathway.status !== '');
```
mikix: You prefer to treat unknown statuses as sent rather than just check for the 'sent' status and treat unknowns as unsent? That's a fine choice, I don't have a strong opinion, just wanted to call it out.

Alternatively, you could use a "sent" bool here instead of a status field?

credentials/static/components/ProgramRecord.jsx
```diff
@@ -366,6 +398,10 @@ class ProgramRecord extends React.Component {
             uuid={uuid}
             typeName={program.type_name}
             platformName={platformName}
+            // Using this JSON trick since the {...} syntax and Object.assign() both don't work here
+            // For some reason they were not properly creating copies
+            creditPathways={JSON.parse(JSON.stringify(this.creditPathways))}
```
mikix: I hate it, but good comment at least.

credentials/static/components/ProgramRecord.jsx
```diff
@@ -398,6 +434,7 @@ ProgramRecord.propTypes = {
     last_updated: PropTypes.string,
   }).isRequired,
   grades: PropTypes.arrayOf(PropTypes.object).isRequired,
+  credit_pathways: PropTypes.arrayOf(PropTypes.object).isRequired,
```
mikix: Required seems unnecessarily strict (and is unfriendly to any existing themes that might override the template with the ProgramRecordFactory). Why not just provide a default of []?

credentials/static/components/SendLearnerRecordModal.jsx
```diff
@@ -7,35 +7,70 @@ import StringUtils from './Utils';
...
-    this.setState({
-      [name]: !this.state[name],
-    });
+
+  getPathwayDisplayName(pathway) {
+    const name = pathway.name;
+
+    if (this.state.creditPathways[pathway.name].sent) {
+      const sent = gettext('Sent');
+      return StringUtils.interpolate('{name} - {sent}', { name, sent });
```
mikix: Optional nit: There is an argument that this should be piped to gettext as '{name} - Sent' since some languages would prefer to reorder maybe or use a different character than '-'. (Chinese often uses fullwidth versions of punctuation like periods, etc.)

credentials/static/components/SendLearnerRecordModal.jsx
```diff
@@ -98,6 +131,10 @@ SendLearnerRecordModal.propTypes = {
   uuid: PropTypes.string.isRequired,
   typeName: PropTypes.string.isRequired,
   platformName: PropTypes.string.isRequired,
+  // TODO: replace with redux global state variables
+  // eslint-disable-next-line react/forbid-prop-types
+  creditPathways: PropTypes.object.isRequired,
+  creditPathwaysList: PropTypes.arrayOf(PropTypes.object).isRequired,
```
mikix: Less important here, but I'd still make these optional.

credentials/static/components/Utils.jsx
```diff
@@ -28,12 +28,6 @@ class StringUtils {
     );
   }
 
-  // Parses a JSON string that contains encoded commas and escaped quotes around keys/values
-  // edx-user-info cookie is an example of where this is needed
-  static parseDirtyJSON(jsonString) {
-    return JSON.parse(jsonString.replace(/\\"/g, '"').replace(/\\054/g, ','));
-  }
-
```
mikix: Not needed anymore?

credentials/static/components/ProgramRecord.jsx
```diff
@@ -366,6 +398,10 @@ class ProgramRecord extends React.Component {
             uuid={uuid}
             typeName={program.type_name}
             platformName={platformName}
+            // Using this JSON trick since the {...} syntax and Object.assign() both don't work here
+            // For some reason they were not properly creating copies
+            creditPathways={JSON.parse(JSON.stringify(this.creditPathways))}
+            creditPathwaysList={this.props.credit_pathways}
```
mikix: Why do we need a list too? Sorting?

spencerhance: Yep that's the only reason I'm doing that.  If we don't need to preserve the API order or if there's a way to order an object in JS I could def remove it.

credentials/apps/records/models.py
spencerhance: Hm okay that does make a lot of sense.  I'll add a status field.

credentials/static/components/ProgramRecord.jsx
mikix: I don't know a better way I guess. Maybe a comment like "list for sorting" or something like that above it?

credentials/apps/records/models.py
```diff
@@ -44,3 +46,26 @@ def __str__(self):
...
+    """
+    Connects a user to a credit pathway
+    This is used to track when a user sends a record to that organization
+    The timestamp is used for error tracking and support
+    """
+    STATUS_CHOICES = [
+        (constants.UserCreditPathwayStatus.SENT, _('sent')),
```
mikix: Just a fun fact - Django does not help us at all here and does no protections against XSS attacks in its admin UI - a place where you'd think you'd want all of them. So our team actually has a history of not translating Django admin strings.

*But* since we recently discovered that we get baked-in xss attack protection (for translations) from i18n_tool being run on our translations, this is back to being a good thing.

PR #485, opened 2018-07-23
Deprecate certificate field on ProgramCertRecord

bbaker6225:
Add check in data migration to prevent error from grabbing a nonexistent certificate.  Additionally, mark certificate as deprecated, since it can not currently be removed since it is used in the data migration.


credentials/apps/records/models.py
```diff
@@ -28,8 +28,11 @@ class Meta(object):
 class ProgramCertRecord(TimeStampedModel):
     """
     Connects a User with a Program
+
+    Note: certificate is deprecated, and is kept around because it is used in
+    an old data migration.
     """
-    certificate = models.ForeignKey(ProgramCertificate, null=True)
+    certificate = models.ForeignKey(ProgramCertificate, null=True, default=None)
```
mikix: Add a help_text="" for it too explaining the deprecation (this will show up in django admin for example).

credentials/apps/records/models.py
```diff
@@ -28,8 +28,16 @@ class Meta(object):
 class ProgramCertRecord(TimeStampedModel):
     """
     Connects a User with a Program
+
+    Note: certificate is deprecated, and is kept around because it is used in
+    an old data migration.
```
mikix: Probably don't need this anymore now that you have help_text. But no matter.

credentials/apps/records/models.py
```diff
@@ -28,8 +28,16 @@ class Meta(object):
...
     """
-    certificate = models.ForeignKey(ProgramCertificate, null=True)
+    certificate = models.ForeignKey(
+        ProgramCertificate,
+        null=True,
+        default=None,
+        help_text='Note: certificate is deprecated, and is kept around because it is used in an old data migration.'
+    )
```
mikix: You might need a new migration for this, eh?

PR #488, opened 2018-07-23
Prevent program field on ProgramCertRecord from being null

bbaker6225:
Final stage of [LEARNER-5753](https://openedx.atlassian.net/browse/LEARNER-5753) migration process, since we currently can not remove the certificate field.


bbaker6225:
Going to wait until tomorrow morning to merge and deploy this, just in case there are any more unexpected issues that need to be dealt with

PR #490, opened 2018-07-25
Adds support for incomplete pathways

spencerhance:
**TICKET:** [LEARNER-5908](https://openedx.atlassian.net/browse/LEARNER-5908)

- Adds is_active flag  to the API that is False if the pathway doesn't have an email set
- Adds support in the UI to gray out checkbox and show org as "Not Yet Available" for inactive pathways
- Updates alert styling so alert body text matches the header
- Increases modal width

**Example alert:**
image

**Mock:**
image


credentials/static/components/SendLearnerRecordModal.jsx
```diff
@@ -1,20 +1,23 @@
...
-import { Button, CheckBoxGroup, CheckBox, Modal } from '@edx/paragon';
+import { Button, CheckBoxGroup, CheckBox, Modal, StatusAlert } from '@edx/paragon';
 import StringUtils from './Utils';
 
 class SendLearnerRecordModal extends React.Component {
   constructor(props) {
     super(props);
     this.checkCreditPathway = this.checkCreditPathway.bind(this);
     this.getCheckedOrganizations = this.getCheckedOrganizations.bind(this);
+    this.checkAnyInactivePathways = this.checkAnyInactivePathways.bind(this);
```
mikix: You don't need to bind here do you? You're not using it as a callback? I mean it's harmless, I'm just asking for myself here.

credentials/static/components/SendLearnerRecordModal.jsx
```diff
@@ -40,11 +43,27 @@ class SendLearnerRecordModal extends React.Component {
...
+  checkAnyInactivePathways() {
+    for (let i = 0; i < this.props.creditPathwaysList.length; i += 1) {
+      const pathway = this.props.creditPathwaysList[i];
+      if (!this.state.creditPathways[pathway.name].isActive) {
+        return true;
+      }
+    }
+
+    return false;
+  }
```
mikix: This method feels weird - iterating over a list to check into a dict. We can't iterate directly over the dict? Is that because foreach is somehow badly optimized?

credentials/apps/records/views.py
```diff
@@ -99,7 +99,9 @@ def get_record_data(user, program_uuid, site, platform_name=None):
 
         credit_pathway_data = [{'name': credit_pathway[0].name,
                                 'id': credit_pathway[0].id,
-                                'status': credit_pathway[1]} for credit_pathway in credit_pathways]
+                                'status': credit_pathway[1],
+                                'is_active': True if credit_pathway[0].email != '' else False}
```
mikix: The following is another option maybe?
'is_active': bool(credit_pathway[0].email),

credentials/static/components/SendLearnerRecordModal.jsx
spencerhance: ah no i'm pretty sure you're right.  I think I just did that out of habit.

credentials/static/components/SendLearnerRecordModal.jsx
spencerhance: So interestingly enough, ESLint forbids foreach loops and recommends using Object.keys(), which has the same overhead of lookups anyway.

PR #496, opened 2018-07-26
Add html version of email and more

mikix:
Also fix the admin UI a little bit and provide a username fallback for user's full name in emails.


credentials/apps/records/admin.py
```diff
@@ -27,4 +27,4 @@ class UserCreditPathwayAdmin(admin.ModelAdmin):
     """ Admin for UserCreditPathway """
     list_display = ('user', 'credit_pathway', 'status',)
     search_fields = ('user__username', 'credit_pathway__org_name', 'status',)
-    raw_id_fields = ('credit_pathway',)
+    raw_id_fields = ('user', 'credit_pathway',)
```
mikix: Without this, django will load all user rows into a dropdown or whatever. On prod, that takes down the system.

Django is dumb.

credentials/apps/records/migrations/0011_auto_20180726_1916.py
```diff
@@ -0,0 +1,19 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('records', '0010_auto_20180726_1334'),
+    ]
+
+    operations = [
+        migrations.AlterModelOptions(
+            name='programcertrecord',
+            options={'verbose_name': 'Shared program record'},
```
mikix: I know, I know. But django was adding an 's' at the end of these names in the admin UI. And my previous name was plural. So it looked dumb.

credentials/apps/records/views.py
```diff
@@ -325,7 +325,7 @@ def post(self, request, **kwargs):
                 'pathway_name': pathway.name,
                 'program_name': program.title,
                 'record_link': record_link,
-                'user_full_name': request.user.get_full_name(),
+                'user_full_name': request.user.get_full_name() or request.user.username,
```
spencerhance: That's pretty neat

PR #497, opened 2018-07-26
LEARNER-5925 Add rate limiting to credentials records endpoint

bbaker6225:
[LEARNER-5925](https://openedx.atlassian.net/browse/LEARNER-5925) Adds django-ratelimit throttling to the credentials records endpoints.

Note:
At this time, we're not adding a higher rate limit for staff/superusers because these endpoints correspond to the share record and send record buttons, which staff should not be using much more frequently than regular users.

### DIFF
--- a/credentials/apps/records/tests/test_views.py
+++ b/credentials/apps/records/tests/test_views.py
@@ -15,6 +15,7 @@ from django.test import TestCase
 from django.test.utils import override_settings
 from django.urls import reverse
 from mock import patch
+from testfixtures import LogCapture
 from waffle.testutils import override_flag
 
 from credentials.apps.catalog.tests.factories import (CourseFactory, CourseRunFactory, CreditPathwayFactory,
@@ -30,8 +31,9 @@ from credentials.apps.records.models import ProgramCertRecord, UserCreditPathway
 from credentials.apps.records.tests.factories import (ProgramCertRecordFactory, UserCreditPathwayFactory,
                                                       UserGradeFactory)
 from credentials.apps.records.tests.utils import dump_random_state
+from credentials.shared.log_checker import assert_log_correct
 
-from ..constants import WAFFLE_FLAG_RECORDS
+from ..constants import RECORDS_RATE_LIMIT, WAFFLE_FLAG_RECORDS
 
 JSON_CONTENT_TYPE = 'application/json'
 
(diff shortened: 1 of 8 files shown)


credentials/apps/records/tests/test_views.py
```diff
@@ -762,3 +764,89 @@ def tests_creates_csv(self, segment_client, track):  # pylint: disable=unused-ar
...
+                log,
+                'credentials.apps.records.views',
+                'WARNING',
+                'Credentials records endpoint is being throttled.',
+            )
+
+    def test_program_send_throttling(self):
+        """ Verify endpoint is throttled and a message is logged after limit. """
+        with LogCapture() as log:
+            pc = ProgramCertificateFactory(site=self.site, program_uuid=self.program.uuid)  # pylint: disable=unused-variable
```
dianakhuang: You can just leave this as `ProgramCertificateFactory(site=self.site, program_uuid=self.program.uuid)` without assigning it a variable if you want so that you don't have to wrestle with pylint disabling, etc.

bbaker6225: Good catch

credentials/apps/records/tests/test_views.py
```diff
@@ -762,3 +764,89 @@ def tests_creates_csv(self, segment_client, track):  # pylint: disable=unused-ar
...
+    """ Tests for throttling the records endpoint. """
+    USERNAME = "test-user"
+
+    def setUp(self):
+        super().setUp()
+        dump_random_state()
+        self.user = UserFactory(username=self.USERNAME)
+        self.client.login(username=self.user.username, password=USER_PASSWORD)
+        self.program = ProgramFactory(site=self.site)
+        self.rate_limit = self.parse_rate(RECORDS_RATE_LIMIT)
```
BenjiLee: Does this always take rate limit in minutes?

bbaker6225: It's duration agnostic

credentials/apps/records/views.py
```diff
@@ -26,7 +28,14 @@
...
 from credentials.apps.records.models import ProgramCertRecord, UserCreditPathway, UserGrade
 
-from .constants import WAFFLE_FLAG_RECORDS
+from .constants import RECORDS_RATE_LIMIT, WAFFLE_FLAG_RECORDS
+
+log = logging.getLogger(__name__)
+
+
+def rate_limited(request, exception):  # pylint: disable=unused-argument
+    log.warning("Credentials records endpoint is being throttled.")
```
BenjiLee: I think you were trying to match the format of the other logs such as `Credentials API endpoint CredentialViewSet is being throttled.` 

We should use a similar format for the new log.

credentials/apps/records/views.py
```diff
@@ -347,11 +360,15 @@ def dispatch(self, request, *args, **kwargs):
...
-class ProgramRecordCreationView(LoginRequiredMixin, View):
+class ProgramRecordCreationView(LoginRequiredMixin, RatelimitMixin, View):
     """
     Creates a new Program Certificate Record from given username and program uuid,
     returns the uuid of the created Program Certificate Record
     """
+    ratelimit_key = 'user'
+    ratelimit_rate = RECORDS_RATE_LIMIT
+    ratelimit_block = True
+    ratelimit_method = 'POST'
```
BenjiLee: I believe we can do this all in a decorator
https://django-ratelimit.readthedocs.io/en/v1.0.0/usage.html#use-as-a-decorator

bbaker6225: It can be done in a decorator.  Is there a reason why you prefer the decorator?  Just to more explicitly signal that we're only rate limiting the specific http method?

BenjiLee: I think that's how we do it in edx-platform. The argument here would be to keep it consistent.

BenjiLee: I cannot find proof that that's we do it in edx-platform. Feel free to keep this as is.

credentials/apps/records/views.py
bbaker6225: I don't think django-ratelimit has a built in way to detect which particular view is being throttled.  Right now the only way I can see to get the view is to get the requested url, and resolve that to a view, but I wasn't sure if we wanted to do that.

BenjiLee: Good to know. Let's leave this as is.

PR #502, opened 2018-07-27
Prevent factory calls from getting included twice

ColeLRogers:
Previously, in cases where a theme template exists, "default" blocks (without icons) for RecordsFactory and ProgramRecord factory were not getting overridden as expected when including the theme template. This meant that there were two calls for the factories on each page, rather than just the one for the themed template. This fixes that by only having the "default" blocks when the theme template doesn't exist. 

[LEARNER-5996](https://openedx.atlassian.net/browse/LEARNER-5996)

### DIFF
--- a/credentials/templates/programs.html
+++ b/credentials/templates/programs.html
@@ -22,17 +22,18 @@
 
 {% block javascript %}
   {% render_bundle 'programs' 'js' %}
-    {% block program_record_factory %}
-      <script type="text/javascript">
-        ProgramRecordFactory('program-record', {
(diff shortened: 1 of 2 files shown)


ColeLRogers:
@mikix It gets defined here https://github.com/edx/credentials/blob/master/credentials/apps/records/views.py#L220.
That checks if the theme template exists first, and returns either the path to it or an empty string if one doesn't exist.
There doesn't seem to be other instances where we do this, where there is a React call that might be overridden.

mikix:
:+1:

PR #505, opened 2018-07-30
Add site config for Records

mikix:
Allow site administrators to disable Learner Records on a site-by-site basis.

Oh yeah and I threw in a string change that got missed.

https://openedx.atlassian.net/browse/LEARNER-5987


credentials/static/components/ProgramRecord.jsx
```diff
@@ -367,6 +367,8 @@ class ProgramRecord extends React.Component {
           <div className="program-record-grades">
             <FoldingTable
               columns={[
+                // Note that when you change one of these strings, you should look at
```
spencerhance: 👍

PR #509, opened 2018-07-31
Show actual issue date

mikix:
We were using the wrong issue date for course certificates in the program record view. Now we use the visible_date attribute that the LMS provides.

And flesh out support for visible_date by not showing certs as awarded until the date has passed.

https://openedx.atlassian.net/browse/LEARNER-5995

### DIFF
--- a/credentials/apps/records/views.py
+++ b/credentials/apps/records/views.py
@@ -23,7 +23,8 @@ from ratelimit.mixins import RatelimitMixin
 from credentials.apps.catalog.models import CourseRun, CreditPathway, Program
 from credentials.apps.core.models import User
 from credentials.apps.core.views import ThemeViewMixin
-from credentials.apps.credentials.models import CourseCertificate, ProgramCertificate, UserCredential
+from credentials.apps.credentials.models import (CourseCertificate, ProgramCertificate, UserCredential,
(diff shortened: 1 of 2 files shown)


credentials/apps/records/views.py
```diff
@@ -97,15 +120,19 @@ def get_record_data(user, program_uuid, site, platform_name=None):
...
             course = course_run.course
             user_credential = user_credential_dict.get(course_run.key)
+            visible_date = visible_date_dict.get(course_run.key)
 
-            if user_credential is not None:
+            is_visible = not visible_date or visible_date < now
+            if is_visible and user_credential is not None:
                 num_attempts_dict[course] += 1
-                last_updated = max(last_updated, course_grade.modified) if last_updated else course_grade.modified
+                last_updated = max(filter(None, [visible_date, course_grade.modified, last_updated]))
```
spencerhance: Is there a need to use filter here if the function is None?

mikix: That filters out any falsy items in the list. So handles last_updated being None (first loop) and visible_date being None (no attribute in the db).

spencerhance: Huh cool, I didn't know you could use filter like that

PR #513, opened 2018-08-01
Temporary fix for translations of ES6/JSX files

spencerhance:
This PR is a temporary fix for a bug in xgettext (the library that django, and consequently, our i18n tool uses for string extractions) that breaks parsing when there are backticks in the file.

- Adds lint rule to prevent use of ES6 template strings (i.e. backticks)
- Converts all template strings in credentials to interpolations or concatenations

All this work should eventually be reverted once we switch to a different parsing library, or xgettext adds support.

Original ticket:
https://openedx.atlassian.net/browse/LEARNER-5085

Bugs Filed:
https://openedx.atlassian.net/browse/LEARNER-6038
https://github.com/edx/i18n-tools/issues/75


credentials/static/components/ProgramRecord.jsx
```diff
@@ -103,15 +103,19 @@ class ProgramRecord extends React.Component {
 
     const date = new Date(isoDate);
 
-    return `${date.getMonth() + 1}/${date.getDate()}/${date.getFullYear().toString().slice(2)}`;
+    return StringUtils.interpolate('{month}/{date}/{year}', {
```
dianakhuang: I think this is okay for now, but this date may need to become localized in the future for locales that use something besides the month/day/year format.

PR #516, opened 2018-08-02
Add user metadata to CSV

ColeLRogers:
[LEARNER-5986](https://openedx.atlassian.net/browse/LEARNER-5986)

Adds metadata to the record data csv

screen shot 2018-08-02 at 3 02 46 pm

### DIFF
--- a/credentials/apps/records/tests/test_views.py
+++ b/credentials/apps/records/tests/test_views.py
@@ -780,7 +780,7 @@ class ProgramRecordCsvViewTests(SiteMixin, TestCase):
     @patch('credentials.apps.records.views.SegmentClient', autospec=True)
     @patch('credentials.apps.records.views.SegmentClient.track', autospec=True)
     def tests_creates_csv(self, segment_client, track):  # pylint: disable=unused-argument
-        """ Verify that the csv parses and contains all of the necessary metadata and headers"""
+        """ Verify that the csv parses and contains all of the necessary titles/headers"""
         response = self.client.get(
             reverse('records:program_record_csv', kwargs={'uuid': self.program_cert_record.uuid.hex})
         )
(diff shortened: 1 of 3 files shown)


credentials/apps/records/tests/test_views.py
```diff
@@ -789,6 +789,11 @@ def tests_creates_csv(self, segment_client, track):  # pylint: disable=unused-ar
         content = response.content.decode('utf-8')
         csv_reader = csv.reader(io.StringIO(content))
         body = list(csv_reader)
+        metadata_titles = ['Program Name', 'Program Type', 'Platform Provider', 'Authoring Organization(s)',
+                           'Learner Name', 'Username', 'Email', '']
+        # check the title of each metadata row
+        for title in metadata_titles:
+            self.assertEqual(title, body.pop(0)[0])
```
dianakhuang: Do we need to check that the data for each of these rows matches what we expect as well?

credentials/apps/records/views.py
```diff
@@ -456,6 +456,27 @@ def get(self, request, *args, **kwargs):
...
+        program_type = program.type
+        platform = program.site
+        authoring_orgs = ','.join([org.name for org in program.authoring_organizations.all()])
+        user = program_cert_record.user
+        name = user.full_name
+        username = user.username
+        email = user.email
+
+        user_metadata = [
+            ['Program Name', program_name],
```
dianakhuang: It seems like maybe these should be translated, but I know we have weird issues translating CSV downloads.

credentials/apps/records/tests/test_views.py
ColeLRogers: I originally wasn't checking the data just to match the format of the rest of the test, but since it is pretty easy in this case, I have added it

ColeLRogers: Update: I realized that get_record_data is already compiling all of the data needed for this csv, so instead of doing the same queries, I'm now just getting the data from that call (which already takes place). As such, I've reverted back to just checking the titles, that way we don't have to start mocking the get_record_data call (which is done in other tests, so it is covered).

PR #522, opened 2018-08-08
Send updated email to pathway org when a program is completed

spencerhance:
If a user previously sent their learner record to a credit pathway partner, we want to send an updated email to the partner to let them know.

Wording confirmed by Marco

https://openedx.atlassian.net/browse/LEARNER-5361


package-lock.json
```diff
@@ -6,7 +6,7 @@
     "@edx/edx-bootstrap": {
```
bbaker6225: I think you want to remove your local changes to package-lock.json unless you actually added an NPM package.  @dianakhuang probably knows more though

credentials/apps/records/utils.py
```diff
@@ -0,0 +1,63 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def send_updated_emails_for_program(username, program_certificate):
+        """ If the user has previously sent an email to a pathway org, we want to send
+        an updated one when they finish the program.  This function is called from the
+        credentials Program Certificate awarding API """
+
+        # Try to get the user
```
bbaker6225: Super nit: This comment seems unnecessary

credentials/apps/records/utils.py
```diff
@@ -0,0 +1,63 @@
...
+        pathways_set = frozenset(program.pathways.all())
+
+        user_pathways = UserCreditPathway.objects.select_related('credit_pathway').filter(
+            user=user, credit_pathway__in=pathways_set, status=UserCreditPathwayStatus.SENT)
+
+        # Return here if the user doesn't have a program cert record
+        try:
+            pcr = ProgramCertRecord.objects.get(program=program, user=user)
+        except ProgramCertRecord.DoesNotExist:
+            logger.exception("Program Cert Record for user_uuid %s, program_uuid %s does not exist",
```
bbaker6225: Nit:  Probably preferable to use format instead of %s in the future, but not a big deal

credentials/apps/credentials/issuers.py
```diff
@@ -91,6 +92,45 @@ class ProgramCertificateIssuer(AbstractCredentialIssuer):
...
+            credential_content_type=ContentType.objects.get_for_model(credential),
+            credential_id=credential.id,
+            defaults={
+                'status': status,
+            },
+        )
+
+        # This function call should be moved into some type of task queue
+        # once credentials has that functionality
+        if created:
```
bbaker6225: Might be nice to have some indication here that it only sends the updated email if the learner already sent one.  I wouldn't change it if it's messy to move the logic out of the send email function though.

credentials/apps/records/utils.py
spencerhance: interestly enough, we have a lint rule that prefers using the logger "printf style" formatting over .format when doing logging

credentials/apps/records/utils.py
spencerhance: whoops good catch, left over from a try-except

package-lock.json
spencerhance: Good catch, also my bad

PR #528, opened 2018-08-10
Webpack Caching

spencerhance:
Adds simple caching for webpack builds

**Before:**
real	1m12.901s
user	1m42.310s
sys	0m3.560s

 **After:**
real	0m39.541s
user	0m53.630s
sys	0m2.530s

~45% perf increase

Experimentally, Travis doesn't keep a persistent cache so it doesn't affect its builds.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
     "@edx/edx-bootstrap": {
       "version": "1.0.0",
       "resolved": "https://registry.npmjs.org/@edx/edx-bootstrap/-/edx-bootstrap-1.0.0.tgz",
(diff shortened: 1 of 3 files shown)


spencerhance:
@mikix I'm inclined to agree with you, it's hard to trust NPM in general.  Although I am seeing a different number of dependencies https://www.npmjs.com/package/hard-source-webpack-plugin, and all of the non dev-dependencies are pretty widely used.

mikix:
Yes, I meant 68 not 63, and I meant dependents, not dependencies.

Yeah, it's fine I'd say. Risk is low.

jibsheet:
There appear to be issues with our sandbox hourly build tests based on this change
Cannot find module 'hard-source-webpack-plugin
https://tools-edx-jenkins.edx.org/job/Sandboxes/job/CreateSandbox/12915/console

PR #531, opened 2018-08-13
Add uuid field to CreditPathway model

ColeLRogers:
Adds a new uuid field to the CreditPathway model, reflecting a [similar change](https://github.com/edx/course-discovery/pull/1498) to be made in course-discovery.
This is an "in between" change. The uuid field here is nullable, and parse_pathway will only assign it if a uuid exists in the pathway retrieved from course-discovery.
Once the course-discovery change is live which will give every pathway a uuid, I will manually call copy_catalog, then create another PR will which make this field non-nullable.


credentials/apps/catalog/models.py
```diff
@@ -88,14 +88,15 @@ def __str__(self):
...
+    uuid = models.UUIDField(verbose_name='UUID', null=True)  # temporary, will be unique in future
     name = models.CharField(max_length=255)
     org_name = models.CharField(max_length=255)
     email = models.EmailField()
     programs = SortedManyToManyField(Program, related_name='pathways')
 
     class Meta:
         unique_together = (
-            ('site', 'name')
+            ('site', 'name', 'uuid')
```
dianakhuang: I think should should just be unique on site and the uuid, even if it's somewhat equivalent right now.

ColeLRogers: The reasoning behind this is so that it can handle the case where Discovery doesn't yet have uuids, and when it does. When it doesn't have uuids, there is a constraint on site and name, whereas with uuids it only uses the uuids. By including all three, it handles both versions of discovery, since uuids will be null if the old constraint is in place, and uuids are unique on themselves so if the new change is in place, the composite would also be unique.

dianakhuang: That's reasonable. I hadn't seen your newer PR that removes this constraint entirely when I wrote that.

PR #534, opened 2018-08-13
Update admin documentation

bbaker6225:
[LEARNER-6107](https://openedx.atlassian.net/browse/LEARNER-6107)


docs/credentials_admin.rst
```diff
@@ -94,3 +98,111 @@ The following image shows the ``ProgramCertificate`` model in the Django adminis
...
+This functionality adds student records, along with the grades and pathways associated with a record.  
+
+Shared Program Records
+~~~~~~~~~~~~~~~~~~~~~~
+
+The ``seed-records`` management command generates example program records for use in development. Run
+``./manage.py seed-records --help`` to view all of the parameters.  To add or update a shared program record, use
+the ``ProgramCertRecord`` model.
+
+The following image shows a configured ``ProgramCertRecord`` model in the Django administration panel.
```
mikix: Might be worth explaining what this model does? It records the uuids used for publicly shared views of a record. To manually unshare a public view, an admin can delete an entry in this table.

docs/credentials_admin.rst
```diff
@@ -94,3 +98,111 @@ The following image shows the ``ProgramCertificate`` model in the Django adminis
...
+
+
+User Grades
+~~~~~~~~~~~
+
+The ``seed-records`` management command generates example grades as part of the creation of program records for
+use in development. Run ``./manage.py seed-records --help`` to view all of the parameters.  To add or update user
+grades, use the ``UserGrade`` model.
+
+The following image shows a configured ``UserGrade`` model in the Django administration panel.
```
mikix: Might be worth explaining that we keep grades in order to show the highest grade earned in student records? And that they come in from the LMS whenever a grade is updated?

docs/credentials_admin.rst
```diff
@@ -94,3 +98,111 @@ The following image shows the ``ProgramCertificate`` model in the Django adminis
...
+
+.. _Waffle: https://github.com/jsocol/django-waffle
+.. _Waffle documentation: https://waffle.readthedocs.io/en/stable/
+
+
+Permanent Feature Rollout
+~~~~~~~~~~~~~~~~~~~~~~~~~
+Over time some features may become permanent and no longer need a feature gate around them. In such instances, the
+relevant code and tests should be updated to remove the feature gate. Once the code is released, the feature flag/switch
+should be deleted.
```
mikix: This whole section is weird. Just saying "waffle exists" is useless to anyone that isn't doing active development. Whereas this doc is focused at administrators, not developers.

But documenting what waffle flags we have going right now would be worth documenting for admins?

docs/credentials_admin.rst
```diff
@@ -1,6 +1,10 @@
 Credentials Administration
 ==========================
 
+
+Site Configuration
+------------------
+
```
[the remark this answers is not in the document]
bbaker6225: 👍 Good catch

docs/credentials_admin.rst
```diff
@@ -94,3 +98,111 @@ The following image shows the ``ProgramCertificate`` model in the Django adminis
...
+Records
+-------
+
+This functionality adds student records, along with the grades and pathways associated with a record.  
+
+Shared Program Records
+~~~~~~~~~~~~~~~~~~~~~~
+
+The ``seed-records`` management command generates example program records for use in development. Run
+``./manage.py seed-records --help`` to view all of the parameters.  To add or update a shared program record, use
```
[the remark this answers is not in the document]
bbaker6225: You're right, probably best to just remove the reference to seed-records here completely, and just cover it in the management docs.

docs/credentials_admin.rst
bbaker6225: At this point I think we've gotten rid of all waffle flags, so maybe a waffle section is redundant now?

mikix: Correct. I don't see any waffle flags in there now.

mikix: I'd say just drop it altogether.

docs/credentials_admin.rst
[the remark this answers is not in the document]
bbaker6225: Just noticed that the site configuration model screenshot is very out of date, and there is no way we can fit all of the current fields into a new screenshot.  How would you feel about not having a screenshot for this particular model configuration?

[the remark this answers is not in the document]
mikix: I don't think we need to fit all current fields in there. Just a glimpse of what they should be looking at is good enough. And in that regard, an out of date one doesn't kill me either. But might as well throw in an update there while you're touching it.

I'd say just include some of the fields at the top is good enough.

PR #539, opened 2018-08-15
Make CreditPathway's uuid field required

ColeLRogers:
This is a follow up to a [previous PR](https://github.com/edx/credentials/pull/531) which added this field as nullable. Once the [discovery PR ](https://github.com/edx/course-discovery/pull/1498) is deployed, I will call the copy_catalog command such that all pathways being copied have uuids, then I will merge this PR which requires them. 

For purposes of reviewing, only look at the last commit of this PR, as it is rebased off of the first credentials PR.


credentials/apps/catalog/models.py
```diff
@@ -88,15 +88,11 @@ def __str__(self):
...
+    uuid = models.UUIDField(verbose_name='UUID', unique=True)
     name = models.CharField(max_length=255)
     org_name = models.CharField(max_length=255)
     email = models.EmailField()
     programs = SortedManyToManyField(Program, related_name='pathways')
 
-    class Meta:
-        unique_together = (
-            ('site', 'name')
-        )
```
dianakhuang: I think that it might be necessary to have this unique on site and uuid. We should ask @mikix if that's a requirement.

mikix: Yes please!

mikix: Specifically, make uuid non-unique and unique-together site&uuid.

PR #547, opened 2018-08-20
Add records documentation for admins

mikix:
And a few miscellaneous fixes along the way.

https://openedx.atlassian.net/browse/LEARNER-6110

Credentials Pull Request
---

  - [x] Request a review if desired
  - [x] Squash/Fixup your branch to one commit
    - [x] Make sure you have updated [edx/configuration](https://github.com/edx/configuration) and [edx/devstack](https://github.com/edx/devstack) if necessary
    - [x] Make sure the change builds successfully in a sandbox
  - UI Changes 
    - [x] Consider other browsers (e.g. Firefox)
    - [x] Check mobile view
    - [x] Consider accessibility (e.g. Run aXe on the page)

### DIFF
--- a/credentials/apps/catalog/management/commands/copy_catalog.py
+++ b/credentials/apps/catalog/management/commands/copy_catalog.py
@@ -32,7 +32,10 @@ class Command(BaseCommand):
         for site in Site.objects.all():
             site_configs = SiteConfiguration.objects.filter(site=site)
             site_config = site_configs.get() if site_configs.exists() else None
-            if not site_config or not site_config.catalog_api_url:
+
+            # We skip the site if records_enabled is false - remember to remove that check once we start
+            # using the catalog data for certificates too.
+            if not site_config or not site_config.catalog_api_url or not site_config.records_enabled:
                 logger.info('Skipping site {}. No configuration.'.format(site.domain))
                 continue
 
--- a/credentials/settings/base.py
+++ b/credentials/settings/base.py
@@ -94,7 +94,7 @@ ROOT_URLCONF = 'credentials.urls'
 WSGI_APPLICATION = 'credentials.wsgi.application'
 
 # Database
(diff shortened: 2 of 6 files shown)


credentials/apps/catalog/management/commands/copy_catalog.py
```diff
@@ -32,7 +32,7 @@ def handle(self, *args, **options):
         for site in Site.objects.all():
             site_configs = SiteConfiguration.objects.filter(site=site)
             site_config = site_configs.get() if site_configs.exists() else None
-            if not site_config or not site_config.catalog_api_url:
+            if not site_config or not site_config.catalog_api_url or not site_config.records_enabled:
```
mikix: I noticed that even if a site disabled records support, we were still copying the catalog for it.

We eventually want to use the copy of the catalog for things that aren't student records, but we don't do that today. So it seems like just a waste of space/time to do the copy today.

I'm happy to take this out if y'all don't like the change.

credentials/settings/base.py
```diff
@@ -94,7 +94,7 @@
 WSGI_APPLICATION = 'credentials.wsgi.application'
 
 # Database
-# https://docs.djangoproject.com/en/1.8/ref/settings/#databases
+# https://docs.djangoproject.com/en/1.11/ref/settings/#databases
```
mikix: I noticed that a related doc had an old reference. So I went through the code and updated these.

credentials/apps/catalog/management/commands/copy_catalog.py
bbaker6225: Makes sense to me.  I think it would be nice to have a comment here reminding us to eventually get rid of it when we use copy catalog for other things

mikix: Good point. Will do.

PR #549, opened 2018-08-20
Add pathway foreign key to UserCreditPathway

bbaker6225:
Credentials Pull Request

Preparation to eventually delete credit_pathway foreign key from UserCreditPathway.
Adds non-nullable pathway foreign key to UserCreditPathway in 3 steps:

1. Add with null=True
2. Data migration to update pathway with credit_pathway data
3. Change null=True to null=False (This will come in a future release)

---

  - UI Changes 
    - [ ] Check mobile view


bbaker6225:
Removed the extra migration that will need to happen later

credentials/apps/records/migrations/0014_auto_20180820_2032.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.15 on 2018-08-20 20:32
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+def seed_user_credit_pathways(apps, schema_editor):
```
bbaker6225: Updated this seeding function

bbaker6225: @mikix Wanted to check in with you about my change to this data migration that is based off what we talked about

mikix: There is a theoretical gap here. Where CreditPathways exist in the db but Pathways don't because copy_catalog hasn't been run recently.

If you wanted to not raise an exception while migrating under that condition, you could use filter().first() instead of get(). But that will still leave your DB in an odd condition that you can't easily recover from (null pathways).

So maybe it's better to fail noisily indeed here.

mikix: All of that is to say :+1:

PR #552, opened 2018-08-22
Adding overview docs

mikedikan:
https://openedx.atlassian.net/browse/LEARNER-6112

Updating overview docs to reflect current state of the system.

FYI: @edx/learner-growth


docs/overview.rst
```diff
@@ -22,39 +22,74 @@ Credential Assets
...
 Interactions
 ============
 
-Internal: Credential Configuration and Administration
------------------------------------------------------
-.. image:: _static/images/overview-interactions-internal.png
+Internal Users: Credential Configuration and Administration
+-----------------------------------------------------------
+.. image:: _static/images/overview-credentials-config.png
+Editable in `Lucidchart <https://www.lucidchart.com/documents/edit/3e27073a-cafc-4673-b8ed-be1553094faf/0>`_
```
mikix: Do you maybe want these "editable in" lines to be non-visible comments? (Can you do that in rst?)

mikedikan: non-visible comments?

mikix: Like, that link isn't useful for the average reader of these docs. They're only useful for people editing these docs. So they should be like code comments. Only visible in the source, not in the rst output.

Apparently you can do that in rst by prefixing the line with ".. "

http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#comments

mikedikan: AHH, gotcha - will do!

PR #556, opened 2018-08-22
Add a program-listing view

mikix:
This adds a new URL, /program-listing/, which shows all active programs, if you are a superuser. If you aren't, it shows a 404.

https://openedx.atlassian.net/browse/LEARNER-6115

Credentials Pull Request
---

  - UI Changes 
    - [ ] Check mobile view


credentials/apps/records/admin.py
```diff
@@ -18,7 +18,7 @@ class ProgramCertRecordAdmin(admin.ModelAdmin):
 class UserGradeAdmin(admin.ModelAdmin):
     """ Admin for the UserGrade model. """
     list_display = ('username', 'course_run', 'letter_grade', 'percent_grade',)
-    search_fields = ('username', 'course_run__key', 'mode',)
+    search_fields = ('username', 'course_run__key',)
```
mikix: Unrelated bug that Bill found. Prevents searching in the admin interface for this model.

credentials/apps/catalog/management/commands/copy_catalog.py
```diff
@@ -50,7 +50,8 @@ def handle(self, *args, **options):
     def fetch_programs(site, client, page_size=None):
         next_page = 1
         while next_page:
-            programs = client.programs.get(exclude_utm=1, page=next_page, page_size=page_size)
+            programs = client.programs.get(exclude_utm=1, published_course_runs_only=1, page=next_page,
+                                           page_size=page_size)
```
mikix: This was not necessary, but I just noticed that we weren't passing it, when I started looking at program statuses. I don't believe we want unpublished course runs?

It doesn't seem to stop unpublished programs coming through though. (Though maybe it will if all the courses inside are unpublished? My devstack data is kinda all over the place.)

bbaker6225:
Could you make a sandbox for this??

mikix:
I'll try. I don't know how easy it will be to pull in catalog data. It might just be easier to try locally? But I'll investigate the sandbox way.

bbaker6225:
Ah didn't think about that.  Don't worry about the sandbox for now.  I'll test it locally

credentials/apps/records/views.py
```diff
@@ -250,36 +271,73 @@ def _get_programs(self, request):
...
+        context = super().get_context_data(**kwargs)
+
+        context['title'] = _('Program Listing View')
+        context['program_help'] = _('The following is a list of all active programs for which program records '
+                                    'are being generated.')
+
+        return context
+
+    def dispatch(self, request, *args, **kwargs):
+        # Kick out superusers (anonymous users are redirected to a login screen instead)
```
bbaker6225: We don't want to kick out superusers here do we? Is this comment just a typo?

bbaker6225:
I'm seeing an issue where the programs are showing up as not completed, even though all the courses are completed, right after I run seed-records.  I dropped and recreated the database, so it shouldn't be just me.

screen shot 2018-08-23 at 4 32 12 pm

mikix:
@bbaker6225 program completion is based on whether there is a program cert or not... wonder if seed-records doesn't do that or not. Will look.

bbaker6225:
@mikix My only other comment right now is that the new view obviously looks different from the mockup since we're just reusing most of the main records view.  That's fine with me, I just wanted to check and make sure we were ok with that.

Mock:
programlisting view

mikix:
@bbaker6225 the lack of "View Example" is a known gap that is covered by another ticket. The lack of a footer in the mock, I assumed was just a Marco oversight/assumption. Seems like we always would want copyright info, etc.

bbaker6225:
@mikix  Yup, I know view example shouldn't be here, I just wanted to check that we're ok with reusing the language from the default view instead of "Program Listing View" and "The following is a list of all active programs for which program records are being generated"

credentials/apps/records/views.py
```diff
@@ -91,6 +91,7 @@ def get_record_data(user, program_uuid, site, platform_name=None):
         # Find program credential if it exists (indicates if user has completed this program)
         program_credential_query = UserCredential.objects.filter(
             username=user.username,
+            status=UserCredential.AWARDED,
```
mikix: This and the other status=AWARDED are unrelated. Just something I noticed.

PR #559, opened 2018-08-23
Add pathway_type field to Pathway model

bbaker6225:
Field used to distinguish between credit pathways and professional
pathways.  Will accompany a similar model change in discovery.

Credentials Pull Request
---

  - UI Changes 
    - [ ] Check mobile view


credentials/apps/records/management/commands/seed-records.py
```diff
@@ -326,3 +328,17 @@ def seed_user_credit_pathways(user, pathways):
...
+
+    @staticmethod
+    def seed_industry_pathway(site, programs, faker):
+        """ Seed one industry pathway """
+        industry_pathway, created = Pathway.objects.get_or_create(
+            pathway_type=PathwayType.INDUSTRY.value,
+            site=site,
+            name='Test industry pathway',
+            org_name='Dunder Mifflin',
+            uuid=faker.uuid4())
```
mikix: I know you were following the pattern already in that file, but I'd prefer these blocks to use the following pattern instead:

        industry_pathway, created = Pathway.objects.update_or_create(
            site=site,
            uuid=faker.uuid4(),
            defaults={
                 'name': 'Test industry pathway',
                 'org_name': 'Dunder Mifflin',
                 'pathway_type': PathwayType.INDUSTRY.value,
            },
        )

That way, it's easier to add/adjust fields in the future. (You don't need to adjust the whole file, but at least for new blocks...)

credentials/apps/records/models.py
```diff
@@ -68,5 +70,14 @@ class UserCreditPathway(TimeStampedModel):
         blank=True,
     )
 
+    def clean(self):
+        # Don't allow pathway to have any type other than the CREDIT type
+        if self.pathway.pathway_type != PathwayType.CREDIT.value:
+            raise ValidationError({'pathway': _('User credit pathways can only be connected to credit pathways.')})
```
mikix: You can also add a validation (only for the admin page unfortunately) directly to the foreignkey using limit_choices_to:

https://docs.djangoproject.com/en/1.11/ref/models/fields/#django.db.models.ForeignKey.limit_choices_to

I wish that also did the clean() style check for you too...

credentials/shared/constants.py
```diff
@@ -0,0 +1,7 @@
+from enum import Enum
+
+
+class PathwayType(Enum):
+    """ Allowed values for Pathway.pathway_type """
+    CREDIT = 'Credit'
+    INDUSTRY = 'Industry'
```
mikix: Capitalized?

credentials/static/components/ProgramRecord.jsx
```diff
@@ -145,20 +145,22 @@ class ProgramRecord extends React.Component {
...
       const pathway = this.props.pathways[i];
       const sent = (pathway.status === 'sent');
 
-      creditPathways[pathway.name] = {
-        sent,
-        checked: sent,
-        id: pathway.id,
-        isActive: pathway.is_active,
-      };
+      if (pathway.pathway_type === 'Credit') {
```
mikix: We could also just filter at the view level and keep the UI dumb? But I guess if we want to show both in the UI at some point, might as well feed the data up now.

Just feels weird to knowingly give the UI stuff it doesn't care about. But really, that's me coupling the backend and frontend too tightly. Good decision here. :)

credentials/apps/records/management/commands/seed-records.py
bbaker6225: 👍

credentials/apps/records/models.py
bbaker6225: Since we have a raw id field for pathway on UserCreditPathway in admin, using the limit_choices_to restriction gives the same functionality but gives the more generic: "Select a valid choice. That choice is not one of the available choices." instead of "User credit pathways can only be connected to credit pathways.".

mikix: Got it. Alright never mind. I'd never really mucked with that option before, just new about it.

mikix:
About the capitalized types -- we should probably match what discovery gives us, I was just asking about it because it seemed unusual to me (for a slug-type enum, rather than something shown to the user). What does discovery give (or is that undecided yet)?

bbaker6225:
@mikix Discovery doesn't give anything right now.  I'm in the process of adding the pathway_type to the discovery pathway model now, so we can pick whatever we want and just make it match on both ends

PR #565, opened 2018-08-28
Filter program certs by visible_date

mikix:
Whenever we show a program credential, make sure that it is actually supposed to be shown. We rely on the visible_date field for program credentials to be actually populated by the LMS for this feature.

https://openedx.atlassian.net/browse/LEARNER-6263

Credentials Pull Request
---

  - UI Changes 
    - [ ] Check mobile view

### DIFF
--- a/credentials/apps/credentials/tests/test_utils.py
+++ b/credentials/apps/credentials/tests/test_utils.py
@@ -1,15 +1,16 @@
+import datetime
 import ddt
 from django.contrib.auth import get_user_model
 from django.test import TestCase
 
-from credentials.apps.credentials.utils import validate_duplicate_attributes
+from credentials.apps.credentials.utils import datetime_from_visible_date, validate_duplicate_attributes
 
 User = get_user_model()
 
 
 @ddt.ddt
-class ValidateDuplicateAttributesTests(TestCase):
-    """ Tests for Validate the attributes method """
+class CredentialsUtilsTests(TestCase):
+    """ Tests for credentials.utils methods """
 
     def test_with_non_duplicate_attributes(self):
         """ Verify that the function will return True if no duplicated attributes found."""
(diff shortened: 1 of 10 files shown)


credentials/apps/records/management/commands/seed-records.py
```diff
@@ -67,9 +68,9 @@ def seed_all(site_name, username):
         programs = Command.seed_programs(site, organizations, course_runs, faker)
         user = Command.get_user(username)
         Command.seed_user_grades(user, course_runs)
-        Command.seed_signatories(organizations)
-        course_certificates = Command.seed_course_certificates(site, course_runs)
-        program_certificates = Command.seed_program_certificates(site, programs)
+        signatories = Command.seed_signatories(organizations)
+        course_certificates = Command.seed_course_certificates(site, course_runs, signatories)
+        program_certificates = Command.seed_program_certificates(site, programs, signatories)
```
mikix: The changes in this file (adding signatories & setting is_active=True / certificate_type=verified for course certs) are not actually related to the rest of this PR.

Just some gaps I noticed when in the admin page.

credentials/apps/credentials/tests/test_utils.py
```diff
@@ -28,3 +29,11 @@ def test_with_duplicate_attributes(self):
         ]
 
         self.assertFalse(validate_duplicate_attributes(attributes))
+
+    def test_datetime_from_visible_date(self):
+        """ Verify that we convert LMS dates correctly. """
+        self.assertIsNone(datetime_from_visible_date(''))
+        self.assertIsNone(datetime_from_visible_date('2018-07-31'))
+        self.assertIsNone(datetime_from_visible_date('2018-07-31T09:32:46+00:00'))  # should be Z for timezone
```
bbaker6225: I think you wanted `2018-07-31T09:32:46:00:00` instead of `2018-07-31T09:32:46+00:00`? (Colon instead of plus)

mikix: The plus is a timezone (in this case, UTC). It's the isoformat() style

bbaker6225: Ah good to know

bbaker6225:
@mikix This looks good to me, but it's probably worth getting feedback from Diana too since she has a bit more context than I do.

mikix:
@dianakhuang ah yeah, that's because that's all already done for us. The UserCredential creation endpoint already allows passing in attributes, of which visible_date is one. LMS uses that for course certificates' visible_date.

So now the only change is that we also look at that attribute for program UserCredentials too.

PR #577, opened 2018-09-17
Remove rate limit tests that have become flaky.

bbaker6225:
Credentials Pull Request
---

  - UI Changes 
    - [ ] Check mobile view

### DIFF
--- a/credentials/apps/api/v2/tests/test_views.py
+++ b/credentials/apps/api/v2/tests/test_views.py
@@ -3,10 +3,10 @@ from decimal import Decimal
 
 import ddt
 from django.contrib.auth.models import Permission
+from django.test import TestCase
 from django.urls import reverse
 from rest_framework.renderers import JSONRenderer
 from rest_framework.test import APIRequestFactory, APITestCase
-from testfixtures import LogCapture
 
 from credentials.apps.api.v2.serializers import (UserCredentialAttributeSerializer, UserCredentialSerializer,
                                                  UserGradeSerializer)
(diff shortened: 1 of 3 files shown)


BenjiLee:
@bbaker6225 Is there a way to expliclty set the rate limit to 1? That way we can test expected behaviour when it does get hit.

bbaker6225:
@BenjiLee We can override the rate limit in the tests, but since they're hitting the actual endpoint they will still use the limit that is set on the actual classes, which I don't think we found a way to override when we tried to do this the first time.

credentials/apps/api/v2/tests/test_views.py
```diff
@@ -503,123 +500,3 @@ def test_update(self, method):
...
-        """ Login as the given user. """
-        self.client.logout()
-        self.client.login(username=user.username, password=USER_PASSWORD)
-
-    def add_user_permission(self, user, permission):
-        """ Assigns a permission of the given name to the user. """
-        user.user_permissions.add(Permission.objects.get(codename=permission))
-
-    @ddt.data('credential_view', 'grade_view', 'staff_override')
-    def test_throttle_configuration(self, scope):
```
BenjiLee: We need to keep this test to verify that there is a rate limit set.

bbaker6225: Added back in

BenjiLee: Do we need a similar test for the other endpoint Records?

bbaker6225: We couldn't find a good way to do it last time, since the other endpoints weren't DRF

PR #584, opened 2018-09-28
Update base.html to include multiple orgs names for orgs on certs.

calistenson:
https://openedx.atlassian.net/browse/LEARNER-5661?devStatusDetailDialog=branch

Credentials Pull Request
---

  - UI Changes 
    - [ ] Check mobile view


credentials/apps/credentials/views.py
```diff
@@ -70,10 +71,22 @@ def get_context_data(self, **kwargs):
...
                                                   format(program_uuid=program_details.uuid))
 
+        org_name_string = ""
+        if len(organization_names) == 1:
+            org_name_string = organization_names[0]
+        elif len(organization_names) == 2:
+            org_name_string = organization_names[0] + " and " + organization_names[1]
+        else:
+            for org in organization_names[:-1]:
+                org_name_string += org + ", "
```
mikix: org_name_string = ', '.join(organization_names[:-1]

(I also switched to single quotes, since my understanding is that those are stylistically preferred over double quotes.)

credentials/apps/credentials/views.py
```diff
@@ -70,10 +71,22 @@ def get_context_data(self, **kwargs):
...
+            organization_names.append(organization.display_name)
             if not organization.certificate_logo_image_url:
                 raise MissingCertificateLogoError('No certificate image logo defined for program: [{program_uuid}]'.
                                                   format(program_uuid=program_details.uuid))
 
+        org_name_string = ""
+        if len(organization_names) == 1:
+            org_name_string = organization_names[0]
+        elif len(organization_names) == 2:
+            org_name_string = organization_names[0] + " and " + organization_names[1]
```
mikix: This and the section below both need translating. Give translators a string like '{first} and {second}' or whatever and format it afterward.

credentials/apps/credentials/views.py
```diff
@@ -70,10 +71,22 @@ def get_context_data(self, **kwargs):
...
+        else:
+            for org in organization_names[:-1]:
+                org_name_string += org + ", "
+            org_name_string += "and "
+            org_name_string += organization_names[-1]
```
mikix: What about an empty name list? The old code didn't handle it well either (probably threw a javascript error). But this here is going to throw an IndexError and give the user a 500, which is not the most graceful error state.

Maybe just have the last else condition be 'else > 0' and leave the name string empty in the 0 case? At least that way the certificate renders. (Though obviously that is a data entry problem, not something you can solve here.)

credentials/apps/credentials/views.py
```diff
@@ -70,10 +72,20 @@ def get_context_data(self, **kwargs):
...
+            organization_names.append(organization.display_name)
             if not organization.certificate_logo_image_url:
                 raise MissingCertificateLogoError('No certificate image logo defined for program: [{program_uuid}]'.
                                                   format(program_uuid=program_details.uuid))
 
+        org_name_string = ""
+        if len(organization_names) == 1:
+            org_name_string = pgettext("One organization", organization_names[0])
+        elif len(organization_names) == 2:
+            org_name_string = pgettext("Two organizations", organization_names[0] + " and " + organization_names[1])
```
mikix: Consider how this will end up hitting gettext. Gettext works by taking an English string and looking up in its catalog if it has a translation for that exact string.

So here... gettext will end up seeing "MIT and Harvard" and won't be able to find a match for that. This is why when you use gettext(), you tend to use format strings like you do below, and then call format on the *returned and already-translated* string from gettext.

credentials/apps/credentials/views.py
```diff
@@ -70,10 +72,20 @@ def get_context_data(self, **kwargs):
...
+        elif organization_names:
+            org_name_string = pgettext("More than two organizations", "{first_orgs} and {last_org}").format(
+                first_orgs=', '.join(organization_names[:-1]), last_org=organization_names[-1])
```
mikix: You're missing the Oxford comma here, which is either a mistake or intentional, in which case that's a whole discussion.

Another thought is that you could maybe avoid the whole pgettext stuff if you make sure to use different format strings for the first part, like below.  That way, there will be two catalog entries without having to give a context.  (I also switched to single quotes, as a style thing. We tend to prefer that for Python code, though I'm not sure if it rises to actual style policy.)

```
# Translators: two organizations
_('{first_org} and {second_org}')

# Translators: a series of at least three organizations
_('{series_of_orgs}, and {last_org}')
```

(I have no idea if the Oxford comma is language-dependent... That's an interesting question. I left it in the translated string just in case.)

credentials/conf/locale/en/LC_MESSAGES/django.po
```diff
@@ -225,6 +225,17 @@ msgid ""
 "I completed a course at {platform_name}. Take a look at my certificate:"
 msgstr ""
 
+#: apps/credentials/views.py
+msgctxt "Two organizations"
+msgid " and "
+msgstr ""
```
mikix: Notice how the extractor didn't quite get this right.

PR #623, opened 2018-12-28
Remove use of edx-pattern-library

mikix:
Convert all sass code to bootstrap instead of e-p-l, and adjust dependencies (you'll notice we've dropped four).

This does not affect credentials-themes (which provides two themes: edx.org and mitpe). The edx.org theme uses the default sass wholesale (which comes from this repo), so we can adjust it here as a whole and we won't be out of sync. And the mitpe theme imports edx-pattern-library itself and overrides all the important stuff. Images below for proof.

I've tested in chromium and firefox. I've tried to test to confirm that Student Records is the same, but my local environment is not set up for that. I need to bug someone in the office to test that. But Student Records has minimal use of edx-pattern-library.

Test sandbox: https://credentials-pattern-removal.sandbox.edx.org/credentials/example/

Here are some comparison images (old first, then new). I don't have images for it, but trust that I've manually checked the 2-org and 3-org cases, plus the multiple-signatory cases. There are some slight whitespace differences, due to bootstrap having some different standard sizes. But nothing big.

OpenEdx Web
cred-old-openedx
cred-new-openedx

edX Web
cred-old-edxorg
cred-new-edxorg

MIT Web
cred-old-mitpe
cred-new-mitpe

OpenEdx PDF
cred-old-openedx-pdf
cred-new-openedx-pdf

edX PDF
cred-old-edxorg-pdf
cred-new-edxorg-pdf

MIT PDF:
cred-old-mitpe-pdf
cred-new-mitpe-pdf

https://openedx.atlassian.net/browse/LEARNER-6318

Credentials Pull Request
---

  - UI Changes 
    - [ ] Check mobile view


credentials/static/sass/_print-rendering.scss
```diff
@@ -1,125 +0,0 @@
-// ------------------------------
```
mikix: This file is deleted because it wasn't being used! Humph.

credentials/apps/credentials_theme_openedx/static/sass/_print.scss
```diff
@@ -1,138 +0,0 @@
-// ------------------------------
```
mikix: This file is deleted because it wasn't being used! Humph.

credentials/apps/credentials_theme_openedx/static/sass/_build.scss
```diff
@@ -16,14 +16,11 @@
 // #LIB
 // ------------------------------
 @import 'lib';
-// Search for #bootstrapupgrade on upgrade to edX-bootstrap.
-@import '~edx-pattern-library/pattern-library/sass/edx-pattern-library';
```
bderusha: 🔥

mikix:
@bderusha mostly tested in Firefox, with spot checks in Chromium and Edge.

PR #636, opened 2019-02-22
Add username replacement API

tuchfarber:
New API replaces all copies of username in this service with a new
username. Requires user be in a username_replacement_admin group. Should
only be ran as a larger job to update usernames across all services,
otherwise the system will be left in a broken state for those users.

Credentials Pull Request
---

  - UI Changes 
    - [ ] Check mobile view

### DIFF
--- a/credentials/apps/api/v2/permissions.py
+++ b/credentials/apps/api/v2/permissions.py
@@ -1,6 +1,7 @@
 """
 Custom permissions classes for use with DRF.
 """
+from django.conf import settings
 from rest_framework import permissions
 
 
@@ -40,3 +41,11 @@ class UserCredentialPermissions(permissions.DjangoModelPermissions):
             return default or (request.user.username.lower() == filtered_username)
 
         return default
+
+
+class CanReplaceUsername(permissions.BasePermission):
+    """
+    Grants access to the Username Replacement API for the service user.
+    """
+    def has_permission(self, request, view):
+        return request.user.username == getattr(settings, "USERNAME_REPLACEMENT_WORKER")
--- a/credentials/apps/api/v2/tests/test_views.py
+++ b/credentials/apps/api/v2/tests/test_views.py
@@ -2,12 +2,14 @@ import json
 from decimal import Decimal
 
 import ddt
+import mock
 from django.contrib.auth.models import Permission
 from django.test import TestCase
 from django.urls import reverse
 from rest_framework.renderers import JSONRenderer
 from rest_framework.test import APIRequestFactory, APITestCase
 
+from credentials.apps.api.tests.mixins import JwtMixin
 from credentials.apps.api.v2.serializers import (UserCredentialAttributeSerializer, UserCredentialSerializer,
                                                  UserGradeSerializer)
 from credentials.apps.api.v2.views import CredentialRateThrottle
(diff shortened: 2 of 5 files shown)


credentials/apps/api/v2/views.py
```diff
@@ -145,3 +148,135 @@ def partial_update(self, request, *args, **kwargs):  # pylint: disable=useless-s
...
+                return False
+        return True
+
+    def _replace_username_for_all_models(self, current_username, new_username, replacement_locations):
+        """
+        Replaces current_username with new_username for all (model, column) pairs in replacement locations.
+        Returns if it was successful or not. Will return successful even if no matching
+
+        TODO: Determine if logs of username are a PII issue.
+        """
```
BessieSteinberg: Remove TODO

credentials/apps/api/v2/tests/test_views.py
```diff
@@ -517,3 +519,82 @@ def test_throttle_configuration(self, scope):
...
+        existing_users = [{user.username: user.username + '_new'} for user in random_users]
+        non_existing_users = [{username: username + '_new'} for username in fake_usernames]
+        data = {
+            "username_mappings": existing_users + non_existing_users
+        }
+        expected_response = {
+            'failed_replacements': [],
+            'successful_replacements': existing_users + non_existing_users
+        }
+        response = self.call_api(self.service_user, data)
```
BessieSteinberg: Same comment as on discovery: Is the expected behavior that there will be successful replacements for non-existing users? Are new user records created for users if the old username doesn't exist? It's fine if that is the case just add that to the docstring for the api and maybe add a comment about that in this test

tuchfarber: I'll definitely add some more comments around this in the services as this was something I realized later and changed. If a user never views a credential, they aren't going to exist on the credentials service at all. So we should treat the lack of a user a success as we don't need to do any work.

credentials/apps/api/v2/views.py
tuchfarber: Removed

credentials/apps/api/v2/views.py
```diff
@@ -145,3 +148,135 @@ def partial_update(self, request, *args, **kwargs):  # pylint: disable=useless-s
...
         """ Update a grade. """
         return super(GradeViewSet, self).update(request, *args, **kwargs)
+
+
+class UsernameReplacementView(APIView):
+    """
+    WARNING: This API is only meant to be used as part of a larger job that
+    updates usernames across all services. DO NOT run this alone or users will
+    not match across the system and things will be broken.
+
```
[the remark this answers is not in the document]
tuchfarber: Added note.

credentials/apps/api/v2/views.py
```diff
@@ -145,3 +148,136 @@ def partial_update(self, request, *args, **kwargs):  # pylint: disable=useless-s
...
+        {
+            "successful_replacements": [
+                {"old_username_1": "new_username_1"}
+            ],
+            "failed_replacements": [
+                {"old_username_2": "new_username_2"}
+            ]
+        }
+
+        TODO: Determine if we need an audit trail outside of logging and API response.
```
georgebabey: Is this TODO done? Do we need anything further?

credentials/apps/api/v2/permissions.py
```diff
@@ -40,3 +41,12 @@ def has_permission(self, request, view):
             return default or (request.user.username.lower() == filtered_username)
 
         return default
+
+
+class CanReplaceUsername(permissions.BasePermission):
+    """
+    Grants access to the Username Replacement API for anyone in the group,
```
georgebabey: remove mention of group

PR #637, opened 2019-02-22
LEARNER-7043: Encode filename to download record

zainab-amir:
[LEARNER-7043](https://openedx.atlassian.net/browse/LEARNER-7043)

Learners enrolled in some programs i.e **Gestión Pública para el Desarrollo** are getting **_internal server error_**. The issue is in the filename. If the filename contains special characters that do not lie in ordinal range(128) of ascii it gives internal server error. The filename is created by following the pattern username_program_name_grades.csv. This filename becomes a part of Content Disposition and the error occurs in [gunicorn package](https://github.com/benoitc/gunicorn/blob/master/gunicorn/http/wsgi.py#L318) as it uses ascii
#### Description
For a fix I changed the code to encode the file in 'utf-8' before sending it to response['Content-Disposition']
#### Tests
- I have setup gunicorn on local system and tested the changes
- Verified on chrome and firefox

### DIFF
--- a/credentials/apps/records/views.py
+++ b/credentials/apps/records/views.py
@@ -551,7 +551,7 @@ class ProgramRecordCsvView(RecordsEnabledMixin, View):
         writer.writeheader()
         writer.writerows(record['grades'])
         string_io.seek(0)
-        filename = '{username}_{program_name}_grades.csv'.format(
+        filename = '{username}_{program_name}_grades'.format(
             username=record['learner']['username'],
             program_name=record['program']['name']
         )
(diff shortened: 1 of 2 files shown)


credentials/apps/records/views.py
```diff
@@ -564,7 +564,9 @@ def get(self, request, *args, **kwargs):
             properties=properties,
             segment_client=segment_client,
         )
-        response['Content-Disposition'] = 'attachment; filename={filename}'.format(
-            filename=filename.replace(' ', '_').lower()
+        filename = filename.replace(' ', '_').lower()
+        filename = filename.encode('utf-8')
```
waheedahmed: I believe we can achieve that in single line `filename = filename.replace(' ', '_').lower().encode('utf-8')`

zainab-amir: okay

PR #639, opened 2019-02-25
Remove PUBLIC_URL default from base - should only be set in devstack.

doctoryes:
Credentials Pull Request
---
I've reproduced https://openedx.atlassian.net/browse/ARCH-414 on my devstack and this change fixes the problem. If *any* `SOCIAL_AUTH_EDX_OAUTH2_PUBLIC_URL_ROOT` is set, that value will be used in authN instead of `SOCIAL_AUTH_EDX_OAUTH2_URL_ROOT`:

https://github.com/edx/auth-backends/blob/ef976034ac0fe6c4b323d7a7773027dee16873a7/auth_backends/backends.py#L48

But this setting is only relevant on devstack and is set here properly:

https://github.com/edx/credentials/blob/master/credentials/settings/devstack.py#L54

---

  - UI Changes 
    - [ ] Check mobile view


nasthagiri:
Thanks for getting to the bottom of this!

PR #643, opened 2019-04-15
PROD-189 send emails for program if records_enabled is true

zainab-amir:
#### Description 
If for some site the Site Configurations for credentials have the option "Enable Learner Records" unchecked, the programs are not being saved into the database when management command copy_catalog is called. Hence send_updated_emails_for_program() should be called only if the site
configurations has the value for records_enabled set to true.

#### Notes
* Added another check to verify that config has records_enabled set to true before sending an updated email to a pathway org

[PROD-189](https://openedx.atlassian.net/browse/PROD-189)

### DIFF
--- a/credentials/apps/api/tests/test_accreditors.py
+++ b/credentials/apps/api/tests/test_accreditors.py
@@ -8,6 +8,7 @@ from testfixtures import LogCapture
 
 from credentials.apps.api.accreditors import Accreditor
 from credentials.apps.api.exceptions import UnsupportedCredentialTypeError
+from credentials.apps.core.tests.mixins import SiteMixin
 from credentials.apps.credentials.issuers import CourseCertificateIssuer, ProgramCertificateIssuer
 from credentials.apps.credentials.models import CourseCertificate, ProgramCertificate
 from credentials.apps.credentials.tests.factories import CourseCertificateFactory, ProgramCertificateFactory
@@ -15,7 +16,7 @@ from credentials.apps.credentials.tests.factories import CourseCertificateFactor
 LOGGER_NAME = 'credentials.apps.api.accreditors'
 
 
-class AccreditorTests(TestCase):
+class AccreditorTests(SiteMixin, TestCase):
     attributes = [{'name': 'whitelist_reason', 'value': 'Reason for whitelisting.'}]
     course_credential = CourseCertificate
     program_credential = ProgramCertificate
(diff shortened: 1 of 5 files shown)


credentials/apps/api/tests/test_accreditors.py
```diff
@@ -23,6 +24,8 @@ class AccreditorTests(TestCase):
     def setUp(self):
         super().setUp()
         self.program_cert = ProgramCertificateFactory()
+        self.site = SiteFactory()
```
waheedahmed: Use `SiteMixin` https://github.com/edx/credentials/blob/4a46362c25e692a8f2427296596cc10239942bf9/credentials/apps/core/tests/mixins.py#L13

credentials/apps/credentials/tests/test_issuer.py
```diff
@@ -126,6 +127,12 @@ def test_existing_attributes_with_empty_attributes_list(self):
             self.user_cred, self.certificate, self.user_cred.username, 'awarded', self.attributes
         )
 
+    def test_records_enabled_is_unchecked(self):
```
waheedahmed: Remove this.

credentials/apps/credentials/tests/test_issuer.py
```diff
@@ -22,10 +23,11 @@ class CertificateIssuerBase(object):
 
     def setUp(self):
         super(CertificateIssuerBase, self).setUp()
+        self.site = SiteFactory()
```
waheedahmed: Can we use the `SiteMixin` here as well?

credentials/apps/credentials/tests/test_issuer.py
```diff
@@ -135,13 +142,28 @@ class ProgramCertificateIssuerTests(CertificateIssuerBase, TestCase):
...
+        self.site_config = SiteConfigurationFactory(site=self.site)
         self.program = ProgramFactory(site=self.site)
         self.certificate = self.cert_factory.create(program_uuid=self.program.uuid, site=self.site)
         self.username = 'tester'
         self.user = UserFactory(username=self.username)
-        self.user_cred = self.issuer.issue_credential(self.certificate, self.username)
+        self.user_cred = self.issuer.issue_credential(self.site, self.certificate, self.username)
         self.attributes = [{"name": "whitelist_reason", "value": "Reason for whitelisting."}]
 
+    def test_records_enabled_is_unchecked(self):
```
waheedahmed: Is there any test available which is asserting for checked state? If not I think we should add that as well.

credentials/apps/credentials/tests/test_issuer.py
zainab-amir: I tried but it gives ```AttributeError: 'ProgramCertificateIssuerTests' object has no attribute 'site'``` error

waheedahmed: Did you tried using it in base class `CertificateIssuerBase` or in `ProgramCertificateIssuerTests`?

zainab-amir: Tried it with both separately, CertificateIssuerBase and ProgramCertificateIssuerTests

PR #647, opened 2019-05-03
Disable share and send record button on onClick event

zainab-amir:
Due to a known issue in get_or_create() method that states that if uniqueness is not enforced at the database level for the kwargs used in a get_or_create call, this method is prone to a race-condition which can result in multiple rows with the same parameters being inserted simultaneously.

Following changes are added:
* a unique_together constraint on ProgramCertRecord
* disable send and share button after onClick event to avoid concurrent requests and enable them once the modal is closed

[PROD-197](https://openedx.atlassian.net/browse/PROD-197)
[stackoverflow for react disable button](https://stackoverflow.com/a/47824795)


schenedx:
@mikix Do you mind do a quick review on this PR? I don't really have records knowledge as well as you do

credentials/apps/records/models.py
```diff
@@ -48,6 +48,7 @@ def __str__(self):
 
     class Meta(object):
         verbose_name = "Shared program record"
+        unique_together = ('program', 'user')
```
mikix: This was intentionally not done at the time we were building this feature.

We wanted to leave open the possibility of multiple shares. Maybe with different parameters to come later (like "share a version with less details"). Or maybe having multiple URL shares that you could control / delete separately (have one URL to give MIT, another to give your boss or something).

This was speculative. But we figured we'd keep the logic about "one public record per user per program" outside of the DB at the time. That said, we haven't built those features and probably don't have them on the horizon?

So sure, add a unique_together if you feel like that would help with data integrity. Can always remove it later.

PR #665, opened 2019-08-07
WIP:Changed logout from get to post

adeelehsan:
whenever user logout from IDA's he
will be redirected to confirm logout
url on lms witout logging out from IDA's.
Once user confirms logout then we'll logout
from all IDA's and lms.

learner-5252


staubina:
@adeelehsan, Is this PR still in development? Should we close it?

PR #673, opened 2019-09-13
Retired program credential is showing on dashboard

uzairr:
### Description
Currently, if a learner is enrolled in a course-run which is a part of a retired program then that retired program start appearing on program dashboard.Once the program is retired,it must not be shown on the program dashboard in any case.

### DIFF
--- a/credentials/apps/records/tests/test_views.py
+++ b/credentials/apps/records/tests/test_views.py
@@ -140,7 +140,6 @@ class RecordsViewTests(SiteMixin, TestCase):
 
     @ddt.data(
         (Program.ACTIVE, True),
-        (Program.RETIRED, True),
         (Program.DELETED, False),
         (Program.UNPUBLISHED, False),
     )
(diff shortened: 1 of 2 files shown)


credentials/apps/records/tests/test_views.py
```diff
@@ -140,7 +140,6 @@ def test_help_url(self):
 
     @ddt.data(
         (Program.ACTIVE, True),
-        (Program.RETIRED, True),
```
mikix: Probably makes more sense to switch True to False, eh?

credentials/apps/records/views.py
```diff
@@ -280,7 +280,7 @@ def get_context_data(self, **kwargs):
 
 class RecordsView(RecordsListBaseView):
     def _get_programs(self):
-        return self._programs_context(include_empty_programs=False, include_retired_programs=True)
+        return self._programs_context(include_empty_programs=False)
```
mikix: So this was a very specific product requirement at the time we built the learner records feature. Will a learner lose access to their record of grades and completion once a program is retired? That seems bad.

Has Product been looped into this ticket? I don't think they would want this change.

uzairr: Thanks @mikix for your review.
Would you please add your concern on this ticket: https://openedx.atlassian.net/browse/PROD-596
There is also a discussion about this change on its linked CR ticket.

CC: @fysheets @waheedahmed

PR #675, opened 2019-09-17
Changed from_address while sending email to partner

adeelehsan:
previously we were using learner-records@edx.org in from address.
Now we will be using learner email address so that partner can
contact with learner.

Prod-624

### DIFF
--- a/credentials/apps/records/messages.py
+++ b/credentials/apps/records/messages.py
@@ -2,13 +2,13 @@ from edx_ace import MessageType
 
 
 class ProgramCreditRequest(MessageType):
-    def __init__(self, site, *args, **kwargs):
+    def __init__(self, site, user_email, *args, **kwargs):
         super(ProgramCreditRequest, self).__init__(*args, **kwargs)
 
-        if site.siteconfiguration.partner_from_address:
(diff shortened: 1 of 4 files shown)


credentials/apps/records/messages.py
```diff
@@ -2,10 +2,12 @@
...
 
 class ProgramCreditRequest(MessageType):
-    def __init__(self, site, *args, **kwargs):
+    def __init__(self, site, user_email=None,  *args, **kwargs):
         super(ProgramCreditRequest, self).__init__(*args, **kwargs)
 
-        if site.siteconfiguration.partner_from_address:
+        if user_email:
+            from_address = user_email
+        elif site.siteconfiguration.partner_from_address:
```
waheedahmed: Why we are still checking for this if we are always going to send it from learner email?

credentials/apps/records/messages.py
```diff
@@ -2,10 +2,12 @@
 
 
 class ProgramCreditRequest(MessageType):
-    def __init__(self, site, *args, **kwargs):
+    def __init__(self, site, user_email=None,  *args, **kwargs):
```
waheedahmed: `user_email` should be a required argument.

PR #683, opened 2019-10-03
Update the key of the course_run object from uuid to key

schenedx:
Update the key to reference the course_run copy of the data in discovery service.

Credentials Pull Request
---

### DIFF
--- a/credentials/apps/catalog/utils.py
+++ b/credentials/apps/catalog/utils.py
@@ -25,9 +25,9 @@ def parse_organization(site, data):
 def parse_course_run(course, data):
     course_run, _ = CourseRun.objects.update_or_create(
         course=course,
-        uuid=data['uuid'],
+        key=data['key'],
         defaults={
-            'key': data['key'],
+            'uuid': data['uuid'],
             'title_override': data['title'] if data['title'] != course.title else None,
             'start': data['start'],
             'end': data['end'],


schenedx:
@kdmccormick, @MichaelRoytman and @mikix please review

mikix:
Hmm, yeah. The unique constraint for CourseRun in course-discovery is simply `('key', 'draft')` -- which for your non-draft purposes is simply `key`.

So the unique together for credentials of `('course', 'uuid')` (that I probably wrote?) is being optimisitic that discovery will enforce something it doesn't actually promise to.

As for what credentials *should* use for a unique together...  It's complicated by the fact that credentials can technically point to multiple course-discovery services via microsites. So it can't simply say `key` itself. Which is why the CourseRun unique_together is the oddly phrased `('course', 'uuid')` -- and then courses are themselves unique for a site.

Anyway, yeah I agree with Kyle that you should only ever pass unique_together elements in as kwargs to update_or_create. So maybe update the unique_together for CourseRun to use `('course', 'key')`?

**Alternative**
There are only 20 UUIDs that are mucked up in course-discovery, all from ~2017 (except for one from 2019 but only because it got accidentally deleted and re-installed in db -- I bet it's originally from 2017). Maybe investigate and then ask devops to manually fix whatever we decide is wrong with them.

Looks like someone was making re-runs (or renaming runs?) with a bad process:

```
select count(id), group_concat(`key`) from course_metadata_courserun where draft = false group by `uuid` having count(id) > 1;
+-----------+-------------------------------------------------------------------------------------+
| count(id) | group_concat(`key`)                                                                 |
+-----------+-------------------------------------------------------------------------------------+
|         2 | course-v1:ASUx+ECN211x+2177A,course-v1:ASUx+ECN211x+2171B                           |
|         2 | course-v1:DelftX+EX101SPx+1T2018,course-v1:DelftX+EX101x+1T2018                     |
|         2 | course-v1:GTx+MGT6203x+1T2018,course-v1:GTx+MGT6203x+2T2017                         |
|         2 | course-v1:UC3Mx+IT.1.2x+1T2018,course-v1:UC3Mx+IT.1.2-ESx+1T2018                    |
|         2 | course-v1:IMFx+FPP.1x_AR+2T2017,course-v1:IMFx+FPP.1x+2T2017                        |
|         2 | course-v1:Microsoft+DAT215.5x+4T2017,course-v1:Microsoft+DAT215.5x+5T2017           |
|         2 | course-v1:UAMx+GenEvolx+1T2017,course-v1:UAMx+GenEvolx+2T2017                       |
|         2 | course-v1:DavidsonX+MedGen101x+3T2016,course-v1:DavidsonNext+CME_MedGen101x+3T2016  |
|         2 | course-v1:UC3Mx+IT.1.1-ESx+1T2018,course-v1:UC3Mx+IT.1.1x+1T2018                    |
|         2 | course-v1:JuilliardX+JCx001+2T2017,course-v1:JuilliardX+JX001x+3T2017               |
|         2 | course-v1:UAMx+QuijoteENx+2T2017,course-v1:UAMx+QuijoteENx+1T2017                   |
|         2 | course-v1:MITx+CITE101x+2T2017,course-v1:MITx+CITE101x+1T2017                       |
|         2 | course-v1:MITProfessionalX+ENX_Fake+3T2016,course-v1:MITx+ENx+2016_3T               |
|         2 | course-v1:UTAustinX+UT.WTCS.15.01x+2T2017,course-v1:UTAustinX+UT.WTCS.14.01x+2T2017 |
|         2 | course-v1:UC3Mx+IT.1.3-ESx+2T2018,course-v1:UC3Mx+IT.1.3x+2T2018                    |
|         2 | course-v1:GTx+CSE88803x+2T2017,course-v1:GTx+CSE6250+2T2017                         |
|         2 | course-v1:IcelandX+UOI001X+1T2018a,course-v1:UIcelandX+UOI001X+1T2018a              |
|         2 | course-v1:TUMx+QPLS3x+2T2018,course-v1:TUMx+QPLS3x+2T_2018                          |
|         2 | course-v1:TUMx+QPLS2x+2T_2018,course-v1:TUMx+QPLS2x+2T2018                          |
|         2 | course-v1:DelftX+EX102SPx+1T2018,course-v1:DelftX+EX102x+1T2018                     |
+-----------+-------------------------------------------------------------------------------------+
20 rows in set (0.51 sec)
```

mikix:
(Or you could land a patch that makes uuid fields editable, fix them, then undo that patch -- to avoid bothering devops)

kdmccormick:
@mikix I like the idea of cleaning up Discovery to solve the issue -- as the central course metadata repo, IDAs should be able to trust Discovery data.

I'm trying to think of a unique_together constraint that we could put on all the draft-enabled models in Discovery to keep the DB clean going forward. Records with a draft could be protected with `('uuid', 'draft_version')` because of the one-to-oneness of of `draft_version`, but I'm not sure what constraint would work for records without a draft.

mikix:
> Edit: After cleaning up the DB, would a unique_together of ('uuid', 'draft') on CourseRun work?

Yes. It would also follow what Course does (it has separate unique_togethers for uuid and key).

The one theoretical wrinkle is Site support. So in theory, the same exact course run could be offered in two Sites? But that is in practice not allowed today because runs have unique keys even across Sites according to the unique_together. So... Might as well also guarantee the uuid across Sites.

That mismatch is annoying to me. We have `partner` in the unique_together for Courses, thus completely partitioning the courses between partners. We should either do that for both courses and runs or neither.

But we wouldn't be making the situation worse by adding the uuid constraint today.

kdmccormick:
Based on the RR, it looks like we could add `(uuid, draft)` as a constraint on `Course` as well. Whether the code supports that constraint, I'm not sure.
```
mysql> select count(id), group_concat(uuid) from course_metadata_course where draft = false group by uuid having count(id) > 1;
Empty set (0.02 sec)
```

mikix:
> Based on the RR, it looks like we could add (uuid, draft) as a constraint on Course as well

Oh sorry, it already has that. That's what I was trying to say with `It would also follow what Course does (it has separate unique_togethers for uuid and key).`

kdmccormick:
@mikix I mean that we could possibly remove the `partner` partitioning from `Course`: https://github.com/edx/course-discovery/blob/master/course_discovery/apps/course_metadata/models.py#L697-L700

mikix:
> So ('partner', 'uuid', 'draft') would become ('uuid', 'draft').

Ah ah got you, sorry. Yeah... I kinda like the idea of partitions, but yeah. I know at edx.org, we are going away from Sites in general. So less need for a partition.

kdmccormick:
I don't have a strong feeling about partitions... but I feel weird about duplicating UUID across active records. Maybe the data model is off? If the same Course can be shared between several partners, shouldn't Course have just have a many-to-many relationship with Partner, but remain unique on UUID?

mikix:
Yeah I think you are right - I don't have historical context for the current setup. I just know that you *can* partition, so assumed it was an intentional feature. But we are clearly not using it (and can't really, case run keys would conflict).

So let's simplify and remove the partitioning...?

schenedx:
@kdmccormick and @mikix Thank you for all of those comments. 
I created a CR ticket for this problem as well. See https://openedx.atlassian.net/servicedesk/customer/portal/9/CR-1311
Reading what you guys said, is it correct that I should just close this PR?

kdmccormick:
@schenedx That is a good ticket, but I think it's also important to add the uniqueness constraints to prevent this from happening again in the future, and tightening `Course`'s (and maybe other course_metadata models') uniqueness constraints to prevent a similar problem there. I'm not sure whose backlog the work belongs in, though.

schenedx:
@kdmccormick Go ahead and create a jira ticket against the Educator project. If we found home about it after some conversation, we can always move. It's more important at this point to just record down the needed system change. We shall find the owner of that change later.

kdmccormick:
https://openedx.atlassian.net/browse/EDUCATOR-4717

PR #704, opened 2019-11-12
Rename start and end fields (4.2/4)

pwnage101:
This is the 4.2th stage of renaming the start and end fields of
CourseRun to start_date and end_date.

This release ONLY removes the old columns via migration.  Note that this
does not include removing the django model fields corresponding to the
old columns.

DE-1708

See also: https://github.com/edx/credentials/pull/703


credentials/apps/catalog/models.py
```diff
@@ -56,10 +56,6 @@ class CourseRun(TimeStampedModel):
     start_date = models.DateTimeField(null=True, blank=True, db_index=True)
     end_date = models.DateTimeField(null=True, blank=True, db_index=True)
 
-    # TODO: delete these two fields during the last stage of rolling out the field renames.
```
schenedx: My experience is, this is wrong.
You should first check in this change on this file, but do not merge in the migration file.
Then you release this django code change out.
Then the last PR is to merge in the migration file and then deploy that.

pwnage101: Yes, I have actually ordered the two PRs (4.1 and 4.2) in that way, but in this PR I included the commit from the previous PR so that I can simply merge it without rebasing.  In retrospect, doing so was unnecessary since the two changes are in different files.

schenedx: Is it the case that in production right now, you already have the state of the code is no longer there, but the migration file is not yet run? You basically reverted too aggressively?

PR #712, opened 2019-12-09
Update path lib so it is renamed

schenedx:
Credentials Pull Request
---
Upgrade `path.py` to the major version where it is renamed to `path`

@edx/masters-devs FYI

### DIFF
--- a/requirements/all.txt
+++ b/requirements/all.txt
@@ -83,7 +83,8 @@ olefile==0.46
 openapi-codec==1.3.2
 packaging==19.2
 paramiko==2.7.0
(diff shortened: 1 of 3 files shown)


kdmccormick:
I was able to reproduce the failure locally, but I think it has to do with some wackyness with Credential's Travis set-up, and wouldn't be an issue in production. It does seem worth sorting out, though.

kdmccormick:
@schenedx can this be closed?

schenedx:
Sure, I don't know the value of keeping this open. We will come back to this later.

PR #733, opened 2020-02-10
Upgrade edx-lint - BOM-1258

iamsobanjaved:
JIRA: [BOM-1258](https://openedx.atlassian.net/browse/BOM-1258)

### DIFF
--- a/credentials/apps/api/v2/permissions.py
+++ b/credentials/apps/api/v2/permissions.py
@@ -48,4 +48,4 @@ class CanReplaceUsername(permissions.BasePermission):
     Grants access to the Username Replacement API for the service user.
     """
     def has_permission(self, request, view):
-        return request.user.username == getattr(settings, "USERNAME_REPLACEMENT_WORKER")
+        return request.user.username == settings.USERNAME_REPLACEMENT_WORKER
--- a/credentials/apps/api/v2/views.py
+++ b/credentials/apps/api/v2/views.py
@@ -19,6 +19,8 @@ from credentials.apps.records.models import UserGrade
 
 log = logging.getLogger(__name__)
 
+# pylint: disable= no-member
+
 
 def credentials_throttle_handler(exc, context):
     """ Exception handler for logging messages when an endpoint is throttled. """
@@ -48,7 +50,7 @@ class CredentialRateThrottle(ScopedRateThrottle):
     def allow_request(self, request, view):
         user = request.user
         if user.is_authenticated and (user.is_staff or user.is_superuser):
-            setattr(view, 'throttle_scope', 'staff_override')
+            view.throttle_scope = 'staff_override'
 
         return super(CredentialRateThrottle, self).allow_request(request, view)
 
--- a/credentials/apps/credentials/tests/test_issuer.py
+++ b/credentials/apps/credentials/tests/test_issuer.py
@@ -15,7 +15,7 @@ LOGGER_NAME = 'credentials.apps.credentials.issuers'
 
 
 # pylint: disable=no-member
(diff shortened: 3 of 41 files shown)


credentials/apps/core/s3utils.py
```diff
@@ -4,14 +4,14 @@
 from functools import partial
 
 from django.conf import settings
-from storages.backends.s3boto import S3BotoStorage
+from storages.backends.s3boto3 import S3Boto3Storage
```
Ayub-Khan: Why we needed to update those. Any chance of bugs ?

iamsobanjaved: @Ayub-Khan due to the make upgrade, the version of django-storage is updated and the already deprecated `s3boto` is now removed in the favour of `s3boto3`

jmbowman: There's a set of things we need to check when making this switch: https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html#migrating-boto-to-boto3 .  I'm inclined to pin `django-storages<1.9` for now and ticket doing that upgrade separately.

credentials/apps/credentials/tests/test_issuer.py
```diff
@@ -135,6 +135,7 @@ class ProgramCertificateIssuerTests(CertificateIssuerBase, TestCase):
     cert_type = ProgramCertificate
 
     def setUp(self):
+        # pylint: disable=super-method-not-called
```
jmbowman: I'd be inclined to call `super().setUp()` instead, it should be harmless.

credentials/apps/records/tests/test_views.py
```diff
@@ -343,6 +343,7 @@ def test_normal_access(self):
         actual_child_templates = response_context_data['child_templates']
         self.assert_matching_template_origin(actual_child_templates['footer'], '_footer.html')
         self.assert_matching_template_origin(actual_child_templates['header'], '_header.html')
+        # pylint: disable=wrong-assert-type
         self.assertFalse('masquerade' in actual_child_templates)  # no masquerading on this view
```
jmbowman: Instead of disabling the warning, could we change this to `assert 'masquerade' not in actual_child_templates`?  That's simpler anyway.

credentials/apps/api/v2/permissions.py
```diff
@@ -48,4 +48,5 @@ class CanReplaceUsername(permissions.BasePermission):
     Grants access to the Username Replacement API for the service user.
     """
     def has_permission(self, request, view):
+        # pylint: disable=literal-used-as-attribute
         return request.user.username == getattr(settings, "USERNAME_REPLACEMENT_WORKER")
```
jmbowman: Should probably just change this to `return request.user.username == settings.USERNAME_REPLACEMENT_WORKER` instead of silencing the warning.  It's simpler, and I'm having trouble thinking of a case where this would change the behavior (the setting should always exist since it's in the base settings module).

credentials/apps/api/v2/views.py
```diff
@@ -19,6 +19,8 @@
 
 log = logging.getLogger(__name__)
 
+# pylint: disable= no-member, literal-used-as-attribute
```
jmbowman: I only see one `setattr` usage that triggers the second warning, and it should probably be fixed to use direct attribute assignment instead (`view.throttle_scope = 'staff_override'`).

iamsobanjaved: Pylint gives no-member error on drf's ModelViewSets methods such as list, retrieve, etc. `setattr` is resolved

PR #734, opened 2020-02-11
BOM-1260

Ayub-Khan:
-Upgrade social-auth-app-django to support django22

### DIFF
--- a/requirements/all.txt
+++ b/requirements/all.txt
@@ -4,7 +4,7 @@
 #
 #    make upgrade
 #
--e git+https://github.com/jsocol/django-ratelimit.git@72edbe8949fbf6699848e5847645a1998f121d46#egg=ratelimit
+-e git+https://github.com/jsocol/django-ratelimit.git@72edbe8949fbf6699848e5847645a1998f121d46#egg=django-ratelimit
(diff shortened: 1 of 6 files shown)


requirements/all.txt
```diff
@@ -4,7 +4,7 @@
 #
 #    make upgrade
 #
--e git+https://github.com/jsocol/django-ratelimit.git@72edbe8949fbf6699848e5847645a1998f121d46#egg=ratelimit
+-e git+https://github.com/jsocol/django-ratelimit.git@72edbe8949fbf6699848e5847645a1998f121d46#egg=django-ratelimit
```
jmbowman: Why was this change needed?  We still have it the old way in edx-platform.

Ayub-Khan: This fixes the warning 
```
WARNING:pip._internal.req.req_install:Generating metadata for package ratelimit produced metadata for project name django-ratelimit. Fix your #egg=ratelimit fragments.
```

jmbowman: Interesting, I'll go re-read the docs on that `#egg` setting and see if we need to fix it elsewhere also.

PR #742, opened 2020-02-19
BOM-1281 : Implement Tox in credentials

mzulqarnain1:
Related JIRA Link : [https://openedx.atlassian.net/browse/BOM-1281](https://openedx.atlassian.net/browse/BOM-1281)


tox.ini
```diff
@@ -0,0 +1,28 @@
...
+whitelist_externals:
+    npm
+    make
+commands =
+    npm install --unsafe-perm
+    npm install --production --no-save
+    make static
+    coverage run -m pytest --ds credentials.settings.test --durations=25
+    coverage report
+    $NODE_BIN/gulp test
```
mzulqarnain1: @feanil We are trying to move tests to tox, and this command we moved from makefile to tox.ini gives an error `ERROR: InvocationError for command could not find executable '$NODE_BIN/gulp'` We couldn't find any solution for this problem, @UsamaSadiq is facing same issue in course-discovery. Can you help us with a solution?

jmbowman: @mzulqarnain1 @UsamaSadiq We discussed some options for fixing the InvocationError in an earlier meeting, have those been tried yet?  I noticed that both this and the course-discovery ticket are marked as blocked, and I want to make sure you have what you need to make progress on them.

jmbowman: New suggestion on this: can we just break out the gulp and react tests into a separate target that isn't called from within tox?  Those tests are executed via JavaScript, so there's no reason to run them from inside a tox virtualenv.  Let's just separate the Python and JS testing targets for Travis, with a target for local development convenience that runs both.

package-lock.json
```diff
@@ -9,8 +9,8 @@
       "integrity": "sha512-3T6yYiLK9WdIn0iwjLoWfrsDqgfd0byiVB3aXK7w4oaltdVzA/fB/qU3iKrJ8cy4T6ZgDA2uFP726fU0StmxUw==",
       "requires": {
         "bootstrap": "4.0.0",
-        "jquery": "3.3.1",
-        "popper.js": "1.14.6"
+        "jquery": "^3.2.1",
+        "popper.js": "^1.12.6"
```
jmbowman: It looks like a lot of npm packages were downgraded here; was this necessary, or was checking in changes to this file accidental?

tox.ini
```diff
@@ -0,0 +1,22 @@
+[tox]
+envlist = {py35}-django{111,20,21,22}
+skipdist = true
```
jmbowman: This should be "skipsdist".  Fixing the spelling should eliminate the need for a `setup.py` file.

tox.ini
```diff
@@ -0,0 +1,22 @@
...
+    django20: Django>=2.0,<2.1
+    django21: Django>=2.1,<2.2
+    django22: Django>=2.2,<2.3
+    -r {toxinidir}/requirements/test.txt
+whitelist_externals:
+    make
+commands =
+    make js-requirements
+    make static
+    coverage run -m pytest --ds credentials.settings.test --durations=25
```
jmbowman: You shouldn't need to specify the settings file here since it's in the pytest configuration above.

package-lock.json
mzulqarnain1: I'll drop this file's changes. Not really needed.

tox.ini
mzulqarnain1: nice catch. Thanks.

tox.ini
mzulqarnain1: will remove.

PR #754, opened 2020-03-10
BOM-1045

Ayub-Khan:
- edx-drf-extensions version bump
- settings update
- fixed tests


credentials/apps/api/tests/test_authentication.py
```diff
@@ -61,7 +62,7 @@ def test_required_claims(self, claim):
         del jwt_payload[claim]
         jwt_value = self.generate_token(jwt_payload)
         request = APIRequestFactory().get('dummy', HTTP_AUTHORIZATION='JWT {}'.format(jwt_value))
-        with self.assertRaises(AuthenticationFailed):
+        with self.assertRaises(MissingRequiredClaimError):
```
awais786: Please add some information why you changed this from AuthenticationFailed ?

Ayub-Khan: This was updated not it raises a more specific error.

requirements/all.txt
```diff
@@ -39,19 +39,19 @@ django-statici18n==1.9.0  # via -r requirements/dev.txt, -r requirements/product
...
 dockerpty==0.4.1          # via -r requirements/dev.txt, docker-compose
 docopt==0.6.2             # via -r requirements/dev.txt, docker-compose
+drf-jwt==1.14.0           # via -r requirements/dev.txt, -r requirements/production.txt, edx-drf-extensions
 edx-ace==0.1.13           # via -r requirements/dev.txt, -r requirements/production.txt
 edx-auth-backends==3.0.2  # via -r requirements/dev.txt, -r requirements/production.txt
 edx-django-release-util==0.3.6  # via -r requirements/dev.txt, -r requirements/production.txt
 edx-django-sites-extensions==2.4.3  # via -r requirements/dev.txt, -r requirements/production.txt
 edx-django-utils==3.0     # via -r requirements/dev.txt, -r requirements/production.txt, edx-drf-extensions, edx-rest-api-client
-edx-drf-extensions==3.0.0  # via -r requirements/dev.txt, -r requirements/production.txt
+edx-drf-extensions==4.0.1  # via -r requirements/dev.txt, -r requirements/production.txt
```
jmbowman: Please run `make upgrade` again so you can get version 4.0.2 and avoid downgrading djangorestframework.

credentials/apps/api/tests/test_authentication.py
jmbowman: This looks like the same problem we hit in the edx-platform tests; credentials doesn't use the `JwtAuthentication` class from edx-drf-extensions, so the change in drf-jwt to no longer catch `InvalidTokenError` is also impacting this subclass of that error.  We'll probably need to update this repo's [JwtAuthentication](https://github.com/edx/credentials/blob/master/credentials/apps/api/authentication.py#L40) class to similarly catch `InvalidTokenError` and re-raise as `AuthenticationFailed`.

Ayub-Khan: sure. I will update this in drf-extensions.

credentials/settings/base.py
```diff
@@ -389,7 +390,7 @@
 # DRF Settings
 REST_FRAMEWORK = {
     'DEFAULT_AUTHENTICATION_CLASSES': (
-        'credentials.apps.api.authentication.JwtAuthentication',
+        'edx_rest_framework_extensions.auth.jwt.authentication.JwtAuthentication',
```
jmbowman: @edx/arch-bom Does this seem like a reasonable change?  credentials hasn't been using the JwtAuthentication class from edx-drf-extensions, should it be?  If it should, are other changes needed in order to switch to it?  This is coming up because drf-jwt changed exception handling in a way that we already updated that authentication class to cope with, so the choice is between switching to it or making a similar fix to the one currently used here.

robrap: It should be, but there is at least one difference between the two classes and we need to determine if we are ok with temporarily losing that functionality in credentials, or if we want to add it to the edx-drf-extensions version.  See https://openedx.atlassian.net/browse/ARCH-269

dianakhuang:
@Ayub-Khan i believe the Cambridge team is going to try to take over this PR just so that we can get it over the line. Thanks for doing the original work!

requirements/all.txt
robrap: @dianakhuang: See @jmbowman's comment.  Need to run `make upgrade` again.

dianakhuang:
fyi @kdmccormick @schenedx , I ended up pinning `edx-ace` to the previous version because it broke tests. I didn't do any particular deep digging into it so that we could unblock the Django 2.2 upgrade, but I feel like you should know what that was pinned.

kdmccormick:
@kdmccormick Thanks for the tag - I'll look into why my `edx-ace` change may have broken something.

PR #766, opened 2020-04-08
Replace deprecated static loading

mzulqarnain1:


mzulqarnain1:
@jmbowman I made it for testing of [script in repo-tools](https://github.com/edx/repo-tools/blob/master/django3_codemods/script_to_replace_static.sh) but IMO we should merge this because we will have to do it eventually.

PR #769, opened 2020-04-10
Address EDUCATOR-4716: Backend

kdmccormick:
Please see https://openedx.atlassian.net/browse/EDUCATOR-4716

Most of the upgrades are to patch/minor versions, which I'm not worried about.

`pyyaml` was upgraded two major versions, but I'm familiar with the change-set and I feel comfortable with this upgrade.

`pillow` I am not familiar with, and it's been upgraded by three major versions. This is the one dependency upgrade that I'm not confident about.

@edx/masters-devs


requirements/all.txt
```diff
@@ -83,15 +83,14 @@ mysqlclient==1.4.6        # via -r requirements/dev.txt, -r requirements/product
...
-olefile==0.46             # via -r requirements/dev.txt, -r requirements/production.txt, pillow
 openapi-codec==1.3.2      # via -r requirements/dev.txt, -r requirements/production.txt, django-rest-swagger
 packaging==20.3           # via -r requirements/dev.txt, pytest, tox
 paramiko==2.7.1           # via -r requirements/dev.txt, docker
 path.py==12.0.2           # via -c requirements/constraints.txt, -r requirements/dev.txt, edx-i18n-tools
 pathlib2==2.3.5           # via -r requirements/dev.txt, pytest
 pbr==5.4.5                # via -r requirements/dev.txt, -r requirements/production.txt, stevedore
 pep8==1.7.1               # via -r requirements/dev.txt
-pillow==4.0.0             # via -c requirements/constraints.txt, -r requirements/dev.txt, -r requirements/production.txt
+pillow==7.1.1             # via -r requirements/dev.txt, -r requirements/production.txt
```
matthugs, line 93: It looks like this is a requirement as a result of our use of https://docs.djangoproject.com/en/3.0/ref/models/fields/#imagefield

Not sure whether there are aspects of their api that our code depends on, though. My inclination is to move forward with it with the limited time we spend on it doing a quick manual test (since if image file uploads are broken it might be a while before that gets escalated to us.)

PR #772, opened 2020-04-16
BOM-1523 : Added Python 3.8 To Travis Tests

mzulqarnain1:
Relevant JIRA : [https://openedx.atlassian.net/browse/BOM-1523](https://openedx.atlassian.net/browse/BOM-1523)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,10 @@
 language: python
 
-python:
-  - "3.5"
-
 branches:
   only:
     - master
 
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,8 +1,5 @@
 language: python
 
-python:
-  - "3.5"
```
jmbowman, line 4: Can we just run all 3 job types with 3.8 also?

PR #773, opened 2020-04-16
Added data migration

adeelehsan:
There are two course runs for course-v1:TUMx+QPLS2x+2T2018 and
one of them is excluded from the program. But user grades are
associated with excluded course run. That's why in program record
status of the course course-v1:TUMx+QPLS2x+2T2018 is not earned for
the learners even though learners program is completed.

Changing course run id in user grade would fix that issue.
PROD-1443

Credentials Pull Request
---

  - UI Changes 
    - [ ] Check mobile view


credentials/apps/records/migrations/0018_change_learners_course_run.py
```diff
@@ -0,0 +1,58 @@
...
+    UserGrade = apps.get_model('records', 'UserGrade')
+    CourseRun = apps.get_model('catalog', 'CourseRun')
+
+    courses = CourseRun.objects.filter(key='course-v1:TUMx+QPLS2x+2T2018')
+    if courses:
+        excluded_run = courses.filter(id=119)
+        included_run = courses.filter(id=8112)
+
+        if excluded_run and included_run:
+            for learner in UserGrade.objects.filter(course_run_id=excluded_run.id):
```
waheedahmed, line 29: Use bulk update.

credentials/apps/records/migrations/0018_change_learners_course_run.py
```diff
@@ -0,0 +1,58 @@
...
+    the learners even though learners program is completed.
+    PROD-1443
+
+    Changing course run id in user grade would fix that issue.
+
+    """
+    UserGrade = apps.get_model('records', 'UserGrade')
+    CourseRun = apps.get_model('catalog', 'CourseRun')
+
+    courses = CourseRun.objects.filter(key='course-v1:TUMx+QPLS2x+2T2018')
```
waheedahmed, line 23: Move all this into a helper function and call it from both.

credentials/apps/records/migrations/0018_change_learners_course_run.py
```diff
@@ -0,0 +1,70 @@
...
+    """
+    get excluded and included course runs based on the id
+    """
+    CourseRun = apps.get_model('catalog', 'CourseRun')
+    courses = CourseRun.objects.filter(key='course-v1:TUMx+QPLS2x+2T2018')
+    excluded_run = None
+    included_run = None
+    if courses:
+        excluded_run = courses.filter(id=excluded_run_id)
+        included_run = courses.filter(id=included_run_id)
```
waheedahmed, line 45: Do we need to call `.first()` as well on each of them, otherwise it will be `queyset` object instead of a single object and when we try to get the `id` it will raise an error?

PR #826, opened 2020-09-28
Update the Credentials program-listing endpoint to use Staff instead of

staubina:
superuser

[MICROBA-598]

The Credentials endpoint program-listing/ was meant for the support
team and internal teams to use as a debugging tool. The user type
Superuser was improperly used and now that less internal users
have that role it is unusable. Changing this to the Django Staff role
will allow internal users to access the tool again.

Credentials Pull Request
---

  - [x] Request a review if desired
  - [x] Squash/Fixup your branch to one commit
    - [x] Make sure you have updated [edx/configuration](https://github.com/edx/configuration) and [edx/devstack](https://github.com/edx/devstack) if necessary
    - [x] Make sure the change builds successfully in a sandbox
  - UI Changes 
    - [x] Consider other browsers (e.g. Firefox)
    - [x] Check mobile view
    - [x] Consider accessibility (e.g. Run aXe on the page)

### DIFF
--- a/credentials/apps/records/tests/test_views.py
+++ b/credentials/apps/records/tests/test_views.py
@@ -255,7 +255,7 @@ class ProgramListingViewTests(SiteMixin, TestCase):
         super().setUp()
         dump_random_state()
 
-        self.user = UserFactory(username=self.MOCK_USER_DATA['username'], is_superuser=True)
+        self.user = UserFactory(username=self.MOCK_USER_DATA['username'], is_staff=True)
         self.orgs = [OrganizationFactory.create(name=name, site=self.site) for name in ['TestOrg1', 'TestOrg2']]
         self.course = CourseFactory.create(site=self.site)
         self.course_runs = CourseRunFactory.create_batch(2, course=self.course)
@@ -328,8 +328,8 @@ class ProgramListingViewTests(SiteMixin, TestCase):
         self.assertRegex(response.url, '^/login/.*')
 
     def test_only_superuser_access(self):
-        """ Verify that the view rejects non-superusers. """
-        self.user.is_superuser = False
+        """ Verify that the view rejects non-staff users. """
+        self.user.is_staff = False
         self.user.save()
         self._render_listing(status_code=404)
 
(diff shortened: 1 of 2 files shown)


tuchfarber (approved):

credentials/apps/records/tests/test_views.py
```diff
@@ -328,8 +328,8 @@ def test_no_anonymous_access(self):
         self.assertRegex(response.url, '^/login/.*')
 
     def test_only_superuser_access(self):
-        """ Verify that the view rejects non-superusers. """
```
crice100, line 331: Do we have a test that ensures staff have access?

Suggestion: it might be good to leave the "is superuser" test here and just add another test for non-staff users.

mikix: `test_normal_access` (the happy path) is testing staff access, I believe.

mikix (approved):
Seems fine.

credentials/apps/records/tests/test_views.py
staubina: I am not sure what that will test @crice100 because I am not including the superuser role here at all. A Superuser would also need to be a Staff user to access the page with this change I believe.

Thoughts?

crice100: Is a superuser also guaranteed to be staff?

I lean toward having more explicit tests, especially around permissions if we're changing who has access to what. Not a big deal either way.

staubina: Ok, this is an straightforward change. Let me add it in.

staubina: I made an update here to ensure superuser still has access. In general, if you are Superuser you are probably also Staff, but it is not explicit permission. Also learner in testing if you take away Staff from a Superuser they can no longer access Django Admin 😰

crice100 (approved):

credentials/apps/records/tests/test_views.py
mikix: Really? Then what is the point of superusers? I thought the one thing they had going was full access to django admin, but maybe that is an LMS-specific thing (or I'm misremembering).

PR #835, opened 2020-10-05
Run make upgrade

crice100:
Changelogs for major version bumps:

- https://django-debug-toolbar.readthedocs.io/en/latest/changes.html
- https://django-statici18n.readthedocs.io/en/latest/changelog.html
- https://importlib-metadata.readthedocs.io/en/latest/changelog%20(links).html


crice100:
I don't get errors when running make upgrade locally, but the automated job does: https://build.testeng.edx.org/view/upgrade-python-requirements/job/credentials-upgrade-python-requirements/67/console

It seems safer to close this PR and focus on fixing the job (so we're using a consistent method of upgrading).

PR #844, opened 2020-10-09
MICROBA-625 Fix accessibility tests

crice100:
The goal of this PR is to fix the existing test failures and re-enable the accessibility tests.

In a future ticket, we'll come back and upgrade dependencies. See [MICROBA-653](https://openedx.atlassian.net/browse/MICROBA-653).

Original text color:
Before

Proposed text color:
After

### DIFF
--- a/credentials/static/sass/_base.scss
+++ b/credentials/static/sass/_base.scss
@@ -20,6 +20,12 @@ html, body {
 // ------------------------------
 // #TYPOGRAPHY
 // ------------------------------
+// TODO: Remove this when we upgrade paragon
+$link-color: #23419f;
+
+a {
+  color: $link-color;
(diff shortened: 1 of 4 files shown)


credentials/static/sass/_base.scss
```diff
@@ -20,6 +20,9 @@ html, body {
 // ------------------------------
 // #TYPOGRAPHY
 // ------------------------------
+a {
+  color: #23419f;
```
Tj-Tracy, line 24: We should make this a variable so we can set it everywhere once.

Tj-Tracy (changes requested):

Tj-Tracy (approved):
:shipit:

PR #850, opened 2020-10-20
MICROBA-628 Remove boto and unpin django-storages

crice100:
Remove _boto_ and unpin _django-storages_ (since we no longer need _boto_). Also unpin _importlib-metadata_ since that upgrade problem should be fixed.


justinhynes (approved):
LGTM

staubina (approved):

PR #855, opened 2020-10-21
ARCHBOM-1569: Add a mock toggles state endpoint

timmc-edx:
As part of Architecture's observability work, we're adding endpoints to IDAs that can dump the current state of various toggles.  This is a mock endpoint that accomplishes two things: 1) confirm that we can require a staff JWT in another IDA than edxapp, and 2) allow us to test integration of multiple IDAs into toggle state reports before the edx-toggles djangoapp is ready for use.

This can go away once edx-toggles is ready and integrated.

You can test locally in devstack: http://localhost:18150/mock-toggles

### DIFF
--- a/credentials/urls.py
+++ b/credentials/urls.py
@@ -26,7 +26,7 @@ from rest_framework_swagger.views import get_swagger_view
 
 from credentials.apps.core import views as core_views
 from credentials.apps.records.views import ProgramListingView
-from credentials.views import FaviconView
+from credentials.views import FaviconView, MockToggleStateView
 
 
 admin.autodiscover()
@@ -47,6 +47,7 @@ urlpatterns = oauth2_urlpatterns + [
     url(r'^records/', include(('credentials.apps.records.urls', 'records'), namespace='records')),
     url(r'^program-listing/', ProgramListingView.as_view(), name='program_listing'),
     url(r'^favicon\.ico$', FaviconView.as_view(permanent=True)),
+    url(r'^mock-toggles$', MockToggleStateView.as_view()),
     url(r'^hijack/', include('hijack.urls', namespace='hijack')),
 ]
 
--- a/credentials/views.py
+++ b/credentials/views.py
@@ -20,8 +20,7 @@ class FaviconView(RedirectView):
         return urljoin(site_configuration.homepage_url, '/favicon.ico')
 
 
-# pragma: no cover
-class MockToggleStateView(views.APIView):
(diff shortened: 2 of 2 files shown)


credentials/urls.py
```diff
@@ -47,6 +47,7 @@
     url(r'^records/', include(('credentials.apps.records.urls', 'records'), namespace='records')),
     url(r'^program-listing/', ProgramListingView.as_view(), name='program_listing'),
     url(r'^favicon\.ico$', FaviconView.as_view(permanent=True)),
+    url(r'^mock-toggles$', MockToggleStateView.as_view()),
```
robrap, line 50: Would it reduce work to have the mock live at the real endpoint, `/api/toggles/v0/state/`, so we can just swap it out for the real version later? Whatever you think is best is fine.

credentials/views.py
```diff
@@ -12,3 +17,38 @@ def get_redirect_url(self, *_args, **_kwargs):
...
+
+
+class MockToggleStateView(views.APIView):
+    """
+    A mock endpoint showing that we can require a staff JWT in this IDA,
+    and allowing us to test integration of multiple IDAs into toggle state
+    reports (ARCHBOM-1569). This can go away once edx-toggles is ready and
+    integrated.
+    """
+    authentication_classes = (JwtAuthentication, SessionAuthentication,)
```
robrap, line 29: I don't think you need this.  It should be default in this IDA.

robrap (approved):

robrap:
@timmc-edx: You can also use `# pragma: no cover` to drop coverage.  Not sure how high of a level you can add this. :)

credentials/views.py
timmc-edx: Ah interesting, I see that in `DEFAULT_AUTHENTICATION_CLASSES` now.

From the perspective of making this as close as possible to the later edx-toggles app (which would declare it explicitly, right?) I'm inclined to leave as-is, but I don't suppose it actually matters for our report testing.

credentials/urls.py
timmc-edx: I think I'm in favor of keeping it very obviously fake for the moment so that it will stick out like a sore thumb in our report generator until it is updated.

tuchfarber (approved):

credentials/views.py
robrap: Good point though on the extracted version.  We do want this to be explicit.  Thanks!

PR #860, opened 2020-10-23
MB-521 | Send email on Program Cert creation

justinhynes:
[MICROBA-521]
- Pull in a stripped version of the base_body email template from edx-platform to re-use the header/footer from other edX automated email messages
- Update Program model to include the 'type_slug' field from the Discovery API object
- Update copy_catalog to sync type_slug
- Adds migration for change to Catalog/Program model
- Update Program factory for type_slug addition
- Add new email templates for new automated email message going to learners
- Add unit tests for new functionality
- Fix existing unit tests after Program model update


justinhynes:
Example of how one of the emails being sent to a learner may look. Example is for a learner in a MicroBachelors Program.
image

justinhynes:
There will be some configuration-related PR's needed as well, but I wanted to wait until after the initial PR in Credentials was ready and looked at a bit by the team.

credentials/apps/catalog/migrations/0014_program_type_slug.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 2.2.16 on 2020-10-21 15:48
```
justinhynes, line 1: Credentials syncs and stores information from the Course-Discovery API. In other services we have moved away from using the Program type directly for comparisons as it may be translated (if memory serves correct). I am adding a `type_slug` field to the (Credentials) Program model, and updated the `copy_catalog` management command to pull in the program-type/slug from the Discovery API.

credentials/apps/credentials/messages.py
```diff
@@ -0,0 +1,23 @@
+from edx_ace import MessageType
```
justinhynes, line 1: For full disclosure...

This is lifted right out of the Credentials Records app's `messages.py`. I had hoped to possibly a Common `CredentialsMessage` class, hopefully in the Core app, and then have both the Credentials and Records app use the common Class.

I thought that implementing this change and testing that everything is still working as expected may take longer than I expected given my familiarity with Credentials.

Hopefully something to try iterating on...

credentials/apps/credentials/utils.py
```diff
@@ -89,3 +94,51 @@ def get_credential_visible_dates(user_credentials):
...
+                    'edx_share_cert_kb': email_config['edx_share_cert_kb'],
+                    'edx_mb_redeem_credit_kb': email_config['edx_mb_redeem_credit_kb'],
+                    'edx_mb_credit_faq_kb': email_config['edx_mb_credit_faq_kb'],
+                    'tesu_mb_redeem_credit_kb': email_config['tesu_mb_redeem_credit_kb'],
+                    'security_plus_info': email_config['security_plus_info'],
+                    'schedule_security_plus': email_config['schedule_security_plus'],
+                    'comptia_discount_code': email_config['comptia_discount_code'],
+                },
+            )
+            log.info("Sending Program completion congratulatory email for learner with id [{}]".format(
```
justinhynes, line 135: Self-critique 

I wanted to come back around and improve these log messages. I think it would be helpful to include the Program UUID as well to give greater context to anyone reading logs related to this feature. I missed this before putting it up for review.

credentials/settings/base.py
```diff
@@ -464,3 +464,15 @@
...
+# .. toggle_implementation: DjangoSetting
+# .. toggle_default: False
+# .. toggle_description: Toggle to control if we send a congratulatory email to learners after being issued a Program Certificate
+# .. toggle_use_cases: temporary
+# .. toggle_creation_date: 2020-10-08
+# .. toggle_target_removal_date: 2020-12-01
+# .. toggle_warnings: This can be removed once we feel confident the new feature is working as desired
+# .. toggle_tickets: MICROBA-521
+SEND_EMAIL_ON_PROGRAM_COMPLETION = False
+PROGRAM_CERTIFICATE_ISSUED_EMAIL_METADATA = None
```
justinhynes, line 478: This was an attempt to have a dict containing all the other URL's (and the Security+ Discount code) we want to embed in the various emails. I didn't want to hard-code each URL into the `send_program_certificate_created_message` utility method I am adding. I wanted the links to be easy-ish to update if required.

credentials/apps/catalog/migrations/0014_program_type_slug.py
crice100: I like including _slug_. This is helpful, thanks. 

For future reference, this might be a good thing to do in a separate PR.

credentials/apps/credentials/messages.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+class ProgramCertificateIssuedMessage(MessageType):
+    def __init__(self, site, user_email=None, *args, **kwargs):  # pylint: disable=keyword-arg-before-vararg
+        super(ProgramCertificateIssuedMessage, self).__init__(*args, **kwargs)
+
+        if site.siteconfiguration.partner_from_address:
+            from_address = site.siteconfiguration.partner_from_address
+        else:
+            from_address = 'no-reply@' + site.domain
```
crice100, line 11: Maybe log an info message if we hit this line? Or is this exactly what we're doing in the other credentials app?

credentials/apps/credentials/messages.py
```diff
@@ -0,0 +1,23 @@
...
+
+        if user_email:
+            self.options.update({
+                'reply_to': [user_email],
```
crice100, line 15: Does this make sense for a reply to address?

credentials/apps/credentials/tests/test_utils.py
```diff
@@ -37,3 +47,327 @@ def test_datetime_from_visible_date(self):
...
+    def _setup_program_and_program_cert(self, program_type, is_nyu_microbachelors_program=False):
+        """
+        Utility method for setting up Program data for email tests.
+        """
+        self.program = ProgramFactory(site=self.site)
+        self.program.type = program_type
+        self.program.type_slug = slugify(program_type)
+
+        if is_nyu_microbachelors_program:
+            self.program.uuid = '401aa2f5-e964-4276-b76c-0a3ff7e5b053'
```
crice100, line 92: Maybe use a fake college and UUID

credentials/apps/credentials/tests/test_utils.py
```diff
@@ -37,3 +47,327 @@ def test_datetime_from_visible_date(self):
...
+        # - nyu_mb_cybersecurity_fragments appear to learners enrolled in the NYU CyberSecurity Fundamentals Program
+        email_fragments = [
+            'Congratulations on completing the {} {} Program!'.format(
+                self.program.title,
+                self.program.type
+            ),
+            'Get your Program Certificate: {}'.format(
+                settings['edx_locate_program_cert_kb']
+            ),
+            'Questions? Email: info@edx.org'
```
crice100, line 122: Suggestion: use a fake email address

credentials/apps/catalog/migrations/0014_program_type_slug.py
justinhynes: Do you think it would be beneficial for me to break it out now?

credentials/apps/credentials/messages.py
justinhynes: This is what we are doing in the other Credentials app. I'll add a log line for more insight in case we ever need it.

credentials/apps/credentials/messages.py
justinhynes: Leaving a comment for the record since we discussed a little bit of this in person on a meeting today. No, I don't. I _thought_ I had seen a comment, somewhere, with an explanation as to why it was set to the user_email value.

I'm having trouble digging that up right now... so I am just going to change it to the from_address. We can always adjust again if needed.

credentials/apps/credentials/tests/test_utils.py
justinhynes: Added email to config options. Since emails change I figured it was better than hard-coding that support email in the message templates.

credentials/apps/credentials/tests/test_utils.py
justinhynes: Scrubbed files of the UUID/NYU mentions.

credentials/apps/catalog/migrations/0014_program_type_slug.py
justinhynes: For the record: Christie and I discussed this a bit further and I agree 100% with the point about splitting it up making it easier to handle/roll-back/cut down on huge PR's.

I am choosing not to break this apart for now in the interest of priority of the story.

credentials/apps/credentials/messages.py
justinhynes: Adding as part of my next PR changes. Will follow up with another task as well so we don't lose sight of this.

credentials/apps/credentials/messages.py
```diff
@@ -0,0 +1,28 @@
...
+    def __init__(self, site, user_email=None, *args, **kwargs):  # pylint: disable=keyword-arg-before-vararg
+        super(ProgramCertificateIssuedMessage, self).__init__(*args, **kwargs)
+
+        if site.siteconfiguration.partner_from_address:
+            from_address = site.siteconfiguration.partner_from_address
+        else:
+            log.info("No partner from address found. Using default no-reply@{}".format(site.domain))
+            from_address = 'no-reply@' + site.domain
+
+        if user_email:
```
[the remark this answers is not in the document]
justinhynes: ... you're 100% correct. Hah. Sorry, that was a total rush job. This will be fixed in next round of changes.

credentials/apps/credentials/messages.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+log = logging.getLogger(__name__)
+
+
+# This code also exists in the Records app `messages.py` file. Any changes here should be duplicated there as well
+# until we can come back around and create a common base Messaging class that the Credentials and Records app will
+# utilize.
+class ProgramCertificateIssuedMessage(MessageType):
+    def __init__(self, site, user_email=None, *args, **kwargs):  # pylint: disable=unused-argument, keyword-arg-before-vararg
```
justinhynes, line 13: We're no longer using `user_email`... but when I tried to remove it things got _very_ sad and exploded... so I asked pylint to stop yelling at me instead.

credentials/apps/credentials/issuers.py
```diff
@@ -147,6 +148,11 @@ def issue_credential(
         if created and site_config and site_config.records_enabled:
             send_updated_emails_for_program(request, username, credential)
 
+        # If this is a new ProgramCertificate and the `SEND_EMAIL_ON_PROGRAM_COMPLETION`
+        # feature is enabled then let's send a congratulatory message to the learner
+        if created and getattr(settings, 'SEND_EMAIL_ON_PROGRAM_COMPLETION', False):
```
crice100, line 153: Just want to confirm I'm reading this correctly (that we won't end up sending multiple emails to a user):

This new if sends an email to the user.

The above if sends an email to the pathway org.

credentials/apps/credentials/issuers.py
```diff
@@ -147,6 +148,11 @@ def issue_credential(
...
+            send_program_certificate_created_message(username, credential)
```
crice100, line 154: If for some reason in the future we disable all pathways (or even just the pathways associated with this particular program), should we still send the email? If yes, should the email content change?

credentials/apps/credentials/issuers.py
justinhynes: Yes, that is my understanding as well. 

`send_updated_emails_for_program` is there to automatically resend records to a pathway if a learner has previously shared their records before finishing a Program and earning the credential. The email recipient is defined here: https://github.com/edx/credentials/blob/c13b9dec0bbaf12f796615b8588e290d06c6f13c/credentials/apps/records/utils.py#L50 and would not go to the learner.

The new section I'm adding is being sent to the learner.

credentials/apps/credentials/issuers.py
justinhynes: Interesting question. I had not thought of this either. Going to think about this for a few and get back to you. :)

credentials/apps/credentials/utils.py
```diff
@@ -90,3 +96,44 @@ def get_credential_visible_dates(user_credentials):
...
+                    custom_completion_email_settings.get(str(program_uuid), {}).get('html')
+                    or custom_completion_email_settings.get(program.type_slug, {}).get('html')
+                    or ""
+                ),
+                # remove any leading spaces of the plaintext content the email doesn't look horrendous
+                'custom_email_plaintext_template_extra': textwrap.dedent((
+                    custom_completion_email_settings.get(str(program_uuid), {}).get('plaintext')
+                    or custom_completion_email_settings.get(program.type_slug, {}).get('plaintext')
+                    or ""
+                )),
```
tuchfarber, line 128: Nit: you could pull the duplication out to the top.
```
# at top 
custom_template_thing = (
    custom_completion_email_settings.get(str(program_uuid), {})
    or custom_completion_email_settings.get(program.type_slug, {})
    or {}
)

# in user_context
custom_template_thing.get("html", "")
textwrap.dedent(custom_template_thing.get("plaintext", ""))
```

credentials/templates/credentials/edx_ace/programcertificateissuedmessage/email/body.html
```diff
@@ -0,0 +1,24 @@
...
+<table width="100%" align="left" border="0" cellpadding="0" cellspacing="0" role="presentation">
+    <tr>
+        <td>
+            <p style="color: rgba(0,0,0,.75);">
+                {% trans "Hello" %},
+            </p>
+            <p style="color: rgba(0,0,0,.75);">
+                {% blocktrans %}Congratulations on completing the {{ program_title }} {{ program_type }} Program!{% endblocktrans %}
+            </p>
+            {{ custom_email_html_template_extra |safe }}
```
tuchfarber, line 16: Since we're not having user input yet, `|safe` is still good. But if we could just put a comment to fix this when we move to user generate text.

tuchfarber (approved):

crice100 (approved):

credentials/apps/credentials/issuers.py
justinhynes: For the record:
I had a chat with Ryan about this question and an additional conversation with Christie. There is still value sending the email even if potential pathways were disabled. We sided on email content _should_ change if something like this were to happen, but we will handle it then if required over trying to add to this work now.

credentials/apps/credentials/utils.py
justinhynes: Thanks! I'll make this change.

PR #861, opened 2020-10-23
Ignore -- Ran make upgrade

crice100:


crice100:
Closing in favor of https://github.com/edx/credentials/pull/862, which is identical

PR #865, opened 2020-10-27
Ignore -- Unpin path

crice100:
Changelogs: https://path.readthedocs.io/en/latest/history.html


crice100:
Unpinning _path_ was merged here: https://github.com/edx/credentials/pull/873

PR #873, opened 2020-11-04
MICROBA-645 Remove path.py constraint

crice100:
Note: we expect the _quality-and-translations_ build to fail due to [path.py being renamed to path](https://path.readthedocs.io/en/latest/history.html) and problems with upgrading prior versions of that.

Once this is merged, a docker image will be created, and any subsequent build should be fine. Discovery took a similar approach when upgrading path: https://github.com/edx/course-discovery/pull/2315

Once the edx i18n-tools repo is updated to depend on path instead of path.py (see [CR-2909](https://openedx.atlassian.net/servicedesk/customer/portal/9/CR-2909)), other repos that are similarly behind shouldn't encounter this issue.


tuchfarber (approved):

crice100:
A later test PR (created after the new credentials image was built and deployed), was green: https://github.com/edx/credentials/pull/875

PR #877, opened 2020-11-05
MICROBA-698 Add push_translations and upgrade credentials-themes

crice100:
* Add _push_translations_ to makefile
* Get _help_ command in sync with other repos
* Upgrade _credentials-themes_ and run _make upgrade_

Note: edX seems inconsistent on the implementation of _push_translations_ and _pull_translations_ (sometimes we call the _i18n_tool_ and other times we call _tx_ directly). I picked the approach that seems more common (call _tx_ directly) but suggestions are welcome.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,14 @@
-.DEFAULT_GOAL := tests
+.DEFAULT_GOAL := help
 NODE_BIN=./node_modules/.bin
 TOX = ''
 
-.PHONY: requirements upgrade piptools production-requirements all-requirements isort isort_check pycodestyle \
-        pii_check exec-pii_check
+.PHONY: help clean production-requirements js-requirements all-requirements requirements isort isort_check pycodestyle \
+        quality quality-js test-react tests js-tests static static.dev static.watch migrate up up-dev up-test \
+        exec-validate-translations exec-check_translations_up_to_date exec-check_keywords exec-pii_check exec-clean \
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -12,7 +17,7 @@ endif
 # Generates a help message. Borrowed from https://github.com/pydanny/cookiecutter-djangopackage.
 help: ## Display this help message
 	@echo "Please use \`make <target>\` where <target> is one of"
-	@perl -nle'print $& if m{^[\.a-zA-Z_-]+:.*?## .*$$}' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m  %-25s\033[0m %s\n", $$1, $$2}'
+	@awk -F ':.*?## ' '/^[a-zA-Z]/ && NF==2 {printf "\033[36m  %-25s\033[0m %s\n", $$1, $$2}' $(MAKEFILE_LIST) | sort
```
tuchfarber, line 20: Just a copy from the other Makefiles?

tuchfarber (approved):

Makefile
crice100: Yep, copied from the cookiecutter

https://github.com/edx/edx-cookiecutters/blob/fde73eb56e85ea0fb98e66b1c7d7f3bb76d1562d/cookiecutter-django-app/Makefile#L7

PR #884, opened 2020-11-09
Add program email configuration models

tuchfarber:
In replacement of the settings based email configuration, this allows
editing in Django Admin. It has a similar hierarchical model where the
most specific key (program, then program type, then default) that
matches becomes the setting used. In addition, this allows us to enable
or disable the sending of the emails at each level.


staubina (approved):
Some questions and comments about the help text in the admin portal. Overall makes sense to me.

credentials/apps/credentials/models.py
```diff
@@ -304,3 +304,55 @@ class UserCredentialAttribute(TimeStampedModel):
 
     class Meta:
         unique_together = (('user_credential', 'name'),)
+
+class ProgramCompletionEmailConfiguration(TimeStampedModel):
+    """
+    Template to add additional content into the program completion emails.
+
+    specificity: this is the level of emails which will be affected by the
+    entry in the table. It can either be the default
```
crice100, line 313: Nit: sentence seems to end early

credentials/apps/credentials/models.py
```diff
@@ -304,3 +304,55 @@ class UserCredentialAttribute(TimeStampedModel):
...
+    """
+
+    DEFAULT_TEMPLATE_IDENTIFIER = "default"
+
+    # identifier will either be a:
+    # - UUID <string> (for a specific program)
+    # - program type <string> (for a program type)
+    # - or "default" (the DEFAULT_TEMPLATE_IDENTIFIER) to be the global template used for all programs
+    identifier = models.CharField(
```
crice100, line 322: Can we add validation to make sure the identifier value exists in credentials (that the slug or UUID hasn't been typo'ed)? Even just logging something in splunk could be useful

credentials/apps/credentials/models.py
```diff
@@ -304,3 +304,55 @@ class UserCredentialAttribute(TimeStampedModel):
...
+        unique=True,
+        help_text="""Should be either "default" to affect all programs, the program type slug, or the UUID of the program"""
+    )
+    html_template = models.TextField()
+    plaintext_template = models.TextField()
+    enabled = models.BooleanField(default=False)
+    history = HistoricalRecords()
+
+    @classmethod
+    def get_email_config_for_program(cls, program):
```
crice100, line 333: Could we add unit tests for this?

credentials/apps/credentials/utils.py
```diff
@@ -107,24 +108,24 @@ def send_program_certificate_created_message(username, program_certificate):
...
+
     try:
         msg = ProgramCertificateIssuedMessage(program_certificate.site, user.email).personalize(
             recipient=Recipient(username=user.username, email_address=user.email),
             language=program_certificate.language,
             user_context={
                 'program_title': program.title,
                 'program_type': program.type,
-                'custom_email_html_template_extra': custom_completion_email_content.get('html', ''),
+                'custom_email_html_template_extra': bleach.clean(email_configuration.html_template, tags=settings.ALLOWED_EMAIL_HTML_TAGS),
```
crice100, line 126: Shouldn't we call bleach on save?

justinhynes:
Since we are relying on bleach to remove any content in the input we need to escape/strip, we would still need the `safe` keyword to prevent the good content being escaped in the template, right?

Does my comment [here](https://github.com/edx/credentials/blob/f1566ac279675ae3a32f294da5a11a9569b7e279/credentials/templates/credentials/edx_ace/programcertificateissuedmessage/email/body.html#L16) make sense anymore then? Can we just delete my comment?

staubina (approved):

PR #891, opened 2020-11-18
Add django versions and codecov to CI

tuchfarber:


.github/workflows/ci.yml
```diff
@@ -42,13 +42,44 @@ jobs:
...
+          make exec-static
+          make exec-tests
+
+      - uses: actions/setup-python@v2
+        with:
+          python-version: "3.8"
+          architecture: x64
+      - name: Run code coverage
+        run: |
+          make exec-coverage
```
crice100, line 83: Should lines 83 and 84 be switched?

tuchfarber: `make exec-coverage` runs inside the container to take the coverage stuff and turn it into XML. 
Then, locally, we install codecov so we can send stuff to codecov

crice100 (approved):

PR #899, opened 2020-11-30
Adjust bleach config for allowed html tags

justinhynes:
Adjust bleach config for allowed html tags


crice100 (approved):
Love the alphabetization!

PR #900, opened 2020-11-30
Switch program_details to use in service catalog [MICROBA-726]

tuchfarber:
Instead of relying on cached calls to the discovery API, this uses the
in-service catalog to support program_details which certificates use.

Also removed subtitle since no code is actively using it.

### DIFF
--- a/credentials/apps/catalog/data.py
+++ b/credentials/apps/catalog/data.py
@@ -1,0 +1,34 @@
+"""
+Public data structures for this app.
+
+This should only ever be consumed and should never import anything other than
+stdlib modules, so that it may be consumed everywhere.
+"""
+
+from dataclasses import dataclass
+
+
+# Currently using a dataclass version of the legacy namedtuples so the consuming
+# app can mutate them for their needs. Future version should use frozen
+# dataclasses and consuming app can create their own internal structures using
+# the values from the frozen data rather than dataclass with too many
+# requirements.
+@dataclass
+class OrganizationDetails:
+    uuid: str
+    key: str
+    name: str
+    display_name: str
+    certificate_logo_image_url: str
+
+
+# See comment above OrganizationDetails
+@dataclass
+class ProgramDetails:
+    uuid: str
+    title: str
+    type: str
+    credential_title: str
+    course_count: int
+    organizations: list
+    hours_of_effort: int
--- a/credentials/apps/credentials/models.py
+++ b/credentials/apps/credentials/models.py
@@ -1,8 +1,8 @@
 """
 Models for the credentials service.
 """
+import logging
 import uuid
-from collections import namedtuple
 
 import bleach
 from django.conf import settings
@@ -19,8 +19,13 @@ from opaque_keys import InvalidKeyError
 from opaque_keys.edx.keys import CourseKey
 from simple_history.models import HistoricalRecords
 
+from credentials.apps.catalog.api import get_program_details_by_uuid
 from credentials.apps.core.utils import _choices
 from credentials.apps.credentials import constants
+from credentials.apps.credentials.exceptions import NoMatchingProgramException
+
+
+log = logging.getLogger(__name__)
 
 
 def signatory_assets_path(instance, filename):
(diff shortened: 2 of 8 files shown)


credentials/apps/credentials/models.py
```diff
@@ -215,7 +216,7 @@ def course_key(self):
 
 OrganizationDetails = namedtuple('OrganizationDetails', ('uuid', 'key', 'name', 'display_name',
                                                          'certificate_logo_image_url'))
-ProgramDetails = namedtuple('ProgramDetails', ('uuid', 'title', 'subtitle', 'type', 'credential_title', 'course_count',
+ProgramDetails = namedtuple('ProgramDetails', ('uuid', 'title', 'type', 'credential_title', 'course_count',
```
crice100, line 219: Do any of the changes in this file need a new migration generated?

credentials/apps/credentials/models.py
```diff
@@ -262,34 +263,32 @@ def get_program_api_data(self):
...
-                name=organization['name'],
-                display_name=organization['name'] if self.use_org_name else organization['key'],
-                certificate_logo_image_url=organization['certificate_logo_image_url']
-            ))
-
-        hours_of_effort = None
-        if self.include_hours_of_effort:
-            hours_of_effort = data.get('total_hours_of_effort')
+        # not catching DoesNotExist because a cert without a matching actual program should blow up
+        program = Program.objects.get(uuid=self.program_uuid)
```
crice100, line 270: Is this changing the code to call the credentials db? If yes, any thoughts on whether this might increase the load on the db, or whether we might need a new index, etc?

credentials/apps/credentials/models.py
tuchfarber: Nope, because it's just a `namedtuple` it's only in code not the DB :)

credentials/apps/credentials/models.py
tuchfarber: Hmmm. Yeah, it'll hit the DB instead of memcached discovery data now. I'll add a UUID index on the Program model. Good catch!

crice100 (dismissed):

credentials/apps/credentials/models.py
staubina: Is this necessary? This seems like a bit of a pre optimization. I am not sure the service gets enough requests to require a cache here.

credentials/apps/credentials/models.py
```diff
@@ -262,34 +263,32 @@ def get_program_api_data(self):
...
-                key=organization['key'],
-                name=organization['name'],
-                display_name=organization['name'] if self.use_org_name else organization['key'],
-                certificate_logo_image_url=organization['certificate_logo_image_url']
-            ))
-
-        hours_of_effort = None
-        if self.include_hours_of_effort:
-            hours_of_effort = data.get('total_hours_of_effort')
+        # not catching DoesNotExist because a cert without a matching actual program should blow up
```
staubina, line 269: Should we consider catching this and adding a helpful log statement here and then raise the exception again? We could then potentially alert on the log statement to we can preemptively call out that we have a cert without a program.

This could happen if syncing on either end fails.

staubina (approved):
LGTM, just some non-blocking questions.

credentials/apps/catalog/data.py
```diff
@@ -0,0 +1,34 @@
...
+"""
+
+from dataclasses import dataclass
+
+
+# Currently using a dataclass version of the legacy namedtuples so the consuming
+# app can mutate them for their needs. Future version should use frozen
+# dataclasses and consuming app can create their own internal structures using
+# the values from the frozen data rather than dataclass with too many
+# requirements.
```
staubina, line 15: This whole file is interesting to me, can we talk about this one at some point? I think there is something for me to learn here.

staubina (approved):
This looks like a really good refactor and a step in the right direction for this code.

crice100 (approved):

crice100:
Adding jira ticket numbers in the headline and/or description would be great

PR #902, opened 2020-12-01
Update toggle_use_case

feanil:
`permanent` is not a valid setting for this.  The use cases are defined here: https://github.com/edx/edx-toggles/blob/master/docs/how_to/documenting_new_feature_toggles.rst#toggle_use_cases.

Based on the use case definitions, it looks like this most closely matches the "Open edX Option" case:

| When a team implements a feature that they do not expect to be adopted by all Open edX instances, they may introduce a toggle to gate the feature.

### DIFF
--- a/credentials/settings/base.py
+++ b/credentials/settings/base.py
@@ -466,7 +466,7 @@ MEDIA_STORAGE_BACKEND = {
 SOCIAL_AUTH_REDIRECT_IS_HTTPS = False
 
 # .. toggle_name: SEND_EMAIL_ON_PROGRAM_COMPLETION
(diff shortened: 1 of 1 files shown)


justinhynes (approved):
Thanks!

credentials/settings/base.py
```diff
@@ -469,7 +469,7 @@
 # .. toggle_implementation: DjangoSetting
```
justinhynes, line 469: Oh, crud, should this be updated to "SettingToggle"?

feanil: Ah, good catch, I'll update it.

PR #908, opened 2020-12-04
Tuchfarber/disable program emails when retired

tuchfarber:


staubina:
Should this PR be base on Master or on PR#900?

credentials/apps/catalog/api.py
```diff
@@ -43,9 +43,11 @@ def _convert_program_to_program_details(
     return ProgramDetails(
         uuid=program.uuid,
         title=program.title,
-        type=program.type_slug,
+        type=program.type,
```
crice100, line 46: Is this going to break any existing code that expects _type_ to be a slug?

credentials/apps/catalog/data.py
```diff
@@ -28,7 +29,16 @@ class ProgramDetails:
...
     type: str
+    type_slug: str
     credential_title: str
     course_count: int
     organizations: list
     hours_of_effort: int
+    status: str
+
+
+class ProgramStatus(Enum):
```
crice100, line 40: I assume these also live in discovery. What happens if they get out of sync (ex. discovery adds a new type)?

credentials/apps/catalog/api.py
tuchfarber: So, the original code was actually `type: program.type`, but I accidentally changed it to slug and the only thing that has a check that catches that was the email tests, which were still using the old stuff before this PR. So is just getting back to normal.

credentials/apps/catalog/data.py
tuchfarber: If discovery has a new type we'll need to add it here manually. This was already happening inside the `Program` class, I just pulled it out into an importable space. Since discovery just stores a `CharField` with `choices`, and all we receive is string, I don't know if there *is* a way to keep this in sync automatically. If you have any ideas though, I'm all for it.

PR #911, opened 2020-12-08
[MICROBA-744] Update logos to source from config parameters

staubina:


credentials/templates/_footer.html
```diff
@@ -29,7 +29,13 @@
             </div>
         </div>
         <div class="footer-app-related">
-            <a class="img-link" href="https://open.edx.org/"><img class="logo-img logo-openedx" src="{% static 'images/edx-openedx-logo-tag.png' %}" alt="{% trans 'Powered by Open edX' as tmsg %}{{tmsg | force_escape}}"/></a>
+            <a class="img-link" href="https://open.edx.org/">
+              <img
+                  class="logo-img logo-openedx"
+                  src="{{ openedx_logo_url }}"
+                  alt="{% trans 'Powered by Open edX' as tmsg %}{{tmsg | force_escape}}"
```
justinhynes, line 36: I know this is how it was done previously, and I understand that this is just alt-text, but it feels weird that we are translating `Open edX`. Is that not considered a product/IP we, edX, own?

credentials/settings/base.py
```diff
@@ -495,3 +495,15 @@
     'strong',
     'ul'
 ]
+
+LOGO_TRADEMARK_URL = 'https://edx-cdn.org/v3/default/logo-trademark.svg'
```
justinhynes, line 499: Do we need to add all of our specific edX logo settings here? 

`edx-internal` was already updated to add these, and it struck me as odd we would add these here in the general credentials settings for Open edX.

Should these just be placeholders with default settings?

staubina: They are the default logos. I can remove the ones we do not use, but I think it is helpful for the community to see what changing them will do.

credentials/apps/core/context_processors.py
```diff
@@ -1,4 +1,5 @@
 """ Core context processors. """
+from django.conf import settings
 
 
 def core(request):
```
justinhynes, line 5: For my knowledge...

This is the first time I've seen a `context_processors` file. Does this mean that these values are available anywhere in the apps? Or is it limited just to the `core` app?

justinhynes (approved):

credentials/settings/base.py
justinhynes: Oh, you're right. Bad assumption on my part. Thanks.

credentials/apps/core/context_processors.py
staubina: This was new to me as well. This processor adds context to every view rendered using a template: https://github.com/edx/credentials/blob/bfb0652af63c86ef1dc3aec07dfbf14ef9158397/credentials/settings/base.py#L278

staubina: Note from the docs.

> context_processors is a list of dotted Python paths to callables that are used to populate the context when a template is rendered with a request. These callables take a request object as their argument and return a dict of items to be merged into the context.

https://docs.djangoproject.com/en/3.0/ref/templates/api/

PR #926, opened 2020-12-18
MICROBA-639 Fix warnings

crice100:
Fixes:

> 
> /edx/app/credentials/venvs/credentials/lib/python3.8/site-packages/django/db/models/fields/__init__.py:1424: RuntimeWarning: DateTimeField CourseRun.end_date received a naive datetime (2018-06-01 00:00:00) while time zone support is active.
>   warnings.warn("DateTimeField %s received a naive datetime (%s)"

and

> edx-credentials@ test-react /edx/app/credentials/credentials
> > jest --no-cache --coverage  --config ./jest-config.json
> 
> jest-haste-map: Watchman crawl failed. Retrying once with node crawler.
>   Usually this happens when watchman isn't running. Create an empty `.watchmanconfig` file in your project's root folder or initialize a git or hg repository in your project.
>   Error: Watchman error: A non-recoverable condition has triggered.  Watchman needs your help!
> The triggering condition was at timestamp=1608310865: inotify-add-watch(/edx/app/credentials/credentials/.tox/django31/lib/python3.8/site-packages/openid/yadis/__pycache__) -> The user limit on the total number of inotify watches was reached; increase the fs.inotify.max_user_watches sysctl
> All requests will continue to fail with this message until you resolve
> the underlying problem.  You will find more information on fixing this at
> https://facebook.github.io/watchman/docs/troubleshooting.html#poison-inotify-add-watch. Make sure watchman is running for this project. See https://facebook.github.io/watchman/docs/troubleshooting.
>

### DIFF
--- a/credentials/apps/records/management/commands/seed-records.py
+++ b/credentials/apps/records/management/commands/seed-records.py
@@ -13,6 +13,7 @@ from credentials.apps.credentials.constants import CertificateType
 from credentials.apps.credentials.models import CourseCertificate, ProgramCertificate, Signatory, UserCredential
 from credentials.apps.records.constants import UserCreditPathwayStatus
 from credentials.apps.records.models import ProgramCertRecord, UserCreditPathway, UserGrade
+from credentials.settings.base import TIME_ZONE_CLASS
 from credentials.shared.constants import PathwayType
 
 
@@ -143,8 +144,8 @@ class Command(BaseCommand):
             course_run, created = CourseRun.objects.get_or_create(
                 course=course,
                 uuid=faker.uuid4(),
-                start_date=datetime(2018, 1, 1),
-                end_date=datetime(2018, 6, 1),
+                start_date=datetime(2018, 1, 1, tzinfo=TIME_ZONE_CLASS),
+                end_date=datetime(2018, 6, 1, tzinfo=TIME_ZONE_CLASS),
                 key=key,
             )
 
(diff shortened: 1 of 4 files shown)


credentials/apps/records/management/commands/seed-records.py
```diff
@@ -143,8 +143,8 @@ def seed_course_runs(courses, faker):
             course_run, created = CourseRun.objects.get_or_create(
                 course=course,
                 uuid=faker.uuid4(),
-                start_date=datetime(2018, 1, 1),
-                end_date=datetime(2018, 6, 1),
+                start_date=datetime(2018, 1, 1, tzinfo=timezone.utc),
+                end_date=datetime(2018, 6, 1, tzinfo=timezone.utc),
```
tuchfarber, line 147: We store `TIME_ZONE` as `UTC` in `settings/base.py`. Might make sense to pull it from settings, just in case, but I'm fine with it as is.

tuchfarber (approved):

credentials/apps/records/management/commands/seed-records.py
crice100: Thanks! I hadn't realized it was in there. 

Turns out that existing constant is a string, and I need a sub-class of timezone, so I added a new constant next to it. I didn't rename the existing one because it matches what we have in the cookiecutter, so some people may be expecting it to exist as-is (as a string).

PR #956, opened 2021-01-13
Update gulp-eslint package

tuchfarber:
Updating gulp-eslint for security. As it's a testing requirement, if tests pass, then we're probably okay :)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ all-requirements: piptools ## Install local and prod requirements
 	pip-sync requirements/all.txt
 
 requirements: piptools ## Install requirements for local development
-	npm install
+	npm install --also=dev
 	pip-sync requirements/dev.txt
 
 piptools:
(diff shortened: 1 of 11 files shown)


Makefile
```diff
@@ -165,12 +165,12 @@ build_test_image: # Builds Docker image used for testing so devs don't need to i
 
 # This should be ran locally, not inside of the devstack container
 acceptance_tests_suite: build_test_image
-	docker run -e "TERM=xterm-256color" credentials:local bash -c 'cd /edx/app/credentials/ && make static && make accept'
+	docker run -e "TERM=xterm-256color" -v `pwd`:/edx/app/credentials/ --read-only credentials:local bash -c 'cd /edx/app/credentials/ && make static && make accept'
```
crice100, line 168: I assume we tested the make commands and they're still passing successfully?

crice100 (approved):

Makefile
tuchfarber: Yep yep! I was comparing the two as I worked on it.

PR #966, opened 2021-01-20
Fix node_modules Mac issues

tuchfarber:
Create node_modules volume instead of bind mounting local folder. Bind
mounting doesn't work because node-sass has environment bindings.


oliviaruizknott (approved):
Approving because this helped me! But I don’t know what’s going on here so you could use a more discerning set of eyes ...

PR #969, opened 2021-01-22
DEPR-127: Convert from EdxRestApiClient to OAuthAPIClient

mikix:


credentials/apps/core/models.py
```diff
@@ -252,17 +210,6 @@ class User(AbstractUser):
 
     full_name = models.CharField(_("Full Name"), max_length=255, blank=True, null=True)
 
-    @property
-    def access_token(self):
```
mikix, line 256: This didn't seem used anywhere? So I just dropped it. Not necessary as part of my conversion, I can add it back if it is necessary. Just some unrelated cleanup I noticed.

credentials/apps/core/tests/test_models.py
```diff
@@ -20,20 +19,6 @@ def setUp(self):
...
-    def test_access_token_without_social_auth(self):
-        """ Verify the property returns None if the user is not associated with a UserSocialAuth. """
-        self.assertIsNone(self.user.access_token)
-
-    def test_access_token(self):
-        """ Verify the property returns the value of the access_token stored with the UserSocialAuth. """
-        social_auth = UserSocialAuth.objects.create(user=self.user, provider="test", uid=self.user.username)
-        self.assertIsNone(self.user.access_token)
-
-        access_token = "My voice is my passport. Verify me."
```
mikix, line 32: I do love this reference.

credentials/apps/core/models.py
staubina: I am not 100% sure, but I think this method needs to exist for oauth. @crice100 Any idea if this is needed?

crice100: I don't know, sorry

tuchfarber (approved):

PR #983, opened 2021-02-18
Add logging when issuing credentials

justinhynes:
Add some logging when attempting to issue a credential so that we have a better idea as to what learner is affected should an issue arise.

### DIFF
--- a/credentials/apps/credentials/issuers.py
+++ b/credentials/apps/credentials/issuers.py
@@ -3,6 +3,7 @@ import abc
 import logging
 
 from django.conf import settings
+from django.contrib.auth import get_user_model
 from django.contrib.contenttypes.models import ContentType
 from django.db import transaction
 
@@ -18,7 +19,8 @@ from credentials.apps.credentials.utils import send_program_certificate_created_
 from credentials.apps.records.utils import send_updated_emails_for_program
 
 
-logger = logging.getLogger(__name__)
+log = logging.getLogger(__name__)
+User = get_user_model()
 
 
 class AbstractCredentialIssuer(metaclass=abc.ABCMeta):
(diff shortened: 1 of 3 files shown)


credentials/apps/credentials/issuers.py
```diff
@@ -123,6 +125,8 @@ def issue_credential(
         Returns:
             UserCredential
         """
+        log.info(f"Attempting to issue or update a program credential for learner '{username}' with status '{status}'")
```
staubina, line 128: Should we log user id rather than username?

justinhynes: I did consider this but ended up using username since it was passed into the function (rather than the entire user). I think it'd be more consistent if I did user the id though. I'll update this.

staubina: I need to dig it up, but I believe user_id is preferred in logging now.

credentials/apps/credentials/tests/test_issuer.py
```diff
@@ -32,99 +32,101 @@ class CertificateIssuerBase:
     def setUp(self):
         super().setUp()
         self.certificate = self.cert_factory.create()
-        self.username = "tester"
-        self.user = UserFactory(username=self.username)
-        self.user_cred = self.issuer.issue_credential(self.certificate, self.username)
+        self.user = UserFactory(username="tester")
```
justinhynes, line 35: Why all the changes?

I think we were trying to be too smart with some of the data used in the tests. We were trying to test `issue_credentials` but also trying to generate an instance of a UserCredential in the setup method to use in some tests. I've tried to simplify the tests by being more explicit about the UserCredential instances we are creating and testing against, rather than trying to re-use the ones generated in the setUp methods.

Also... I believe the main purpose of these tests are to test the `issue_credentials` method, so it probably shouldn't be called in a setup method...

staubina (approved):

staubina:
Take a quick look at whether or not we should use `get_user_model` vs. going directly to the User model.

justinhynes:
Good call on the user model piece. I'll make that update before merging.

PR #1074, opened 2021-04-05
fix: Protect records data in case of catalog issue

tuchfarber:
The copy catalog script now has a --delete option, which means
CASCADE-ing user grade deletions is dangerous. So to make it safer, this
now protects a CourseRun from being deleting if it's attached to a grade
, or a program from being deleted if it's attached to a program cert
record.


crice100 (commented):
Do we want to add any unit tests?

tuchfarber:
@crice100 Have no fear! Tests were added!

crice100 (approved):
Thanks!

PR #1076, opened 2021-04-05
feat: enable program search from user cred admin

tuchfarber:
Link ProgramCertificate to a Program instead of just storing the program_uuid. This enables an admin to search for a program title in the UserCredential admin page.


credentials/apps/credentials/admin.py
```diff
@@ -35,6 +35,25 @@ def certificate_uuid(self, obj):
...
         return obj.uuid.hex
 
+    def title(self, obj):
+        """Retrieves the title of either the course or program that the certificate was awarded for"""
+        if obj.credential_content_type.model_class() == ProgramCertificate:
+            program = obj.credential.program
+            if program:
+                return program.title
+            return ""
+        # TODO: return Course title once courses are linked to model
```
staubina, line 45: Thank you for this comment. Will this be simpler once the Course is linked as well?

credentials/apps/credentials/migrations/0019_populate_program_on_program_cert.py
```diff
@@ -0,0 +1,24 @@
...
+from django.db import migrations
+
+
+def populate_program(apps, schema_editor):
+    Program = apps.get_model("catalog", "Program")
+    ProgramCertificate = apps.get_model("credentials", "ProgramCertificate")
+    for program_cert in ProgramCertificate.objects.all():
+        try:
+            program = Program.objects.get(uuid=program_cert.program_uuid)
+        except Program.DoesNotExist:
```
staubina, line 12: What is the risk that we have a missing program?

credentials/apps/credentials/models.py
```diff
@@ -231,6 +232,9 @@ class ProgramCertificate(AbstractCertificate):
     """
 
     program_uuid = models.UUIDField(db_index=True, null=False, blank=False, verbose_name=_("Program UUID"))
+    # PROTECT prevents the Program from being delete if it's being used for a program cert. This allows copy_catalog
+    # to be safer when deleting
+    program = models.OneToOneField(Program, null=True, on_delete=models.PROTECT)
```
staubina, line 237: I have some concerns that this and program_uuid are two different fields. Will they be combined eventually?

staubina (commented):
Overall I like this approach, just some questions on how we are approaching the new field.

credentials/apps/credentials/admin.py
tuchfarber: There will just be an additional `if` block like the one above. Because we are using a `GenericRelation` and the values locations will be different depending on the related model, we have to check them separately.

credentials/apps/credentials/migrations/0019_populate_program_on_program_cert.py
tuchfarber: The risk after this PR is that the admin page might break. My plan is to see what prod looks like after this merges before I do anything else.

credentials/apps/credentials/models.py
tuchfarber: Yes! That's part of my cleanup stuff that's going to follow this ticket. `program_uuid` is used in a lot of places, so I wanted to break it out into more digestible PRs

credentials/apps/credentials/admin.py
staubina: Ok, thanks for the clarification.

staubina (approved):

PR #1078, opened 2021-04-06
feat: enable program search from user cred admin

tuchfarber:
Link ProgramCertificate to a Program instead of just storing the
program_uuid. This enables an admin to search for a program title in
the UserCredential admin page.

NOTE: This is the same PR as #1076  with one line change, which I've noted


credentials/apps/credentials/migrations/0019_populate_program_on_program_cert.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def populate_program(apps, schema_editor):
+    Program = apps.get_model("catalog", "Program")
+    ProgramCertificate = apps.get_model("credentials", "ProgramCertificate")
+    for program_cert in ProgramCertificate.objects.all():
+        try:
+            program = Program.objects.get(uuid=program_cert.program_uuid)
+        except Program.DoesNotExist:
+            continue
```
tuchfarber, line 13: Note: this was a `pass` causing program to be set to the previous program value in the `program_cert.program = program` line below. By using `continue` it'll skip over the assignment completely.

justinhynes (approved):

PR #1081, opened 2021-04-07
feat: enable courserun search from user cred admin

tuchfarber:
Link CourseCertificate to a CourseRun instead of just storing the
course_id (course key). This enables an admin to search for a course
title in the UserCredentials admin page


credentials/apps/credentials/migrations/0021_auto_20210407_1758.py
```diff
@@ -0,0 +1,24 @@
...
+from django.db import migrations
+
+
+def populate_course_run(apps, schema_editor):
+    CourseRun = apps.get_model("catalog", "CourseRun")
+    CourseCertificate = apps.get_model("credentials", "CourseCertificate")
+    for course_cert in CourseCertificate.objects.all():
+        try:
+            course_run = CourseRun.objects.get(key=course_cert.course_id)
+        except CourseRun.DoesNotExist:
```
crice100, line 12: Have we checked in prod to see if all course runs should exist? And/or should this log anything?

credentials/apps/credentials/models.py
```diff
@@ -199,6 +199,7 @@ class CourseCertificate(AbstractCertificate):
     """
 
     course_id = models.CharField(max_length=255, validators=[validate_course_key])
+    course_run = models.OneToOneField(CourseRun, null=True, on_delete=models.PROTECT)
```
crice100, line 202: Can't a course run have more than one certificate?

crice100: Oh, I'm thinking of user credentials aren't I? I always get that and course certificates mixed up.

crice100 (approved):
See my question in models.py, but I think I answered it myself. :-)

credentials/apps/credentials/migrations/0021_auto_20210407_1758.py
tuchfarber: My plan is to run this, check for null `course_run`s in stage/prod, then eventually make the the field non-nullable

credentials/apps/credentials/models.py
tuchfarber: Second comment is right! CourseCertificate is _naturally_ the course certificate configuration, not a course certificate. yay.

credentials/apps/api/v2/serializers.py
```diff
@@ -46,9 +46,20 @@ def to_internal_value(self, data):
...
+                try:
+                    course_run = CourseRun.objects.get(key=course_run_key)
+                except ObjectDoesNotExist:
+                    logger.exception(
+                        "Course run certificate failed to create "
+                        f"because the course run [{course_run_key}] doesn't exist in the catalog"
+                    )
+                    # TODO: course_run is still nullable while we validate the data. Re raise the exception once the
+                    # field is no longer nullable
+                    course_run = None
```
staubina, line 58: So this will cover the situation where copy fails to pull in a Course run before we get a cert? Should we alert on this?

staubina (approved):

credentials/apps/api/v2/serializers.py
tuchfarber: This is going to go away once I make these fields non-nullable. When that happens, I'll log the exception and we'll alert on it.

PR #1090, opened 2021-04-09
Bump credentials-themes version

edx-deployment:


oliviaruizknott (commented):
I know this is after the fact, but this was merged slightly prematurely. 😅 To follow-up: I looked into the breaking changes in the `docker[ssh]` upgrade:

```
Make Python 3.6 the minimum version supported
```

We are on 3.8.5. ✅

PR #1113, opened 2021-04-27
[feat] MB-1167 Certificate Availability API

Tj-Tracy:
This PR will create an API to expose the course certificate configuration so that the LMS can talk directly with Credentials and explicitly add the certificate availability date to a certificate configuration.

Previously, we were running into an issue where the LMS would send over a "visible date" that would be one of a few values; possibly the availability date, possibly the created date, modified date, etc. In order to reduce confusion and prevent bugs, we plan to add the availability date explicitly on any configuration that has one.

### DIFF
--- a/credentials/apps/api/v2/serializers.py
+++ b/credentials/apps/api/v2/serializers.py
@@ -257,3 +257,46 @@ class UserGradeSerializer(serializers.ModelSerializer):
             defaults=validated_data,
         )
         return grade
+
+
+class CourseCertificateSerializer(serializers.ModelSerializer):
+    course_run = CourseRunField(read_only=True)
+
+    class Meta:
+        model = CourseCertificate
+        fields = (
+            "id",
+            "course_id",
+            "course_run",
+            "certificate_type",
+            "certificate_available_date",
+            "is_active",
+        )
+        read_only_fields = ("id", "course_run")
+
+    def create(self, validated_data):
+        site = self.context["request"].site
+        # A course run may not exist, but if it does, we want to find it. There may be times where course
+        # staff will change the date on a newly created course before credentials has pulled it from the catalog.
+        course_id = validated_data["course_id"]
+        course_run = None
+        try:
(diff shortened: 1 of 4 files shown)


oliviaruizknott (approved):
I’m a little fuzzy on how exactly Django serializers work and haven’t dug into that to review this; but if you’re confident about the specific implementation here, the general approach looks good to me!

tuchfarber (commented):
Small request to tighten up the API.

Regarding updating/creating: If we only update,  we'll have to make sure the `CourseCertificate` for that course exists before we call this API to update it. The `CourseCertificate` is only created when a certificate for that course run is awarded though. So if the course team changes the date before anyone has earned a certificate, we'd need to create instead of update. 

We'll need to decide where to handle this "create or update" logic. If we have all the data during  the call to this API, we could do a similar thing that the `UserCredential` serializer does when creating a credential for a course that doesn't exist yet. 

Sorry for not catching this when I drew out the steps. I forgot the `UserCredential` API created the `CourseCertificate` (as opposed to a PC creating it) until you brought it up today.

credentials/apps/api/v2/serializers.py
```diff
@@ -257,3 +257,44 @@ def create(self, validated_data):
...
+    def create(self, validated_data):
+        site = self.context["request"].site
+        # A course run may not exist, but if it does, we want to find it. There may be times where course
+        # staff will change the date on a newly created course before credentials has pulled it from the catalog.
+        course_run = None
+        course_run_id = validated_data["course_run"]
+        try:
+            course_run = CourseRun.objects.get(key=course_run_id)
+        except ObjectDoesNotExist:
+            logger.exception(
```
crice100, line 284: If this is an expected scenario, should we maybe log it with a lower log level?

crice100 (commented):
Could we add unit tests?

crice100 (approved):
Could you rebase on master (or merge master in) before merging this? Just want to make sure everything looks ok with the latest version of black (our formatter).

PR #1121, opened 2021-04-30
feat: add toggle for certificate_available_date

oliviaruizknott:
This work adds a temporary WaffleSwitch to gate the use of the
certificate_available_date field on the CourseCertificate model while
we develop features around it.

This ONLY adds the switch, but doesn’t use it yet.

To use the toggle:

  1. In the waffle section of the django admin
     (< credentials host >admin/waffle/), add a new Switch with the name:
     `credentials.use_certificate_available_date`. Make sure it is
     "active".

  2. In the code, call the switch like this:
     ```
     from credentials.settings.base import USE_CERTIFICATE_AVAILABLE_DATE
     if USE_CERTIFICATE_AVAILABLE_DATE.is_enabled():
        # do your stuff here
     ```

Original ticket: https://openedx.atlassian.net/browse/MICROBA-1169
Ticket to remove flag: https://openedx.atlassian.net/browse/MICROBA-1198

### DIFF
--- a/requirements/all.txt
+++ b/requirements/all.txt
@@ -37,11 +37,11 @@ bleach==3.3.0
     #   -r requirements/production.txt
 bok-choy==1.1.1
     # via -r requirements/dev.txt
-boto3==1.17.60
+boto3==1.17.61
     # via
     #   -r requirements/production.txt
     #   django-ses
-botocore==1.20.60
+botocore==1.20.61
     # via
     #   -r requirements/production.txt
     #   boto3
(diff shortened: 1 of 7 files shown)


requirements/all.txt
```diff
@@ -37,11 +37,11 @@ bleach==3.3.0
     #   -r requirements/production.txt
 bok-choy==1.1.1
     # via -r requirements/dev.txt
-boto3==1.17.60
+boto3==1.17.61
```
oliviaruizknott, line 40: Wups. Some other tiny upgrades might’ve slipped in with my `make upgrade`.

staubina (approved):

Tj-Tracy (approved):

PR #1143, opened 2021-05-13
fix: fix failing unit tests after `django-webpack-loader` upgrade

justinhynes:
- Update requirements via `make upgrade`, including `django-webpack-loader` at V1.x.
- Upgrade `webpack-bundle-tracker` from `0.4.3` to `1.0.0` as mentioned here: https://github.com/django-webpack/django-webpack-loader/issues/272. 

Upgrading both packages and running `make static` afterwards locally seems to resolve the failing Python unit tests.


justinhynes:
While the unit tests are now passing with these changes, I am getting error rendering Django templates locally. I am closing this PR as I am not planning to upgrade these packages at this time.

PR #1148, opened 2021-05-18
feat: use certificate_available_date field for filters

oliviaruizknott:
This commit updates / adds some utility methods to use our new
certificate_available_date field on CourseCertificate (instead of the
visible_date on UserCredentialAttribute) for any methods that need
to filter certs to show only those that should be visible (i.e., whose
certificate_availbable_date is not in the future).

The filter_visible method (altered here) is used in the following
places:

  - get_record_data: a method called by ProgramRecordView and
    ProgramRecordCsvView. It provides all the record data for a given
    user and program--but only the visible course and program
    certificates

  - RecordsListBaseView: gets all the credentials for a given user,
    sorted by program or course certificate type—-but only wants
    the visible ones.

  - handle_only_visible: through the UserCredentialFilter. This allows
    callers of the credentials api to filter by this visibility.
		(This is used by edx-platform.)

This work is gated behind the USE_CERTIFICATE_AVAILABLE_DATE toggle
while development is in progress.

In partial fulfillment of MICROBA-1170 and MICROBA-1171.

### DIFF
--- a/credentials/apps/credentials/utils.py
+++ b/credentials/apps/credentials/utils.py
@@ -11,8 +11,6 @@ from credentials.apps.catalog.data import ProgramStatus
 from credentials.apps.core.models import User
 from credentials.apps.credentials.messages import ProgramCertificateIssuedMessage
 from credentials.apps.credentials.models import (
-    CourseCertificate,
-    ProgramCertificate,
     ProgramCompletionEmailConfiguration,
     UserCredential,
     UserCredentialAttribute,
@@ -53,6 +51,8 @@ def validate_duplicate_attributes(attributes):
     return True
 
 
+# TODO: Delete this when removing USE_CERTIFICATE_AVAILABLE_DATE toggle:
+# MICROBA-1198
 def datetime_from_visible_date(date):
     """Turn a string version of a datetime, provided to us by the LMS in a particular format it uses for
     visible_date attributes, and turn it into a datetime object."""
@@ -66,8 +66,21 @@ def datetime_from_visible_date(date):
         return None
 
 
+# TODO: Refactor this when removing USE_CERTIFICATE_AVAILABLE_DATE toggle:
+# MICROBA-1198
 def filter_visible(qs):
-    """Filters a UserCredentials queryset by excluding credentials that aren't supposed to be visible yet."""
+    """
+    Filters a UserCredentials queryset by excluding credentials that aren't
+    supposed to be visible yet.
+    """
+
+    if settings.USE_CERTIFICATE_AVAILABLE_DATE.is_enabled():
+        visible_course_certs = _filter_visible_course_certificates(qs.filter(course_credentials__isnull=False))
+        visible_program_certs = _filter_visible_program_certificates(qs.filter(program_credentials__isnull=False))
+        visible_certs = visible_course_certs | visible_program_certs
+
+        return visible_certs
+
     # The visible_date attribute holds a string value, not a datetime one. But we can compare as a string
     # because the format is so strict - it will still lexically compare as less/greater-than.
     nowstr = datetime.datetime.now(datetime.timezone.utc).strftime(VISIBLE_DATE_FORMAT)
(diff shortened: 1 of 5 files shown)


credentials/apps/credentials/utils.py
```diff
@@ -53,6 +52,8 @@ def keyfunc(attribute):
     return True
 
 
+# TODO: Delete this when removing USE_CERTIFICATE_AVAILABLE_DATE toggle:
```
justinhynes, line 55: Quick question: Having this as a TODO is great, is this mentioned in the ticket (MICROBA-1198) anywhere? Just to make sure that this TODO item isn't somehow missed when the cleanup actually happens?

oliviaruizknott: Good point. Will put a note in 1198.

credentials/apps/records/views.py
```diff
@@ -140,7 +140,7 @@ def get_record_data(user, program_uuid, site, platform_name=None):
         "name": program.title,
         "type": slugify(program.type),
         "type_name": program.type,
-        "completed": program_credential_query.exists(),
+        "completed": bool(program_credential_query),
```
justinhynes, line 143: Was there any particular reason for this change? This made me lookup the difference between `bool()` and `.exists()` for QuerySet's. As I was reading, this line came up in the Django documentation:
```Note: If you only want to determine if at least one result exists (and don’t need the actual objects), it’s more efficient to use exists().```
https://docs.djangoproject.com/en/dev/ref/models/querysets/

Comment (oliviaruizknott) [reply credentials/apps/records/views.py:143]: Yes! This is a notable change. This is related to the note in the commit message:
> Because of the logic required to calculate program cert available
date, this change makes filter_visible return a list instead of a queryset.

You can only call `.exists()` on a QuerySet. It’s a great and efficient thing for us to use when `filter_visible` is returning a QuerySet (as it is currently when the feature flag is off). However, our new filter methods return plain old python lists, which don’t know about the `.exists()` method. `bool()` will return `False` for both an empty QuerySet and an empty list, and `True` if there is at least one thing in either. That’s all we want for this bit of data.

This is technically less efficient, as it means that now the query will actually be executed where it wasn’t being executed before (she says with 90% confidence). But since there should only ever be a maximum of 1 record in the query, I think it’ll be alright.

Comment (justinhynes) [reply credentials/apps/records/views.py:143]: Ahh! Thanks. I didn't make that connection. Appreciate the break down!

File: credentials/apps/credentials/utils.py [hunk 3]
```diff
@@ -66,8 +67,30 @@ def datetime_from_visible_date(date):
...
+
+
+def program_certificate_type():
+    return ContentType.objects.get(app_label="credentials", model="programcertificate")
+
+
+# TODO: Refactor this when removing USE_CERTIFICATE_AVAILABLE_DATE toggle:
+# MICROBA-1198
 def filter_visible(qs):
     """Filters a UserCredentials queryset by excluding credentials that aren't supposed to be visible yet."""
```
justinhynes, line 81: We may want to add some additional doc here to indicate, for some amount of time, this can return a QuerySet OR a List depending on configuration.

oliviaruizknott: 👍 Will add with the tests.

staubina (commented):
I have added some suggestions. I started to note that some of these methods could potentially be moved to another file, but I now realize that some of them like `filter_visible` predate this work. 

Overall you are heading in the right direction.

tuchfarber (commented):
One small request. 

Generally, I think I'm disliking more and more our whole "Certs are different but the same approach". Us having to have branches in every function to handle program certs and course certs separately is starting to get frustrating. (This comment doesn't matter for this review, but I think it's something we should fix in the future)

credentials/apps/credentials/utils.py
```diff
@@ -66,8 +67,30 @@ def datetime_from_visible_date(date):
         return None
 
 
+def course_certificate_type():
```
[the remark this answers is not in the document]
oliviaruizknott: I went Matt’s route on this one and made these constants.

credentials/apps/credentials/utils.py
```diff
@@ -76,6 +99,46 @@ def filter_visible(qs):
     )
 
 
+def filter_visible_course_certificates(qs):
+    """
+    Filters a UserCredentials queryset by excluding credentials that aren’t
+    supposed to be visible yet according to their certificate_available_date.
+
+    Arguments:
+        qs (UserCredential QuerySet): A queryset of UserCredential objects of
```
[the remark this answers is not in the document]
oliviaruizknott: I went the `_` route to at least signal to developers to use this with intention. Write now, this will error and break if the wrong this is passed. What should happen instead might be use-case-specific.

oliviaruizknott:
@tuchfarber, I agree about the frustration of handling course certs and program certs differently now. While I think the whole UserCredentialAttribute thing is unnecessary and clunky, we’ve lost some elegance on this front. 😕

justinhynes (approved):

tuchfarber (approved):
Don't forget to squash!

staubina (approved):

credentials/apps/records/views.py
oliviaruizknott: Aaaaaand in the end, I’m changing this back to a QuerySet because some django filtering that the credentials api uses expects that.

credentials/apps/credentials/utils.py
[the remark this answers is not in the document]
oliviaruizknott: The constant route ended up not working; it broke all the tests because it looks like the constants were getting defined before the database was ready. Changed to using filters in the methods where this method had been used, and got rid of this.

credentials/apps/credentials/utils.py
```diff
@@ -102,15 +99,16 @@ def _filter_visible_course_certificates(query_set):
...
     now = datetime.datetime.now(datetime.timezone.utc)
-    return [
-        user_credential
+    visible_course_cert_ids = [
+        user_credential.id
         for user_credential in query_set
         if not user_credential.credential.certificate_available_date
         or user_credential.credential.certificate_available_date <= now
     ]
+    return UserCredential.objects.filter(pk__in=visible_course_cert_ids)
```
tuchfarber, line 111: Since we're doing querysets now, you can just filter the original queryset, rather than traversing it and doing another query off the IDs. E.g. (untested)
```
    return query_set.filter(
        Q(credential__certificate_available_date__isnull=True)
        | Q(credential__certificate_available_date__lte=now)
    )
```

tuchfarber (approved):

PR #1155, opened 2021-06-02
[feat] Credentials masters CourseCertificate support

Tj-Tracy:
While trying to update the available_date for CourseCertificates in
credentials, we were getting errors that master's program course
certificates were not being properly updated. this add's Masters to the
list of valid CourseCertificate configurations.


justinhynes (approved):
LGTM

PR #1184, opened 2021-06-25
fix: update program credential visible date logic

oliviaruizknott:
Addresses MICROBA-1303, which was made from CR-3802. A learner’s program
certificate was showing an issue date earlier than expected, earlier
than the date when they completed all the required courses.

This was because of some faulty logic in our
`_get_program_certificate_visible_date` method. It was only returning
the latest associated `certificate_available_date`. But
`certificate_available_dates` are only set on courses that are
instructor-paced and have a certificate availability date manually set.
In the case of our learner, only one course in the program had a
`certificate_available_date`--and it was one of the earlier courses
completed by the learner. The rest of the courses did not have a
`certificate_available_date` because they were self-paced. Since the
logic only returned the latest `certificate_available_date`, and because
the later courses didn’t have that particular field, the program visible
date was being returned as a date that was too early.

Now, we have updated the `_get_program_certificate_visible_date` logic
to use the course credential `created` date when there is no
`certificate_available_date`. This will ensure that the program
credential displays the date of the latest course completed, whether the
course was instructor- or self-paced.

Note: we have some questions about whether it is more appropriate to
use the course credential’s `modified` date rather than the `created`
date, but are sticking with `created` for now because that is closer to
the original design.

### DIFF
--- a/credentials/apps/credentials/utils.py
+++ b/credentials/apps/credentials/utils.py
@@ -114,8 +114,8 @@ def _filter_visible_program_certificates(query_set):
     supposed to be visible yet according to their certificate_available_date.
 
     Arguments:
-        query_set (UserCredential QuerySet): A queryset of UserCredential objects of
-        the ProgramCertificate ContentType.
+        query_set (UserCredential QuerySet): A queryset of UserCredential
+        objects of the ProgramCertificate ContentType.
 
     Returns:
         (QuerySet): A queryset of program UserCredentials that should be visible.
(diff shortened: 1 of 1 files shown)


credentials/apps/credentials/utils.py
```diff
@@ -114,8 +114,8 @@ def _filter_visible_program_certificates(query_set):
     supposed to be visible yet according to their certificate_available_date.
 
     Arguments:
-        query_set (UserCredential QuerySet): A queryset of UserCredential objects of
-        the ProgramCertificate ContentType.
+        query_set (UserCredential QuerySet): A queryset of UserCredential
+        objects of the ProgramCertificate ContentType.
```
oliviaruizknott, line 119: (This was just a whitespace change.)

credentials/apps/credentials/utils.py
```diff
@@ -124,23 +124,24 @@ def _filter_visible_program_certificates(query_set):
     visible_program_cert_ids = []
     for user_credential in query_set:
         program_visible_date = _get_program_certificate_visible_date(user_credential)
-        if not program_visible_date or program_visible_date <= now:
```
oliviaruizknott, line 127: Previously, what was returned here as `program_visible_date` was really the latest `certificate_available_date` for an associated course. It was possible that it could be `None` (if the program was full of entirely self-paced courses). We were happy to display programs with no associated `certificate_available_date`s, or else programs where the associated `certificate_available_date`s were in the past.

But now that we’ve updated the logic of `_get_program_certificate_visible_date`, it should not return `None`. If it does return `None`, something else is wrong, and proceeding would cause an error.

credentials/apps/credentials/utils.py
```diff
@@ -170,13 +170,15 @@ def get_credential_visible_dates(user_credentials):
...
             if user_credential.course_credentials.exists():
-                date = user_credential.credential.certificate_available_date
+                date = user_credential.credential.certificate_available_date or user_credential.created
 
+            # If this is a program credential
             if user_credential.program_credentials.exists():
                 date = _get_program_certificate_visible_date(user_credential)
 
-            visible_date_dict[user_credential] = date or user_credential.created
+            visible_date_dict[user_credential] = date
```
oliviaruizknott, line 181: Previously, it was possible that `date` was `None` at this point. Now, `date` will always be a date at this point.

(Again, caveat that it can _technically_ be `None`, but unlikely logically.)

Tj-Tracy (approved):

staubina (approved):
These changes look good to me.

PR #1189, opened 2021-07-01
feat: Add a program cert preview for admin users

staubina:


credentials/apps/credentials/views.py
```diff
@@ -143,6 +153,86 @@ def get_child_templates(self):
...
+    This will render the template for a Programs UUID. This is currently limited to Program Credential types
+    """
+
+    template_name = "credentials/base.html"
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+
+        # Only staff users should have access to this template.
+        if not self.request.user.is_authenticated and not self.request.user.is_staff:
```
crice100, line 168: Should the _and_ be an _or_?

staubina: I actually am thinking it should be 

`if not self.request.user.is_staff:
    raise Http401`
And return a 401 and not a 404.

staubina: I moved to user a decorator that will require the user to be a Staff user.

crice100 (approved):

PR #1228, opened 2021-07-30
mscandinaro/MICROBA-1414 feat: added CORS and CSRF support

mlscandinaro:


tuchfarber (approved):

credentials/settings/devstack.py
```diff
@@ -66,6 +66,8 @@
     "http://edx.devstack.lms:18000/oauth2",
 )
 
+CORS_ORIGIN_WHITELIST = ("http://localhost:8080",)
```
justinhynes, line 69: I can't remember if we have yet, but if we haven't, can you add a task to your current story to update the default port this MFE is running on (`8080`) to a new value? We will have to come and update this setting in `devstack.py` as well.

Thanks!

justinhynes (approved):

PR #1242, opened 2021-08-06
chore: webpack 5 upgrade

tuchfarber:


credentials/static/components/SendLearnerRecordModal.jsx
```diff
@@ -1,4 +1,4 @@
-import '@babel/polyfill'; // Needed to support Promises on legacy browsers
```
tuchfarber, line 1: Per https://babeljs.io/docs/en/babel-polyfill
>>> As of Babel 7.4.0, this package has been deprecated in favor of directly including core-js/stable (to polyfill ECMAScript features) and regenerator-runtime/runtime (needed to use transpiled generator functions)

We don't need `regenerator-runtime` since we're only adding this for Promises. By *only* polyfilling promises, we're also shaving 200kb off the bundle size :)

webpack.config.js
```diff
@@ -9,7 +9,7 @@ const isDevstack = (process.env.DJANGO_SETTINGS_MODULE === 'credentials.settings
 function getPlugins() {
   const plugins = [];
   plugins.push(new BundleTracker({ filename: './webpack-stats.json' }));
-  plugins.push(new MiniCssExtractPlugin({ filename: '[name]-[hash].css' }));
+  plugins.push(new MiniCssExtractPlugin({ filename: '[name]-[contenthash].css' }));
```
tuchfarber, line 12: v5 upgrade doc says 
> When using [hash] placeholder in webpack configuration, consider changing it to [contenthash]. It is not the same, but proven to be more effective.
seems to work fine, so might as well get the benefits.

webpack.config.js
```diff
@@ -29,8 +29,9 @@ module.exports = {
   },
   output: {
     path: path.resolve('./credentials/static/bundles/'),
-    filename: '[name]-[hash].js',
+    filename: '[name]-[contenthash].js',
     libraryTarget: 'window',
+    publicPath: '/static/bundles/'
```
tuchfarber, line 34: This was the thing that broke everything. The `webpack-stats.json` output changed and it was trying to resolve all bundles at a relative path `/records/program/<program>/<bundle>` instead of `/static/bundles/<bundle>`

credentials/static/components/SendLearnerRecordModal.jsx
tuchfarber: The polyfill was also spitting out bad code with runtime errors and this fixed it.

Tj-Tracy (approved):

PR #1243, opened 2021-08-06
chore: Manual python upgrade and moved social-auth-app-django to common

staubina:
Manual python upgrade and moved social-auth-app-django to common constraint file


crice100 (approved):
I know this is an automatic "ran out of space" thing, but _file_ is a great PR description. :-)

PR #1257, opened 2021-08-10
feat: use verified name on program certificate

alangsto:
[MST-895](https://openedx.atlassian.net/browse/MST-895)

If user has chosen to use their verified name on certificates, use the verified name when rendering a program certificate.

### DIFF
--- a/credentials/apps/core/models.py
+++ b/credentials/apps/core/models.py
@@ -133,6 +133,10 @@ class SiteConfiguration(models.Model):
         return "{}/api/user/v1/".format(self.lms_url_root.strip("/"))
 
     @property
+    def name_verification_api_url(self):
+        return "{}/api/edx_name_affirmation/v1/verified_name".format(self.lms_url_root.strip("/"))
+
+    @property
     def api_client(self):
         """
         Returns a requests client for this site's service user.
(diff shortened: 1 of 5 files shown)


bseverino (approved):

credentials/apps/core/models.py
```diff
@@ -165,10 +170,28 @@ def get_user_api_data(self, username):
...
+                verification_data = verification_response.json()
+
+                # add relevant verified name data to user_data
+                user_data["verified_name"] = verification_data.get("verified_name")
+                user_data["use_verified_name_for_certs"] = verification_data.get("use_verified_name_for_certs")
+            except HTTPError as e:
+                # 404 response means that verified name does not exist for this user. We don't want to throw
+                # an error if that's the case, and instead we should continue.
+                if e.response.status_code != 404:
+                    raise e
```
tuchfarber, line 193: I don't think I want all certs to fail if there's a bug in the name affirmation API. Can we either:
1. Fall back to using the normal name if there's an API issue, or
2. Put this behind a setting (we can default it on) that we can remote-config off if there's an issue.

tuchfarber (changes requested):

credentials/apps/core/models.py
alangsto: @tuchfarber, thanks for calling that out! I've updated the code so that we fall back to using the profile name if there is an API issue.

tuchfarber (approved):
Thanks for the fix!

PR #1264, opened 2021-08-11
feat: Add dockerfile and push image to Dockerhub

tuchfarber:
This image builds on our existing (non-ansible) Dockerfiles with the
following exceptions:
- It's the first to require node to build static assets. I've followed
the pattern we use in the ansible builds where we use nodeenv and I've
also taken the node and npm versions from ansible. This pattern may
change in the future but for now I'm prioritizing consistency with the
existing images.
- This is also the first one being used in devstack, so to utilize
existing devstack code, the folder structure has to have an additional
layer. The repo folder will live at /edx/app/credentials/credentials
instead of the /edx/app/<service> that other's use.

I tested this by building the image locally and pointing my devstack at it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,97 @@
+FROM ubuntu:focal as app
+MAINTAINER devops@edx.org
+
+
+# Packages installed:
+# git; Used to pull in particular requirements from github rather than pypi,
+# and to check the sha of the code checkout.
+
+# language-pack-en locales; ubuntu locale support so that system utilities have a consistent
+# language and time zone.
+
+# python; ubuntu doesnt ship with python, so this is the python we will use to run the application
+
+# python3-pip; install pip to install application requirements.txt files
+
+# libssl-dev; # mysqlclient wont install without this.
+
+# libmysqlclient-dev; to install header files needed to use native C implementation for
+# MySQL-python for performance gains.
+
+# wget to download a watchman binary archive
+
+# unzip to unzip a watchman binary archive
+
+# If you add a package here please include a comment above describing what it is used for
+RUN apt-get update && \
+apt-get install -y software-properties-common && \
+apt-add-repository -y ppa:deadsnakes/ppa && apt-get update && \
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,97 @@
...
+ENV VIRTUAL_ENV=/edx/app/credentials/venvs/credentials
+RUN python3.8 -m venv $VIRTUAL_ENV
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
+
+# Create Node env
+RUN pip install nodeenv
+ENV NODE_ENV=/edx/app/credentials/nodeenvs/credentials
+RUN nodeenv $NODE_ENV --node=12.11.1 --prebuilt
+ENV PATH="$NODE_ENV/bin:$PATH"
+RUN npm install -g npm@6.12.1
```
tuchfarber, line 43: This section is new. Mostly pulled from ansible.

Dockerfile
```diff
@@ -0,0 +1,97 @@
...
+RUN wget https://github.com/facebook/watchman/releases/download/v2020.08.17.00/watchman-v2020.08.17.00-linux.zip
+RUN unzip watchman-v2020.08.17.00-linux.zip
+RUN mkdir -p /usr/local/{bin,lib} /usr/local/var/run/watchman
+RUN cp watchman-v2020.08.17.00-linux/bin/* /usr/local/bin
+RUN cp watchman-v2020.08.17.00-linux/lib/* /usr/local/lib
+RUN chmod 755 /usr/local/bin/watchman
+RUN chmod 2777 /usr/local/var/run/watchman
+
+# Now install credentials
+WORKDIR /edx/app/credentials/credentials
```
tuchfarber, line 64: This and below is where I had to add the extra folder layer.

christopappas (approved):

tuchfarber:
Merging despite codecov/patch tests because the gunicorn file does not require testing.

PR #1279, opened 2021-08-19
feat: Store date_override sent from platform

oliviaruizknott:
A course certificate date_override may be sent from edx-platform. This
commit handles that data on the credentials side. The date_override key
will be sent over with every post for a course certificate. If the key
has a value, save a UserCredentialDateOverride record associated with
the given UserCredential. If the date_override value is empty, delete
any UserCredentialDateOverride that might be associated with the given
UserCredential.

Toward [MICROBA-1455](https://openedx.atlassian.net/browse/MICROBA-1455)


credentials/apps/api/v2/serializers.py
```diff
@@ -167,6 +179,7 @@ class UserCredentialCreationSerializer(serializers.ModelSerializer):
 
     credential = CredentialField()
     attributes = UserCredentialAttributeSerializer(many=True, required=False)
+    date_override = UserCredentialDateOverrideSerializer(required=False, allow_null=True)
```
oliviaruizknott, line 182: This was frustrating. I believe that `required=False, allow_null=True` should allow for `date_override` to be empty, but the serializer kept throwing an error:
```python
{'date_override': {'date': [ErrorDetail(string='This field may not be null.', code='null')]}}
```

I eventually fixed it by explicitly declaring the serializer field above (line 139), but it makes things a little redundant, and makes inheriting from `ModelSerializer` (instead of just `Serializer`) kind of pointless. 

Resources:
- https://www.django-rest-framework.org/api-guide/serializers/#dealing-with-nested-objects
- https://stackoverflow.com/questions/29606115/django-rest-framework-nested-serializer-required-false-error

oliviaruizknott: Woohoo Matt helped me figure out what was wrong here! 🙌

credentials/apps/api/v2/tests/test_views.py
```diff
@@ -168,6 +170,46 @@ def test_create_with_existing_user_credential(self):
...
+        self.assertEqual(response.data, self.serialize_user_credential(user_credential))
+
+        self.assertEqual(user_credential.date_override.date.strftime("%Y-%m-%d"), expected_date_override)
+
+        # Verify that the date_override is removed if not present in the post
+        data["date_override"] = None
+        response = self.client.post(self.list_path, data=json.dumps(data), content_type=JSON_CONTENT_TYPE)
+        user_credential.refresh_from_db()
+        with self.assertRaises(ObjectDoesNotExist):
+            print(user_credential.date_override)
```
oliviaruizknott, line 211: This `print` feels a bit silly to me but had to do something with `user_credential.date_override` (the thing that actually triggers the exception) or else `pylint` failed with `W0104: Statement seems to have no effect (pointless-statement)`

tuchfarber (approved):

PR #1284, opened 2021-08-20
feat: Use date override on Program Record

oliviaruizknott:
When displaying the Program Record, use the date override (if present)
for the date value in the "Date Earned" column. This will make the
Program Record match the date displayed on the learner’s course
certificate.

Screen Shot 2021-08-20 at 3 16 59 PM

# NOTE
**This will only display properly if the `USE_CERTIFICATE_AVAILABLE_DATE` waffle switch is enabled.** No, these changes are not related to that feature, but take my choice to leave this out of the old code paths as a bet that we’ll move forward and clean up that old code.

### DIFF
--- a/credentials/apps/credentials/utils.py
+++ b/credentials/apps/credentials/utils.py
@@ -4,6 +4,7 @@ import textwrap
 from itertools import groupby
 
 from django.conf import settings
+from django.core.exceptions import ObjectDoesNotExist
 from django.db.models import Q
 from edx_ace import Recipient, ace
 
@@ -49,6 +50,11 @@ def validate_duplicate_attributes(attributes):
         if len(list(group)) > 1:
             return False
     return True
+
+
+def to_datetime(date):
+    """Turn a date object into a datetime object by setting time to 00:00"""
+    return datetime.datetime.combine(date, datetime.datetime.min.time())
 
 
 # TODO: Delete this when removing USE_CERTIFICATE_AVAILABLE_DATE toggle:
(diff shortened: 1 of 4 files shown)


credentials/apps/credentials/utils.py
```diff
@@ -179,6 +188,13 @@ def get_credential_visible_dates(user_credentials):
             if user_credential.course_credentials.exists():
                 date = user_credential.credential.certificate_available_date or user_credential.created
 
+                if use_date_override:
+                    try:
+                        date_override = user_credential.date_override.date
+                        date = to_datetime(date_override)
```
oliviaruizknott, line 194: If this is passed to the front as a date instead of a datetime, it gets messed up with timezones. I made the choice to store these date overrides in the database as `Date`s rather than `DateTime`s. This is making me question that a little bit, but unless someone feels very strongly, I’m going to move forward with this solution instead of changing the database fields (here in credentials and also in edx-platform, for consistency).

crice100: I think this is fine for now, but we probably want to come back and improve it.

credentials/apps/records/views.py
```diff
@@ -183,7 +183,8 @@ def get_record_data(user, program_uuid, site, platform_name=None):
 
         # If the user has taken the course, show the course_run info for the highest grade
         elif grade is not None and grade.course_run == course_run:
-            issue_date = visible_dates[user_credential_dict[course_run.key]]
+            user_credential = user_credential_dict.get(course_run.key)
+            issue_date = get_credential_visible_date(user_credential, use_date_override=True)
```
crice100, line 187: Should we be reading use_date_override from a flag? How do we decide if it's true/false?

credentials/apps/credentials/utils.py
```diff
@@ -51,6 +52,11 @@ def keyfunc(attribute):
     return True
 
 
+def to_datetime(date):
+    """Turn a date object into a datetime object"""
+    return datetime.datetime.combine(date, datetime.datetime.min.time())
```
crice100, line 57: Is this trying to set the time to midnight UTC? If so, a comment would be great.

credentials/apps/credentials/utils.py
```diff
@@ -160,7 +166,7 @@ def _get_program_certificate_visible_date(user_program_credential):
 
 # TODO: Refactor this when removing USE_CERTIFICATE_AVAILABLE_DATE toggle:
 # MICROBA-1198
-def get_credential_visible_dates(user_credentials):
+def get_credential_visible_dates(user_credentials, use_date_override=False):
```
crice100, line 169: Suggestion: add a comment saying why/when we might want to set _use_date_override_ to either True or False.

credentials/apps/records/views.py
oliviaruizknott: For posterity: TLDR: No, this `use_date_override` is not coming from a flag.

That `get_credential_visible_date` util function (and its friend `get_credential_visible_dates`) is (are) called several places in the code, for various uses.

Per Ryan’s current decisions about how we want the (course) certificate date override to be used in / relate to credentials, this `issue_date` line is the only place where `get_credential_visible_date` is called that we want it to care about the date override.

That may change in future tickets, in which case more places that call `get_credential_visible_date` can call it with `use_date_override` set to `True`.

credentials/apps/credentials/utils.py
oliviaruizknott: Will make a ticket for changing the database tables (and relevant code).

credentials/apps/credentials/utils.py
```diff
@@ -160,25 +166,42 @@ def _get_program_certificate_visible_date(user_program_credential):
...
+                <UserCredential>: None,
+                ...
+            }
     """
 
     if settings.USE_CERTIFICATE_AVAILABLE_DATE.is_enabled():
         visible_date_dict = {}
-        date = None
         for user_credential in user_credentials:
+            date = None
```
oliviaruizknott, line 192: @staubina Note this change, and the change in comment above.

PR #1301, opened 2021-08-31
fix: Correct issue_date to match date sent from the LMS

staubina:


credentials/apps/credentials/utils.py
```diff
@@ -188,19 +188,29 @@ def get_credential_visible_dates(user_credentials, use_date_override=False):
 
     if settings.USE_CERTIFICATE_AVAILABLE_DATE.is_enabled():
         visible_date_dict = {}
-        for user_credential in user_credentials:
-            date = None
+        date = None
```
oliviaruizknott, line 191: Is there a particular reason you want to move this out of the loop below? I think this might introduce other bugs.

credentials/apps/credentials/utils.py
```diff
@@ -188,19 +188,29 @@ def get_credential_visible_dates(user_credentials, use_date_override=False):
...
             if user_credential.course_credentials.exists():
-                date = user_credential.credential.certificate_available_date or user_credential.created
-
                 if use_date_override:
                     try:
                         date_override = user_credential.date_override.date
                         date = to_datetime(date_override)
                     except ObjectDoesNotExist:
                         pass
+                elif user_credential.credential.certificate_available_date:
```
oliviaruizknott, line 202: (Wups, this is meant for line 201 above but I can’t comment there.)

I think the date override check needs to be last. Otherwise, if `use_date_override` is truthy but the credential doesn’t have a `date_override`, it will just `pass` and never enter the `elif` and `else` logic; it will just use whatever `date` is currently set to. (And if `date` is outside/above the `user_credential` loop on line 193, it will just use whatever the last `date` in the loop was instead of `None`.)

credentials/apps/credentials/utils.py
staubina: I did not mean to move this 🤦. Bad rebase. I will fix it.

staubina: This should be fixed now.

credentials/apps/credentials/utils.py
staubina: I corrected this and tested it.

credentials/apps/credentials/utils.py
```diff
@@ -188,12 +188,21 @@ def get_credential_visible_dates(user_credentials, use_date_override=False):
...
+                if user_credential.credential.certificate_available_date:
+                    date = user_credential.credential.certificate_available_date
+                else:
+                    try:
+                        visible_date = UserCredentialAttribute.objects.prefetch_related(
+                            "user_credential__credential"
+                        ).get(user_credential=user_credential, name="visible_date")
+                        date = datetime_from_visible_date(visible_date.value)
+                    except UserCredentialAttribute.DoesNotExist:
+                        date = user_credential.created
```
justinhynes, line 205: I am still shakey on all the certificate date stuff the team worked on while I was focusing on the other certificate project.
 
Is there any crucial information we are swallowing here we would want to log? Or would logging something here be spammy and less useful? Is it more likely the UserCredentialAttribute doesn't exist?

staubina: I don't think it would hurt to log it, I believe this not existing would be a surprise.

credentials/apps/records/tests/test_views.py
```diff
@@ -667,15 +667,39 @@ def test_program_visible_date(self, date, completed):
     # USE_CERTIFICATE_AVAILABLE_DATE waffle switch enabled. Clean up previous
     # tests in MICROBA-1198.
```
oliviaruizknott, line 668: Hmm. This comment is wrong now. Could you maybe move your new tests below the `test_program_visible_date_with_certificate_available_date` test, or else clean up (remove?) the comment? Corresponding comment on 619, too.

oliviaruizknott (approved):
A question and a nit about comments, but general approval! 😄

credentials/apps/credentials/utils.py
```diff
@@ -188,12 +188,21 @@ def get_credential_visible_dates(user_credentials, use_date_override=False):
...
-                date = user_credential.credential.certificate_available_date or user_credential.created
+                if user_credential.credential.certificate_available_date:
+                    date = user_credential.credential.certificate_available_date
+                else:
+                    try:
+                        visible_date = UserCredentialAttribute.objects.prefetch_related(
+                            "user_credential__credential"
+                        ).get(user_credential=user_credential, name="visible_date")
+                        date = datetime_from_visible_date(visible_date.value)
+                    except UserCredentialAttribute.DoesNotExist:
```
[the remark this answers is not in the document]
staubina: The test I added is specifically looking for the created date and it is resolved. Do you think that is enough verification @oliviaruizknott

credentials/apps/records/tests/test_views.py
staubina: I will update the comment a little.

credentials/apps/records/tests/test_views.py
```diff
@@ -667,15 +667,39 @@ def test_program_visible_date(self, date, completed):
...
+            value=test_date,
+        )
+
+        response = self.client.get(reverse("records:private_programs", kwargs={"uuid": self.program.uuid.hex}))
+        grades = json.loads(response.context_data["record"])["grades"]
+        self.assertEqual(len(grades), 1)
+        self.assertEqual(grades[0]["issue_date"], "2021-01-01T01:01:01+00:00")
+
+    @override_switch("credentials.use_certificate_available_date", active=True)
+    def test_created_date_as_issue_date_with_no_certificate_available_date(self):
```
staubina, line 697: @oliviaruizknott This is the test I mentioned in my reply.

justinhynes (approved):

PR #1305, opened 2021-09-01
fix: corrects issue with issue_date and visible_date

staubina:
This PR fixes an issue where the Program certificate would show an
incorrect issue date based on when a course cert was created in
Credentials rather than when it was issued from the LMS.

### DIFF
--- a/credentials/apps/credentials/utils.py
+++ b/credentials/apps/credentials/utils.py
@@ -57,8 +57,6 @@ def to_datetime(date):
     return datetime.datetime.combine(date, datetime.datetime.min.time())
 
 
-# TODO: Delete this when removing USE_CERTIFICATE_AVAILABLE_DATE toggle:
-# MICROBA-1198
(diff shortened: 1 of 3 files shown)


credentials/apps/credentials/utils.py
```diff
@@ -158,12 +156,27 @@ def _get_program_certificate_visible_date(user_program_credential):
         ).first()
 
         if course_run_cert:
-            date = course_run_cert.credential.certificate_available_date or course_run_cert.created
+            date = _get_issue_date_for_course_credential(course_run_cert)
```
oliviaruizknott, line 159: Could you update the docstring for this method? (~lines 139-140)

oliviaruizknott (approved):
See note about keeping our doc up-to-date. But looks good!

credentials/apps/credentials/utils.py
staubina: Done

PR #1312, opened 2021-09-02
[feat] MB-1464: Add lms_user_id

Tj-Tracy:
A small function that populates the LMS user-id when a user logs into
credentials for the first time. This will only update the lms_user_id
for users that specifically go to credentials; either to their records
page or otherwise. This will not update for users who are awarded a
prgram/course credential.

### DIFF
--- a/credentials/settings/base.py
+++ b/credentials/settings/base.py
@@ -337,6 +337,7 @@ SOCIAL_AUTH_PIPELINE = (
     "auth_backends.pipeline.update_email",
     # Credentials-specific pipeline below
     "credentials.apps.core.utils.update_full_name",
+    "credentials.apps.core.utils.update_lms_user_id",
 )
 
 # Set these to the correct values for your OAuth2 provider (e.g., devstack)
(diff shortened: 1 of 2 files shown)


credentials/settings/base.py
```diff
@@ -460,7 +461,17 @@
         "BACKEND": "django.core.cache.backends.locmem.LocMemCache",
     }
 }
-EDX_DRF_EXTENSIONS = {"OAUTH2_USER_INFO_URL": "http://127.0.0.1:8000/oauth2/user_info"}
+EDX_DRF_EXTENSIONS = {
+    "JWT_PAYLOAD_USER_ATTRIBUTE_MAPPING": {
+        'administrator': 'is_staff',
+        'email': 'email',
+        'full_name': 'full_name',
+        'tracking_context': 'tracking_context',
```
justinhynes, line 469: I'm pretty sure this line will cause issues, it did for me when I had it locally. There is no `tracking_context` field on the User model, and we will try to map a value to the model that doesn't exist.

Tj-Tracy: removed!

tuchfarber (approved):

PR #1330, opened 2021-09-10
feat: Change override date to datetime

oliviaruizknott:
When first building the Certificate Date Override feature, I set up the
UserCredentialDateOverride model to store the override dates as Dates
instead of DateTimes.

Turns out this is not how edX typically handles dates, and it’s causing
some minor headaches around needing to convert values.

There is a timezone issue which I thought this would fix, but it
doesn’t. Will need to look into JavaScript’s localization to solve
that separately.

MICROBA-1488


credentials/apps/credentials/utils.py
```diff
@@ -52,11 +52,6 @@ def keyfunc(attribute):
     return True
 
 
-def to_datetime(date):
-    """Turn a date object into a datetime object by setting time to 00:00"""
-    return datetime.datetime.combine(date, datetime.datetime.min.time())
```
staubina, line 57: 🥳

staubina (approved):

PR #1334, opened 2021-09-13
BOM-2798: Upgrade Credentials to Django 3.2

mzulqarnain1:
## Description

- Upgraded To Use Django 3.2
- Dropped Django 2.2 Support
- Ran Make Upgrade

## Ticket Link

Link to the associated ticket
[BOM-2798](https://openedx.atlassian.net/browse/BOM-2798)


awais786:
I am not sure but is there any chance of templates coming from DB ? If yes need to check the `{% load staticfiles %}` tag in those entries.

awais786 (approved):

staubina:
> I am not sure, is there any chance of templates coming from DB ? If yes need to check the {% load staticfiles %} tag in those entries.

@awais786 We will look into the DB templates, I think those are only for Course Certs in the LMS though. 

Did you look at https://github.com/edx/credentials-themes though?

requirements/constraints.txt
```diff
@@ -11,5 +11,7 @@
 # Common constraints for edx repos
 -c common_constraints.txt
 
+Django<4.0
```
justinhynes, line 14: Should this be `Django<3.3` to stay on the LTS version?

justinhynes (dismissed):
At first glance, this looks good and seems to work fine locally. 

I have a few questions to make sure we are constraining the versions of Django to the LTS versions.

mzulqarnain1:
> > I am not sure, is there any chance of templates coming from DB ? If yes need to check the {% load staticfiles %} tag in those entries.
> 
> @awais786 We will look into the DB templates, I think those are only for Course Certs in the LMS though.
> 
> Did you look at https://github.com/edx/credentials-themes though?

Yes @staubina We have upgraded the templates in [credentials-themes ](https://github.com/edx/credentials-themes ) and the tag is bumped in this PR.

requirements/constraints.txt
justinhynes: I think my concern is all set for now according to this: https://www.djangoproject.com/download/. It doesn't look like there are any versions planned past 3.2.x, and they will move to Django 4.x next. According to Django's plans, I think this constraint is fine.

Tj-Tracy (approved):

PR #1346, opened 2021-09-17
feat: add new management command that will sync lms_user_id data from a temp table

justinhynes:
[MICROBA-1465]
- add new management command that will sync lms_user_id data from a temp table to the core_user table of the Credentials IDA

### DIFF
--- a/credentials/apps/core/management/commands/sync_lms_user_ids.py
+++ b/credentials/apps/core/management/commands/sync_lms_user_ids.py
@@ -1,0 +1,203 @@
+"""
+Django management command to load the `lms_user_id` column from historical user data.
+"""
+
+
+import logging
+
+from django.contrib.auth import get_user_model
+from django.core.management.base import BaseCommand
+from django.db import connection, transaction
+from django.db.models import Max
+
+
+User = get_user_model()
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    """
+    Import username/lms_user_id mappings into the `lms_user_id` column of the Credentials IDA's `core_user` table.
+
+    Export data for this by running this query from the LMS:
+        SELECT *
+        INTO OUTFILE '<csv_file_name>.csv'
+        FIELDS TERMINATED BY ','
+        OPTIONALLY ENCLOSED BY '"'
+        LINES TERMINATED BY '\n'
+        FROM (
+            SELECT
+                username,
+                id as user_id
+            FROM auth_user
+            WHERE id NOT IN (
+                SELECT user_id FROM user_api_userretirementstatus
+            )
+            UNION SELECT
+                original_username AS username,
+                user_id
+            FROM user_api_userretirementstatus
+        ) u
+
+    This management command assumes the data will be accessible from an AWS S3 bucket hosting the exported CSV.
+
+    This commands capabilities are split into multiple 'stages' in order to allow time for manual verifications before
+    and after syncing the LMS user data:
+        - The 'create' stage will create a temporary table and an index on the `username` column.
+        - The 'update_temp' stage will load data from a CSV containing an export of username and user-id records from
+          the LMS.
+        - The 'update_credentials' stage will sync the LMS user ids with the Credentials user records.
+        - The 'delete' stage will remove the temporary table created in the 'create' stage.
+
+    Example usage:
+        - To create the temp table:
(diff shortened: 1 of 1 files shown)


justinhynes:
I didn't write tests for this management command as it is intended to run once to do the backfill. I decided it wasn't worth the time to write extensive tests. I _did_ test this locally (with a slightly modified version that loaded a local CSV over pulling data from an S3 bucket).

staubina (commented):
Do you think it would be worthwhile to make it possible to run each step independently through the command line? That way we can check if each step was successful through the read replica or on the machine?

justinhynes:
> Do you think it would be worthwhile to make it possible to run each step independently through the command line? That way we can check if each step was successful through the read replica or on the machine?

I could definitely see the benefit of breaking it up so we could mix in a bit of manual verification. Going to noodle on this a bit.

credentials/apps/core/management/commands/sync_lms_user_ids.py
```diff
@@ -0,0 +1,210 @@
...
+                user_id
+            FROM user_api_userretirementstatus
+        ) u
+
+    This management command assumes the data will be accessible from an AWS S3 bucket hosting the exported CSV.
+
+    This commands capabilities are split into multiple 'stages' in order to allow time for manual verifications before
+    and after syncing the LMS user data:
+        - The 'create' stage will create a temporary table and an index on the `username` column.
+        - The 'update_temp' stage will load data from a CSV containing username and
```
tuchfarber, line 50: I think you missed some words after "and __"

credentials/apps/core/management/commands/sync_lms_user_ids.py
```diff
@@ -0,0 +1,210 @@
...
+            help="Name of CSV file containing the data used to populate the temp table",
+        )
+        parser.add_argument(
+            "--stage",
+            default=None,
+            help="The 'stage' of the script to run. Can be one of the following choices: 'CREATE', 'UPDATE', or "
+            "'DELETE'",
+        )
+
+    def create_temp_table_with_data(self):
```
tuchfarber, line 87: nit: method name implies data will be loaded here. I'd probably just remove the `with_data` unless there's a reason for it.

credentials/apps/core/management/commands/sync_lms_user_ids.py
```diff
@@ -0,0 +1,210 @@
...
+        Loads data from a CSV file (housed in an S3 bucket) to a temporary table in the Credentials database.
+        """
+        load_data_sql = """
+            LOAD DATA FROM S3 FILE '%s/%s'
+            INTO TABLE %s
+            FIELDS TERMINATED BY ','
+            OPTIONALLY ENCLOSED BY '"'
+            LINES TERMINATED BY '\\n'
+                (username, lms_user_id)
+        """ % (
```
tuchfarber, line 129: (params during execute)

credentials/apps/core/management/commands/sync_lms_user_ids.py
```diff
@@ -0,0 +1,210 @@
...
+            if max_user_id is None:
+                max_user_id = User.objects.aggregate(Max("id"))["id__max"]
+
+            batch_size = options.get("batch_size")
+            starting_user_id = options.get("starting_user_id")
+
+            self.update_credentials_user_table(batch_size, starting_user_id, max_user_id)
+        elif stage.lower() == "delete":
+            self.drop_temp_table()
+        else:
```
tuchfarber, line 209: You can avoid this by adding a `choices` to your argument.
src: https://docs.python.org/3/library/argparse.html#choices

credentials/apps/core/management/commands/sync_lms_user_ids.py
```diff
@@ -0,0 +1,210 @@
...
+                (username, lms_user_id)
+        """ % (
+            bucket_name,
+            file_name,
+            TEMP_TABLE_NAME,
+        )
+
+        with connection.cursor() as cursor:
+            logger.info("Loading temp table with LMS user data")
+            cursor.execute(load_data_sql)
```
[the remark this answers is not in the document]
justinhynes: Good idea! I'll see if I can add this in.

tuchfarber (approved):

PR #1353, opened 2021-09-22
fix: Updated middleware to support edX JWT auth

staubina:


justinhynes (approved):
Thanks for digging into this issue!

PR #1377, opened 2021-10-05
Move django rest swagger to drf-yasg

NeOneSoft:
This PR was created in order to remove django-rest-swagger package and use drf-yasg.


natabene:
@NeOneSoft Thanks so much for this PR, is this for https://github.com/edx/upgrades/issues/17 ?

PR #1556, opened 2022-02-10
build: Adding migrations check for mysql8.

awais786:
build: Adding migrations check for mysql8

https://openedx.atlassian.net/browse/BOM-3279


iamsobanjaved (approved):
Looks good

PR #1696, opened 2022-06-29
fix: fix syntax error

oliviaruizknott:


justinhynes (approved):
Thanks for fixing forward so quick!!

PR #1743, opened 2022-09-02
feat: Add ProgramCertRecord data to API response for Learner Record MFE rendering decisions.

justinhynes:
[APER-1946]

Adds a new field to the dictionary response of the `get_record_data` function. The new data will be information about the shared ProgramCertRecord for a Program, if it exists. This data will be returned in an existing REST API to the Learner Record MFE where it will be used to conditionally render button states in an individual Program Record.

* Move private function `_get_course_data` to top of the file.
* Updated code comments and docstrings to provide additional context wherever possiible.
* Refactors the existing logic that retrieves the data presented in our Program Record page in an attempt to reduce cognitive load. I found it very difficult to follow all the logic in the original mega-function. While this is not perfect, I think it's better than it was before.
* Include the UUID of a shared program record if it exists
* Add tests for refactored functions and new functionality


credentials/apps/records/api.py
```diff
@@ -177,23 +214,84 @@ def _get_course_data(program_course_runs, highest_attempt_dict, user_credential_
...
+    Returns:
+        String: The UUID of the shared program record converted to a String
+    """
+    # if a ProgramCertRecord instance is found then return the UUID (as a string), otherwise return None
+    try:
+        shared_program_record = ProgramCertRecord.objects.get(user=user.id, program=program.id)
+    except ProgramCertRecord.DoesNotExist:
+        return None
+    else:
+        return str(shared_program_record.uuid)
```
justinhynes, line 244: This is the new functionality, we check and return the UUID (as a string) of the shared program cert record (if it exists).

MaxFrank13, line 244: If I understand correctly, this will get a `ProgramCertRecord` that matches the `user.id` and `program.id`. If this doesn't exist, it returns nothing (`None`). Otherwise, it hits the `else` and returns a the UUID string.

Does `None` evaluate to `null`, `false`, or an empty string here?

MaxFrank13 (approved):
This all looks solid to me. I just had a couple questions for my own sake. Thanks for refactoring this massive function and making it easier to follow!

credentials/apps/records/api.py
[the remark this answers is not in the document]
MaxFrank13: Thanks for pointing that out!

credentials/apps/records/api.py
```diff
@@ -15,140 +15,179 @@
...
+
+    Args:
+        program (Program): Program object instance
+        user (User): Django User object
+    """
     program_pathways = program.pathways.all()
     program_pathways_set = frozenset(program_pathways)
+
     user_credit_pathways = (
         UserCreditPathway.objects.select_related("pathway").filter(user=user, pathway__in=program_pathways_set).all()
```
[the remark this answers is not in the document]
justinhynes: I think the official docs will do a better job explaining this than I could: https://docs.djangoproject.com/en/4.1/ref/models/querysets/#select-related.

It is a way to get data related to a model (through its foreign key relations) in a single database query through the ORM. I think the intent here is to have the ORM do the heavy lifting and fetch extra data related to the item you are querying for in a single database query (and hopefully reducing the overall number of queries you need to make to the backend/data system).

credentials/apps/records/api.py
justinhynes: Good question! This is a neat little feature of Python.

Basically this is saying; Go and get the ProgramCertRecord for user x in program y. If a record does not exist for this combination, we hit a `DoesNotExist` exception and we pass back None. This will get converted to null in our response by DRF (as `None` is not a valid JSON value). 

The `else` is only evaluated if (and only if) an exception isn't generated. We only want to convert that UUID to a string if we successfully found a ProgramCertRecord.

There is also an option for a `finally` clause in a try/except that will _always_ run regardless of an exception or not. 

Some more examples here: https://www.geeksforgeeks.org/try-except-else-and-finally-in-python/.

PR #1747, opened 2022-09-07
refactor: REST API to return a different response if LR MFE is enabled

MaxFrank13:
This PR refactors a REST API endpoint that creates a `ProgramCertRecord` and returns the URL for that public record. Currently, that URL is routing to the legacy frontend. With this refactor, if the Learner Record MFE is enabled, it will instead return the URL for the MFE.

### DIFF
--- a/credentials/apps/records/tests/test_views.py
+++ b/credentials/apps/records/tests/test_views.py
@@ -953,7 +953,7 @@ class ProgramRecordTests(SiteMixin, TestCase):
         self.assertTrue(response.url.startswith("/login/?next="))
 
     def test_user_creation(self):
-        """Verify successful creation of a ProgramCertRecord and return of a uuid"""
+        """Verify successful creation of a ProgramCertRecord and return of a URL with the uuid for the public record"""
         rev = reverse("records:share_program", kwargs={"uuid": self.program.uuid.hex})
         data = {"username": self.USERNAME}
         jdata = json.dumps(data).encode("utf-8")
@@ -962,6 +962,22 @@ class ProgramRecordTests(SiteMixin, TestCase):
 
         self.assertEqual(response.status_code, 201)
         self.assertRegex(json_data["url"], UUID_PATTERN)
+
+    @override_settings(USE_LEARNER_RECORD_MFE=True)
+    @override_settings(LEARNER_RECORD_MFE_RECORDS_PAGE_URL="http://some.website/page/")
+    def test_user_creation_with_MFE(self):
+        """
+        Verify successful creation of a ProgramCertRecord and return of a URL with the uuid for the public record
+        (Learner Record MFE)
+        """
+        rev = reverse("records:share_program", kwargs={"uuid": self.program.uuid.hex})
+        data = {"username": self.USERNAME}
+        jdata = json.dumps(data).encode("utf-8")
+        response = self.client.post(rev, data=jdata, content_type=JSON_CONTENT_TYPE)
+        json_data = response.json()
+
+        self.assertEqual(response.status_code, 201)
+        self.assertRegex(json_data["url"], rf"http://some.website/page/shared/{UUID_PATTERN}")
 
     def test_different_user_creation(self):
         """Verify that the view rejects a User attempting to create a ProgramCertRecord for another"""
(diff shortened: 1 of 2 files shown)


credentials/apps/records/views.py
```diff
@@ -289,7 +296,12 @@ def post(self, request, **kwargs):
         pcr, created = ProgramCertRecord.objects.get_or_create(user=user, program=program)
         status_code = 201 if created else 200
 
+        if settings.USE_LEARNER_RECORD_MFE:
+            response = {"url": settings.LEARNER_RECORD_MFE_RECORDS_PAGE_URL + "shared/" + pcr.uuid.hex}
```
justinhynes, line 300: Instead of using string concatenation to build the URI, let's use a format string. You should be able to update this to something that looks like...
```
response = {"url": f"{settings. LEARNER_RECORD_MFE_RECORDS_PAGE_URL}shared/{pcr.uuid.hex}"
```
While I don't _think_ this particular piece of code could result in something malicious, using string concatenation to build URI's and SQL is generally considered a dangerous practice (from a security standpoint). 

In the above example, everything in our brackets, e.g. `{pcr.uuid.hex}`, would be converted to a string.

IIRC, while this isn't a particular complicated piece of code, f-strings are also the most performant way to do string interpolation in Python.

More links on format-strings: https://peps.python.org/pep-0498/ and https://realpython.com/python-f-strings/.

credentials/apps/records/views.py
```diff
@@ -266,8 +266,15 @@ def post(self, request, **kwargs):
...
 class ProgramRecordCreationView(LoginRequiredMixin, RecordsEnabledMixin, View):
     """
-    Creates a new Program Certificate Record from given username and program uuid,
-    returns the uuid of the created Program Certificate Record
+    Creates a new Program Certificate Record from given username and program uuid
+
+    POST: /programs/:program_uuid/share/
+
+    Returns:
+        URL for the created Program Certificate Record
```
justinhynes, line 274: I think the format for the `Returns` here _should_ be something like...
```
Returns:
    Dict: Dictionary containing the public URL to the program certificate record instance
```
While it is true that the data _inside_ the dictionary is the URL to the shared record, the "Returns" section of the docstring should describe the type of data structure (in this case a dictionary) the data is. 

If I was reading this and had no prior knowledge I would expect that I am getting a string returned when calling this function. That is not the case.

credentials/apps/records/tests/test_views.py
```diff
@@ -953,7 +953,7 @@ def test_login_required(self):
         self.assertTrue(response.url.startswith("/login/?next="))
 
     def test_user_creation(self):
-        """Verify successful creation of a ProgramCertRecord and return of a uuid"""
+        """Verify successful creation of a ProgramCertRecord and return of a URL with the uuid for the public record"""
```
justinhynes, line 956: Thank you for adding additional context. I appreciate you taking the time here! :)

credentials/apps/records/tests/test_views.py
```diff
@@ -963,6 +963,21 @@ def test_user_creation(self):
...
+        Verify successful creation of a ProgramCertRecord and return of a URL with the uuid for the public record
+        (Learner Record MFE)
+        """
+        rev = reverse("records:share_program", kwargs={"uuid": self.program.uuid.hex})
+        data = {"username": self.USERNAME}
+        jdata = json.dumps(data).encode("utf-8")
+        response = self.client.post(rev, data=jdata, content_type=JSON_CONTENT_TYPE)
+        json_data = response.json()
+        self.assertEqual(response.status_code, 201)
+        self.assertRegex(json_data["url"], UUID_PATTERN)
```
justinhynes, line 979: One last suggestion. Can we improve the verification of this unit test? Instead of checking if the UUID is contained within the string, can we verify that the entire string is being built as we expected? 

Since we are are now building a link to an external system... this would be a stronger test if we verified that the link is exactly as we expect it to be.

justinhynes (approved):
LGTM! Thanks for being open to the suggestions I posed. :)

credentials/apps/records/tests/test_views.py
MaxFrank13: Sounds like another job for a format string!

PR #1753, opened 2022-09-12
feat: Add program to UserCreditPathway

cdeery:
Add program to userCreditPathway.
This will not backfill the existing UserCreditPathway entries, they will have a null value for Program.
If there is an existing ProgramCertRecord for a given user and program, and a userCreditPathway with a null program, then posting the UserCreditPathway for that user+pathway+program will update the record with a null program.
Otherwise a new UserCreditPathway with that combination of user+pathway+program will be created.

Also removed the limit on only being able to send a userCreditPathway one time.


credentials/apps/records/migrations/0021_alter_usercreditpathway_unique_together.py
```diff
@@ -0,0 +1,20 @@
+# Generated by Django 3.2.14 on 2022-09-08 20:55
```
justinhynes, line 1: Would it be possible to rework the PR in such a way so that there is just one migration on this table? Does there need to be two? It looks like this one is just further altering the same table/model.

credentials/apps/records/tests/test_views.py
```diff
@@ -1121,20 +1124,64 @@ def test_email_content_incomplete(self):
         email = mail.outbox[0]
         self.assertIn("has sent their partially completed program record for", str(email.message()))
 
-    def prevent_sending_second_email(self):
+    def test_allowed_sending_second_email(self):
         """Verify that an email can't be sent twice"""
```
justinhynes, line 1128: Nit: Can we update the description here? It now conflicts with what the test is actually doing :)

credentials/apps/records/views.py
```diff
@@ -252,13 +249,33 @@ def post(self, request, **kwargs):
...
+            preexisting_program_cert_record
+            and UserCreditPathway.objects.filter(user=user, pathway=pathway, program=None).exists()
+        ):
+            # A program for this user already existed, and a pathway for this user without a program exists.
+            # Update the status and program
+            UserCreditPathway.objects.update_or_create(
+                user=user,
+                pathway=pathway,
+                program=None,
+                defaults={"status": UserCreditPathwayStatus.SENT, "program": program},
```
justinhynes, line 270: I wanted to check an assumption I'm making reading this. We have... 

```
            UserCreditPathway.objects.update_or_create(
                user=user,
                pathway=pathway,
                program=None,
                defaults={"status": UserCreditPathwayStatus.SENT, "program": program},
``` and then we pass the desired program in the `defaults` dict.  

My assumption is that we have to pass `program=None` in the `update_or_create` call to ensure we select the correct record to update and not create a new one? Otherwise, if we pass the desired value of `program` as a keyword argument (rather than in the `defaults` dict) the ORM won't find a match?

File: docs/credentials_admin.rst [hunk 4]
```diff
@@ -141,7 +141,7 @@ User Credit Pathways
 ~~~~~~~~~~~~~~~~~~~~
 
 To add or update a user credit pathway, use the ``UserCreditPathway`` model. This model is used to track which
-users have sent which records to which credit pathways.
+users have sent which records to which credit pathways for which program.
```
justinhynes, line 144: Thanks for makign sure the README's are correct :D

credentials/apps/records/views.py
cdeery: Yes, the syntax is a little confusing. The program=None argument is used to match the object in the database, and the defaults argument sets the values to be updated. If the object is not found, then the defaults override the query values. 
Should I add a comment? It is certainly a little confusing to read.

credentials/apps/records/tests/test_views.py
cdeery: The test actually switched to testing the opposite behavior. We can now send a second email, so the name was changed (so that it would run as a test) as well as the acceptance criteria switching to checking for an HTTP 200.

justinhynes: I've specifically mentioning the docstring that says:
` """Verify that an email can't be sent twice"""`. We _can_ send the email twice now. The docstring is just incorrect now.

credentials/apps/records/views.py
justinhynes: I think that's a good idea. It took me a moment to figure out, at first I was like "why can't we just set program to the program we want?" but then I realized the ORM might have trouble finding the record we want otherwise.

credentials/apps/records/tests/test_views.py
cdeery: Whoops! Thanks

credentials/apps/records/migrations/0021_alter_usercreditpathway_unique_together.py
cdeery: Done. Thanks for the tips on how to do it

credentials/apps/records/tests/test_views.py
cdeery: Done

credentials/apps/records/views.py
cdeery: Done

justinhynes (approved):
Thanks for squashing the migration into one file and accommodating the other requests. LGTM.

PR #1757, opened 2022-09-15
feat: Add `lms_last_updated_at` field to the UserGrade model

justinhynes:
[APER-1968]

Today, we don't have a good way to understand if grade data is out of sync with the LMS. Grades are sent to Credentials via a REST API call originating from an asynchronous Celery task on the LMS side.

This PR introduces a new field to the UserGrade model named `lms_last_updated_at`. This field will store the `modified` date of the corresponding grade coming from the LMS for a learner. This way we will know exactly the last time the grade was updated in the LMS and will be able to tell how in sync the grades are from the two systems.

* Add the (optional) `lms_last_updated_at` field to the UserGrade model.
* Update existing Grades endpoint to accept this data if provided.

Related LMS PR: https://github.com/openedx/edx-platform/pull/30994


credentials/apps/records/admin.py
```diff
@@ -41,6 +41,7 @@ class UserGradeAdmin(admin.ModelAdmin):
         "course_run__key",
     )
     raw_id_fields = ("course_run",)
+    readonly_fields = ["lms_last_updated_at"]
```
justinhynes, line 44: Make is so the DateTime associated with a UserGrade instance can't be modified via Django Admin.

credentials/apps/api/v2/tests/test_serializers.py
```diff
@@ -167,13 +170,16 @@ def test_to_representation(self):
     def test_to_internal_value(self):
         Request = namedtuple("Request", ["site"])
         serializer = UserGradeSerializer(context={"request": Request(site=self.site)})
+        updated_at_dt = datetime.now()
+        updated_at_utc = updated_at_dt.replace(tzinfo=pytz.UTC)
```
cdeery, line 174: Are the lms_updated_at entries always in UTC?

cdeery:
Is the date always stored in GMT? Is it required to be sent in GMT? I think I need a comment somewhere about the expected or allowed time zone.

credentials/apps/api/v2/tests/test_serializers.py
justinhynes: @cdeery Yes. Additionally, `USE_TZ` is enabled in Credentials and I don't believe it will accept timezone naive datetimes (so I had to convert this to a UTC DateTime in the test).

cdeery (approved):
Thanks for answering my questions. This looks good to go.

PR #1760, opened 2022-09-21
refactor: modify REST API endpoints for LR MFE

MaxFrank13:
This is a PR to modify the existing REST API endpoints used by the LR MFE. This modification adds a `username` query param that is employed the the Support Tools MFE to fetch the same data learners see. 

Custom permissions classes were necessary here to block access from unauthorized users. We needed to ensure that learners (or other non-staff, non-superusers) couldn't view the records of others.

[Here is a truth table](https://docs.google.com/spreadsheets/d/1kNZ1Ff0jz_-cQ00yyZrpII95R7565QuRVeghEAppfI0/edit#gid=0) that attempts to illustrate how access is handled with the addition of this query parameter.

### DIFF
--- a/credentials/apps/records/rest_api/v1/permissions.py
+++ b/credentials/apps/records/rest_api/v1/permissions.py
@@ -1,0 +1,36 @@
+"""
+Custom permissions classes for use with DRF.
+"""
+from edx_rest_framework_extensions.permissions import get_username_param
+from rest_framework import permissions
+
+
+class IsPublic(permissions.BasePermission):
+    """
+    Allows access if URL is for a public record.
+    """
+
+    def has_permission(self, request, view):
+        """
+        Check to see what type of record (private/public) in the query parameter.
+        If it is not public, run a normal authentication permission check.
+        """
+        query_param_is_public = request.query_params.get("is_public", "")
+        is_public = query_param_is_public.lower() == "true"
+        if not is_public:
+            return bool(request.user and request.user.is_authenticated)
+        return True
+
+
+class CanAccessProgramRecord(permissions.BasePermission):
+    """
+    Allows access if the requesting user a staff member or a superuser.
+    """
+
+    def has_permission(self, request, view):
+        """
+        If there is a username query param, check for either superuser or staff access.
+        """
+        if get_username_param(request):
+            return request.user and (request.user.is_superuser or request.user.is_staff)
+        return True
--- a/credentials/apps/records/rest_api/v1/tests/test_views.py
+++ b/credentials/apps/records/rest_api/v1/tests/test_views.py
@@ -58,6 +58,9 @@ class ProgramRecordsViewTests(SiteMixin, TestCase):
             credential_content_type=self.program_credential_content_type,
             credential=self.program_cert,
         )
+        self.program_cert_record = ProgramCertRecordFactory(
+            uuid=self.program.uuid, program=self.program, user=self.user
+        )
 
     def serialize_program_records(self):
         request = APIRequestFactory(SERVER_NAME=self.site.domain).get("/")
@@ -69,13 +72,11 @@ class ProgramRecordsViewTests(SiteMixin, TestCase):
             many=True,
         ).data
 
-    def serialize_program_record_details(self, is_public):
-        url = "/" + str(self.program.uuid) + "/?is_public=" + str(is_public)
+    def serialize_program_record_details(self, uuid, is_public):
+        url = f"/{uuid}/?is_public={is_public}"
         request = APIRequestFactory(SERVER_NAME=self.site.domain).get(url)
         return ProgramRecordSerializer(
-            get_program_details(
-                request_user=self.user, request_site=self.site, uuid=self.program.uuid, is_public=is_public
-            ),
+            get_program_details(request_user=self.user, request_site=self.site, uuid=uuid, is_public=is_public),
             context={"request": request},
         ).data
 
@@ -84,11 +85,30 @@ class ProgramRecordsViewTests(SiteMixin, TestCase):
         response = self.client.get("/records/api/v1/program_records/")
         self.assertEqual(response.status_code, 401)
 
+    def test_details_deny_unauthenticated_user(self):
+        self.client.logout()
+        uuid = str(self.program.uuid).replace("-", "")
+        response = self.client.get(f"/records/api/v1/program_records/{uuid}/?is_public=false")
+        self.assertEqual(response.status_code, 401)
+
     def test_allow_authenticated_user(self):
         """Verify the endpoint requires an authenticated user."""
         self.client.logout()
         self.client.login(username=self.user.username, password=USER_PASSWORD)
         response = self.client.get("/records/api/v1/program_records/")
+        self.assertEqual(response.status_code, 200)
+
+    def test_details_allow_authenticated_user(self):
+        self.client.logout()
+        self.client.login(username=self.user.username, password=USER_PASSWORD)
+        uuid = str(self.program.uuid).replace("-", "")
+        response = self.client.get(f"/records/api/v1/program_records/{uuid}/?is_public=false")
+        self.assertEqual(response.status_code, 200)
+
+    def test_allow_public_record(self):
+        public_record_uuid = str(self.program_cert_record.uuid).replace("-", "")
+
+        response = self.client.get(f"/records/api/v1/program_records/{public_record_uuid}/?is_public=true")
         self.assertEqual(response.status_code, 200)
 
     def test_get(self):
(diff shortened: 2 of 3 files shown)


credentials/apps/records/rest_api/v1/permissions.py
```diff
@@ -0,0 +1,40 @@
...
+            return bool(request.user and request.user.is_authenticated)
+        return True
+
+
+class CanAccessProgramRecord(permissions.BasePermission):
+    """
+    Allows access if the requesting user is the owner of the record, a staff member, or a superuser.
+    """
+
+    def has_permission(self, request, view):
```
cdeery, line 34: How does this check if the requesting user is the owner of the record?

MaxFrank13: Oh nice catch! It actually doesn't. I need to change the documentation. I'm under the impression that a learner/normal user should never need to use this query parameter.

credentials/apps/records/rest_api/v1/tests/test_views.py
```diff
@@ -117,17 +138,14 @@ def test_get_private_details(self):
         self.assertFalse(response.data["is_public"], "Query paramater is set to public when it should be private")
```
cdeery, line 138: the tiniest nit, and not your fault: "parameter"

credentials/apps/records/rest_api/v1/permissions.py
```diff
@@ -0,0 +1,40 @@
+"""
+Custom permissions classes for use with DRF.
+"""
+from django.contrib.auth import get_user_model
+from edx_rest_framework_extensions.permissions import get_username_param
+from rest_framework import permissions
+
+
+User = get_user_model()
```
justinhynes, line 9: Do we need to import the User model in here? I don't think I see it used anywhere?

credentials/apps/records/rest_api/v1/tests/test_views.py
```diff
@@ -69,13 +72,11 @@ def serialize_program_records(self):
             many=True,
         ).data
 
-    def serialize_program_record_details(self, is_public):
-        url = "/" + str(self.program.uuid) + "/?is_public=" + str(is_public)
+    def serialize_program_record_details(self, uuid, is_public):
+        url = "/" + str(uuid) + "/?is_public=" + str(is_public)
```
justinhynes, line 76: Nit: You can simplify this, make it more readable, _and_ make it more "Pythonic" by using Python's [f-strings
](https://peps.python.org/pep-0498/) to handle the string interpolation. This would would become something like...
```
url = f"/{uuid}/?is_public={is_public}"
```

credentials/apps/records/rest_api/v1/tests/test_views.py
```diff
@@ -84,13 +85,32 @@ def test_deny_unauthenticated_user(self):
         response = self.client.get("/records/api/v1/program_records/")
         self.assertEqual(response.status_code, 401)
 
+    def test_details_deny_unauthenticated_user(self):
+        self.client.logout()
+        uuid = str(self.program.uuid).replace("-", "")
+        response = self.client.get("/records/api/v1/program_records/" + uuid + "/?is_public=false")
```
justinhynes, line 91: Nit: Another place I'd use f-strings over string concatenation:
```
response = self.client.get(f"/records/api/v1/program_records/{uuid}/?is_public=false"
```

credentials/apps/records/rest_api/v1/tests/test_views.py
```diff
@@ -84,13 +85,32 @@ def test_deny_unauthenticated_user(self):
...
         self.client.logout()
         self.client.login(username=self.user.username, password=USER_PASSWORD)
         response = self.client.get("/records/api/v1/program_records/")
         self.assertEqual(response.status_code, 200)
 
+    def test_details_allow_authenticated_user(self):
+        self.client.logout()
+        self.client.login(username=self.user.username, password=USER_PASSWORD)
+        uuid = str(self.program.uuid).replace("-", "")
+        response = self.client.get("/records/api/v1/program_records/" + uuid + "/?is_public=false")
```
justinhynes, line 105: Nit: Another place here; favor f-strings over string concatenation.

credentials/apps/records/rest_api/v1/permissions.py
MaxFrank13: I believe I had it in there for a previous implementation of the permission class. I have removed it now though

credentials/apps/records/rest_api/v1/views.py
```diff
@@ -28,9 +36,20 @@ def list(self, request, *args, **kwargs):
...
+        if query_param_username:
+            # fetch Django User object
+            searched_user = User.objects.get(username=query_param_username)
+
+            programs = get_user_program_data(
+                searched_user, request.site, include_empty_programs=False, include_retired_programs=True
+            )
+        else:
+            programs = get_user_program_data(
+                request.user.username, request.site, include_empty_programs=False, include_retired_programs=True
```
justinhynes, line 51: A couple of things...

1. I think the information we're passing to `get_user_program_data` here differs in your `if` vs. your `else`. In the `if`, you are sending the entire User object to `get_user_program_data`. In your `else`, you are sending just the username. Which should it be?

2. I think we want to code a bit more defensively. I think this code has a chance to throw an exception and maybe not return any data back to the caller. A call to {Model}.objects.get(blah=blah) will throw a `DoesNotExist` exception if the object doesn't exist. We will likely want to do something like...
```
username = None

query_param_username = request.query_params.get("username", "")
if query_param_username:
    try:
        other_user = User.objects.get(username=query_param_username)
    except User.DoesNotExist:
        # Log something here
        return Response({"enrolled_programs": []}) # empty list, string, dictionary, whatever we would pass back when there is no data
    else:
        (else section only executes if the "try" didn't throw an exception)
        username = other_user.username

programs = get_user_program_data(
    username, request.site, include_empty_programs=False, include_retired_programs=True
)
```
This way we also cut down on code duplication of the call to `get_user_program_data`.

credentials/apps/records/rest_api/v1/views.py
```diff
@@ -50,13 +69,27 @@ def retrieve(self, request, *args, **kwargs):
...
+                    uuid=kwargs["pk"],
+                    is_public=is_public,
+                )
+            else:
+                program = get_program_details(
+                    request_user=request.user,
+                    request_site=request.site,
+                    uuid=kwargs["pk"],
+                    is_public=is_public,
+                )
```
justinhynes, line 92: Similarly here, can we find a way to code this a bit more defensively and reduce the code duplication? Thanks!

justinhynes (changes requested):
Let me know if you have any questions regarding my comments/suggestions!

credentials/apps/records/rest_api/v1/views.py
justinhynes, line 92: Similarly here, can we find a way to code this a bit more defensively and reduce the code duplication? Thanks!

credentials/apps/records/rest_api/v1/views.py
MaxFrank13: I had a feeling there may be a better way to do this. Thanks for the example!

MaxFrank13: An interesting note about the first question you had: It is supposed to be a username, however, it was still working by passing in the `User`. Is this some Django magic perhaps?

credentials/apps/records/rest_api/v1/views.py
```diff
@@ -28,9 +39,23 @@ def list(self, request, *args, **kwargs):
         Returns:
             response(dict): Information about the user's enrolled programs
         """
+        username = request.user.username
+        # check for username param
```
justinhynes, line 43: Could we expand on this comment slightly? Something so anyone reading this knows why we're substituting the username in the query params over the one in the request object? Maybe something along the lines of...

"Check for a username query parameter. If it exists, we know we want to make this request on behalf of the specified user in the query parameters. Otherwise, we will make the request using the user from the request object. Used by the the support tools MFE blah blah blah"

This could certainly help dispel confusion from anyone in the Open edX community who may read this code, or even ourselves a year from now when we come back and look at something in this IDA. :)

credentials/apps/records/rest_api/v1/views.py
```diff
@@ -28,9 +39,23 @@ def list(self, request, *args, **kwargs):
...
+        query_param_username = request.query_params.get("username", "")
+
+        if query_param_username:
+            try:
+                searched_user = User.objects.get(username=query_param_username)
+            except User.DoesNotExist:
+                log.error("A user matching this username does not exist")
```
justinhynes, line 50: Can we also log the username in this log message? Otherwise, when we look at logs, we won't know what data was bad data.

credentials/apps/records/rest_api/v1/views.py
```diff
@@ -50,13 +75,27 @@ def retrieve(self, request, *args, **kwargs):
         query_param_is_public = request.query_params.get("is_public", "")
         is_public = query_param_is_public.lower() == "true"
 
+        user = request.user
+        # check for username param
```
justinhynes, line 79: Similar comment here; can we add some extra context here?

credentials/apps/records/rest_api/v1/views.py
```diff
@@ -50,13 +75,27 @@ def retrieve(self, request, *args, **kwargs):
...
+        query_param_username = request.query_params.get("username", "")
+
         try:
+            if query_param_username:
+                try:
+                    searched_user = User.objects.get(username=query_param_username)
+                except User.DoesNotExist:
+                    log.error("A user matching this username does not exist")
```
justinhynes, line 87: Similar comment here; can we add the username to the log message?

justinhynes (approved):
I had a few minor requests to update logs and comments but I don't need to re-review afterwards. Thanks, Max!

PR #1777, opened 2022-10-05
fix: revert recent change to CourseGradeSerializer

justinhynes:
In PR #1763, Django Rest Framework made a breaking change in a minor version release of their package. When attempting to fix this issue, I noticed some logic that was changing/updating data before validation in an overridden `is_valid` function. I made this change based on some reading I had done suggesting that this kind of data modification should be done in a dedicated Serializer Field class or by overriding the `to_internal_value` function in the Serializer class.

After making this change we were alerted to a number of errors occurring in the Credentials service. It looks like the data dict I was attempting to modify is immutable and can't be changed. Thus, I need to go back to the original implementation.


MaxFrank13 (approved):
LGTM!

PR #1793, opened 2022-10-18
feat: redirect users to LR MFE records pages where appropriate

justinhynes:
[APER-1975]

It is possible for learners to generate a link to a public version of their program record(s) to allow external entities to view their progress. This is particularly important as we embed these links within messages that get shared via credit pathways. We wanted to ensure that these links are not broken when the default experience becomes the Learner Record MFE, so we will be redirecting requests to the legacy view to the Learner Record MFE.

* Implement a redirect when a public program record is being requested (and the Learner Record MFE is enabled)
* Implement a redirect when hitting the records index page (and the Learner Record MFE is enabled)
* Fix an issue with the shared program record UUID being returned with dashes (Credentials can't process UUIDs with dashes in the program record view)

### DIFF
--- a/credentials/apps/records/api.py
+++ b/credentials/apps/records/api.py
@@ -241,7 +241,7 @@ def _get_shared_program_cert_record_data(program, user):
     except ProgramCertRecord.DoesNotExist:
         return None
     else:
-        return str(shared_program_record.uuid)
+        return str(shared_program_record.uuid.hex)
 
 
 def get_program_record_data(user, program_uuid, site, platform_name=None):
--- a/credentials/apps/records/tests/test_views.py
+++ b/credentials/apps/records/tests/test_views.py
@@ -286,6 +286,18 @@ class RecordsViewTests(SiteMixin, TestCase):
         ]
         self.assertEqual(program_data, expected_program_data)
 
+    @override_settings(USE_LEARNER_RECORD_MFE=True)
+    @override_settings(LEARNER_RECORD_MFE_RECORDS_PAGE_URL="http://some.website/page/")
+    def test_index_page_redirect(self):
+        """
+        Verify that a public program cert record request is redirected to the Learner Record MFE when enabled.
+        """
+        expected_redirect_url = "http://some.website/page/"
+
+        response = self.client.get(reverse("records:index"))
+        assert response.status_code == 302
+        assert response["Location"] == expected_redirect_url
+
 
 # This view shares almost all the code with RecordsView above. So we'll just test the interesting differences.
 @ddt.ddt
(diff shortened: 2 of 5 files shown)


credentials/apps/records/api.py
```diff
@@ -241,7 +241,7 @@ def _get_shared_program_cert_record_data(program, user):
     except ProgramCertRecord.DoesNotExist:
         return None
     else:
-        return str(shared_program_record.uuid)
+        return str(shared_program_record.uuid.hex)
```
justinhynes, line 244: This fixed a bug with the generated links, Credentials won't process a URL with a UUID that contains dashes, so we strip them out.

credentials/apps/records/tests/test_views.py
```diff
@@ -966,22 +990,6 @@ def test_user_creation(self):
...
-        """
-        rev = reverse("records:share_program", kwargs={"uuid": self.program.uuid.hex})
-        data = {"username": self.USERNAME}
-        jdata = json.dumps(data).encode("utf-8")
-        response = self.client.post(rev, data=jdata, content_type=JSON_CONTENT_TYPE)
-        json_data = response.json()
-
-        self.assertEqual(response.status_code, 201)
-        self.assertRegex(json_data["url"], rf"http://some.website/page/shared/{UUID_PATTERN}")
-
```
justinhynes, line 984: This test no longer makes sense to keep around. The behavior of the functionality under test no longer acts differently if the LR MFE is enabled.

credentials/apps/records/views.py
```diff
@@ -98,6 +99,15 @@ def _get_programs(self):
             include_retired_programs=True,
         )
 
+    # NOTE: We _should_ keep this for redirecting users to the Learner Record MFE to continue to work correctly
+    def get(self, request, *args, **kwargs):
+        # If the Learner Record MFE is enabled, redirect our user to the MFE, otherwise we use the legacy frontend
+        if settings.USE_LEARNER_RECORD_MFE:
+            return HttpResponseRedirect(settings.LEARNER_RECORD_MFE_RECORDS_PAGE_URL)
+
+        return super().get(request, *args, **kwargs)
```
justinhynes, line 108: I also set up a redirect for learners who would land on the index page (credentials.edx.org/records). I believe our support documentation (and possibly other places) direct here, and I thought it would be a good use of a redirect as well.

This could potentially allow us to help learners avoid broken links in our documentation when we switch over to using the Learner Record MFE.

credentials/apps/records/views.py
```diff
@@ -161,6 +176,20 @@ def _get_record(self, uuid, is_public):
...
+    def get(self, request, *args, **kwargs):
+        # If we're using the Learner Record MFE for displaying program records AND this is a public program record
+        # request, redirect the request to the Learner Record MFE.
+        is_public = kwargs["is_public"]
+        if settings.USE_LEARNER_RECORD_MFE and is_public:
+            uuid = kwargs["uuid"]
+            url = urllib.parse.urljoin(settings.LEARNER_RECORD_MFE_RECORDS_PAGE_URL, f"shared/{uuid}")
+            return HttpResponseRedirect(url)
+
+        return super().get(request, *args, **kwargs)
```
MaxFrank13, line 190: I remember we discussed this but for further clarification, why don't we want to redirect a link to an individual Program Record page? i.e. `credentials.edx.org/records/programs/{uuid}`

justinhynes: 🤔 You know, I don't have a good reason. We had already done some work in the LMS to make sure we format the links to the Credentials IDA correctly (from the Program Dashboard). 

I think the main way I was looking at this was that I know we have links to the index page in the wild (in our Help Center documentation) as well as links to the public records (from pathways and users generating/sending them themselves). 

I just had some mental assumption that learners would navigate to the private pages in a more "traditional" manner by: 
- Hitting the index page and selecting which Program they want to view, or
- Using the Program Dashboard to get to the Credentials IDA (and the program record) directly (and the LMS will format the link to the right place)

I could add some additional logic to do a redirect if you think it's a good addition.

MaxFrank13: It seems like a pretty rare case that someone would end up using this link. I guess the only consideration is that it's still possible to reach the legacy UI this way and maybe we'd rather have users see the new UI

justinhynes: I am not so concerned about people getting to the legacy UI (as there is no functional difference), but I was concerned about when we _remove_ the legacy UI. Either way, should be easy enough to correct through proper channels.

If anyone else sees this comment thread: our team talked about this a bit and we're not going to implement any additional redirects.

Issue #1722, opened 2022-08-19
[DEPR]: Program Record UI


hurtstotouchfire:
### Proposal Date

19 aug 2022

### Ticket Acceptance Date

25 sept 2022

### Technology Removal Date

~30 Jan 2023 (date subject to change but will be after Olive cutoff)~
August 2023

### First Open edX Named Release Without This Functionality

Quince

### Rationale

The Credentials front end uses an outdated approach to React JS usage, and still uses django templating.

### Removal

Entire front end: https://github.com/openedx/credentials/tree/master/credentials/templates

### Replacement

We are replacing the UI with an MFE: https://github.com/edx/frontend-app-learner-record/

Core functionality will be 1:1. Exceptions include:

- Credit pathway requests UI will be updated to support data model changes discussed in the following ADR: https://github.com/openedx/credentials/pull/1695
- Masquerading will be replaced with a support tool.

pdpinch:
This frontned-app-learner-record MFE hasn't been included in an open release yet. I'm concerned about removing a feature in the same release that the MFE appears for the first time, i.e. olive. 

But, that said, I'm not sure that the program-record UI has been supported in any prior open edX releases.

hurtstotouchfire:
@pdpinch yes, that's understandable, but the impression that I've gotten is that swapping them in one release should be possible and generally fails due to oversights -- and I'd love to pin down what those all are. It sounds like it's largely related to pain around configuration defaults?

pdpinch:
Sorry for the delayed response. BTR has an issue for documenting the requirements, https://github.com/openedx/build-test-release-wg/issues/131 but it doesn't seem complete yet. 

Off the top of my head, some of the hold-ups have been:

- documentation for setup and configuration 
- support for translations 
- ability to deploy to a relative path, instead of to a separate named host (e.g. edx.org/learn instead of learn.edx.org) 
- lingering edX references

hurtstotouchfire:
@arbrandes Are all DEPRs epics? If so that would save me having a dedicated view for them on https://github.com/orgs/openedx/projects/23/

arbrandes:
@hurtstotouchfire, I tagged this one with `epic` for the benefit of the Frontend Development board, which has [a dedicated view](https://github.com/orgs/openedx/projects/29/views/2) for issues labelled as such.

[...]

PR #1796, opened 2022-10-19
docs: Add comments on legacy frontend components/classes that can be removed as part of issue #1722

justinhynes:
[APER-1975]

After releasing the Learner Record MFE we plan on starting removal of the legacy frontend components within the Credentials IDA. This PR just adds comments to the more obvious things which are clear for removal.


justinhynes:
@MaxFrank13, Good point!

Some of the other non-obvious pieces I was worried about was some custom CSS classes and what not, among other things (like... did we install any Python/NPM packages to support this functionality that we can get rid of now??)

PR #1800, opened 2022-10-25
fix: broken styles in program cert print media

Tj-Tracy:
Right now, building static assets is broken for credentials and credentials themes. This PR fixes an issues in the print media query that was blocking the build.

This fix is for the base openedX theme **ONLY** and the edx-next theme will be fixed in credentials themes.

image


credentials/apps/credentials_theme_openedx/static/sass/_print-rendering.scss
```diff
@@ -35,7 +35,8 @@
           .accomplishment-title {
             @include make-col-ready();
             @include make-col(8);
-            @extend %mt-2, %mb-0;
+            margin: 0;
+            margin-bottom: $spacer * 1.5;
```
Tj-Tracy, line 39: I decided to just manually copy these styles over from the utility classes. The mixin approach as not working for these particular classes, and its a very small amount of styling bloat.

credentials/apps/credentials/views.py
```diff
@@ -226,6 +226,7 @@ def get_context_data(self, **kwargs):
 
         context.update(
             {
+                "credential_name": "John Doe",
```
Tj-Tracy, line 229: This is added to give the preview certificate and actual recipient name because it was annoying me.

credentials/apps/credentials_theme_openedx/static/sass/_print-rendering.scss
MaxFrank13, line 39: The previous class was `.mt-2` so should this be `margin-top`? Also being that it is 2 and not 4, should it be `$spacer * .5`?

I'm basing this mostly on the edits to `.signatory` and `.copy` classes.

Tj-Tracy: good catch. Yeah thats a typo.

MaxFrank13 (approved):

PR #1809, opened 2022-11-01
fix: fix team tagging issue causing the Python upgrade automation to throw an error

justinhynes:
The edx-aperture team doesn't seem to exist anymore. I believe it may have been renamed to 2U-aperture which would explain the reason the automation can't add the correct reviewers.


MaxFrank13 (approved):
Looks good!

PR #1834, opened 2022-11-22
chore: Regenerate package lock after Node 16 Upgrade

justinhynes:
[APER-2063]

This PR is for some cleanup after the Node 16 upgrade. The aim here is to update dependencies that will cause issues and to regenerate our `package-lock.json` file in the Credentials repo.

This PR also updates the Credentials repo to use an NPM script to run our linting tests. Previously, we used `gulp` to run our linters, but `gulp` is no longer an active project (no updates since 2019) and doesn't appear to be compatible with later version of Node/NPM.

Some consequences of this PR:
1. We will stop running our (bok choy) acceptance tests suite as part of CI with these changes. They don't run successfully anymore after this upgrade and the bok choy framework has been deprecated for awhile now, there is nothing to gain by fixing them at this time.
2. We will stop running our non-React JS tests during CI because of an issue with Karma + Firefox + Node16 running in containerized GitHub action environments. I will create a followup task to explore using Chromium or another headless browser to run these tests.

Changes:
- regenerate the `package-lock.json` file
- upgrade the `eslint-config-airbnb` and `eslint-plugin-react-hooks` dependencies
- convert linting to be kicked off by NPM script, rather than using our gulpfile
- update Makefile to call NPM script (intstead of Gulp)
- update translation files
- update CI workflows to install specific versions of Node (matching what is deployed to the environment)
- don't run (bok choy) acceptance tests as part of our CI
- do not run non-React JS tests as part of CI as there is an issue with Node16 + Karma + Firefox + GitHub Actions.


.github/workflows/ci.yml
```diff
@@ -7,21 +7,6 @@ on:
     branches: [ master ]
 
 jobs:
-  acceptance_tests:
```
justinhynes, line 10: The BokChoy tests were failing after the Node 16 upgrade. I know that the BokChoy test framework has been deprecated for a bit so I didn't spend any time trying to determine the issue or see if we can fix them.

There aren't many BokChoy tests in Credentials (four total). I figured it was a better use of our time to just not run them during CI.

.github/workflows/ci.yml
```diff
@@ -50,7 +35,7 @@ jobs:
       matrix:
         python-version: ["py38"]
         django-version: ["django32"]
-        node-version: [ 16 ]
+        node-version: [ 16.14.0 ]
```
justinhynes, line 38: Without specifying the exact version, we are installing the latest version (16.18.x at the time of writing this comment). I figured it would be best to install the same version being created via Ansible.

.github/workflows/ci.yml
```diff
@@ -63,10 +48,10 @@ jobs:
           node-version: ${{ matrix.node-version }}
       - name: Install Dependencies
         run: make requirements
-      - name: Install Firefox
+      - name: Install Chromium
```
justinhynes, line 51: I ran into an issue with CI failing to run our non-React JS tests correctly due to an issue with Node16+Karma+Firefox when running tests in a containerized GitHub environment (through GitHub actions). 

Chromium _does_ work though and I thought it would be best to keep as many tests as possible. There will be other changes to facilitate moving from Firefox to Chromium later on.

credentials/static/components/ProgramIcon.jsx
```diff
@@ -1,4 +1,5 @@
 // TODO: We should be able to remove this as part of https://github.com/openedx/credentials/issues/1722
+/* eslint-disable react/function-component-definition */
```
justinhynes, line 2: To complete the upgrade we had to install a newer version of `eslint` and it started yelling at us for this violation. We have tickets/plans to remove this file and I didn't think it was a good use of time to rewrite it to fix the linting violation.

It seemed acceptable/appropriate to ignore this rule in this file.

credentials/static/components/ProgramRecord.jsx
```diff
@@ -302,6 +307,7 @@ class ProgramRecord extends React.Component {
             />
           </div>
           )}
+        {/* eslint-disable react/no-unstable-nested-components */}
```
justinhynes, line 310: See above comment in `ProgramIcon.jsx`. Same thought process here.

credentials/static/components/RecordsHelp.jsx
```diff
@@ -1,4 +1,5 @@
 // TODO: We should be able to remove this as part of https://github.com/openedx/credentials/issues/1722
+/* eslint-disable react/function-component-definition */
```
justinhynes, line 2: See above comment in ProgramIcon.jsx. Same thought process here.

karma.conf.js
```diff
@@ -1,6 +1,8 @@
 const webpackConfig = require('./webpack.config.js');
 delete webpackConfig.entry;
 
+process.env.CHROME_BIN = require('puppeteer').executablePath();
```
justinhynes, line 4: This was added to help configure our move from running JS tests with Firefox to Chromium.

package.json
```diff
@@ -22,6 +23,7 @@
     "js-cookie": "3.0.1",
     "mini-css-extract-plugin": "2.6.0",
     "prop-types": "15.8.1",
+    "puppeteer": "19.2.2",
```
justinhynes, line 26: I am using `puppeteer` (https://pptr.dev/) here alongside Karma and Chromium to run our non-React JS tests.

Makefile
```diff
@@ -117,7 +117,6 @@ tests: ## Run tests and generate coverage report
 	make js-tests
 
 js-tests: ## Run javascript tests
-	$(NODE_BIN)/gulp test
```
justinhynes, line 120: The non-React JS tests were failing to run correctly in CI (and during CI only, it worked locally). I found some reports of others having issues with Karma/Node16/Firefox when running CI tests in GitHub action/containerized environments (which is our configuration too).

I wanted to keep these tests running and tried to convert our Karma config to use Chromium instead of Firefox but ran into other issues (locally but not in CI 😢). 

Long story short, by removing this line we will not be running our vanilla JS tests in CI anymore. I am planning on opening a followup ticket/issue to reinstate these tests.

MaxFrank13 (approved):
Based on all of the context provided and my own investigation, I can't see any issues here. The only thing that sticks out is the JS tests that we are now omitting but since we have it on the radar in an upcoming ticket, I'd say this is good to go 👍

PR #1839, opened 2022-11-29
chore: remove `gulp` from the Credentials IDA

justinhynes:
[APER-2135]

This PR will remove the use and configuration of `gulp` from teh Credentials IDA. The `gulp` package was used in the past in order to configure and manage running our non-React JS tests and JS linters.

The `gulp` project hasn't had any updates since 2019 and looks to be dead. Recently, we have updated our configuration to invoke our linters via an NPM script. Our React tests are already invoked via NPM script, and soon enough our non-React JS tests will also be configured to do the same.

There is no reason to keep the `gulp` dependencies or configuration files in the IDA.


MaxFrank13 (approved):
LGTM 👍

PR #1851, opened 2022-12-06
fix: added Makefile target to run tests with Karma

MaxFrank13:
This adds a new Makefile target that will install Firefox and run the tests using Karma. There is also a PR template that will remind future contributors to run these tests locally. Once a fix is found for getting these tests to run in CI, we will need to remove this PR template.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -119,6 +119,11 @@ tests: ## Run tests and generate coverage report
 js-tests: ## Run javascript tests
 	npm run test-react
 
+test-karma: ## Run JS tests through Karma & install firefox. This command needs to be ran manually in the devstack container before submitting a pull request. It can not be run in CI as of APER-2136.
+	sudo apt-get update
+	sudo apt-get install --no-install-recommends -y firefox xvfb
+	xvfb-run $(NODE_BIN)/karma start
+
 # Requires locally installed software (firefox / xvfb). Easiest to run using `acceptance_tests_suite` command at the bottom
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -119,6 +119,10 @@ tests: ## Run tests and generate coverage report
 js-tests: ## Run javascript tests
 	npm run test-react
 
+test-karma: ## Run JS tests through Karma & install firefox. This command needs to be ran manually in the devstack container before submitting a pull request. It can not be run in CI as of APER-2136.
+	sudo apt-get install --no-install-recommends -y firefox
```
justinhynes, line 123: Just in case, I would also add a line before this to update the package information for all the configured sources.

So, before the `sudo apt-get install...` line, add `sudo apt-get update`

justinhynes: I think in the original CI workflow we were also installing `xvfb`. Maybe we should also include that package to cover our bases?

justinhynes (changes requested):
Good stuff! One requested change before accepting this change. Let me know if you have any questions!

Makefile
MaxFrank13: Good call! The command now installs `xvfb` and then uses `xvfb-run` similar to [how it's done here](https://github.com/openedx/edx-custom-a11y-rules/blob/86bd06bfdaaead4be81a3b46b0833b29add46362/.github/workflows/ci.yml#L34)

justinhynes (approved):

PR #1857, opened 2022-12-07
fix: pin django-ratelimit to versions < 4.x

justinhynes:
Pins the `django-ratelimit` package to versions < 4.x as it introduced a breaking change. This will be re-evaluated and removed as part of APER-2169.


MaxFrank13 (approved):
LGTM! Thanks for updating the outdated ticket names as well

PR #1858, opened 2022-12-07
fix: Web Certificate help documentation hyperlink

jsnwesson:


credentials/apps/edx_credentials_extensions/migrations/0001_initial.py
```diff
@@ -12,7 +12,7 @@ def add_siteconfiguration_data(apps, schema_editor):
         "privacy_policy_url": "https://www.edx.org/edx-privacy-policy",
         "homepage_url": "https://www.edx.org",
         "company_name": "edX Inc.",
-        "certificate_help_url": "https://edx.readthedocs.org/projects/edx-guide-for-students/en/latest/SFD_certificates.html#web-certificates",
+        "certificate_help_url": "https://support.edx.org/hc/articles/207206647-How-are-edX-certificates-delivered-",
```
justinhynes, line 15: I don't think this is where we're going to want to change this. We _definitely_ don't want to change an existing, older migration. These are (normally, but there are exceptions) auto-generated files created by the Django ORM.

In addition, I'm not sure we want to hard-code an `edx.org` specific KB article into the product.

I have a half-baked assumption that this is extracted somewhere in the form of a setting. I'm not 100% off the top of my head though. I'll reach out and we can try and find the right place to make the change.

justinhynes (changes requested):
See my comment, I don't think this is the right place to make this change. We can work together and figure out where the _right_ place is as I'm not sure off the top of my head.

PR #1903, opened 2023-01-18
build: an action to auto-add issues and PRs to a project.

nedbat:
Try out an action to add pull requests and issues to the credentials project.

### DIFF
--- a/.github/workflows/add-to-project.yml
+++ b/.github/workflows/add-to-project.yml
@@ -1,0 +1,33 @@
+name: Add Issues to Project
+
+on:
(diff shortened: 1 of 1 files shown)


feanil (approved):
This looks good to me, it seems like it's worth it to update the org-level re-usable action with this if it works well?

nedbat:
yes, that's the goal.

nedbat:
Looks like it worked: https://github.com/openedx/credentials/actions/runs/3961192901/jobs/6786318152

PR #1908, opened 2023-01-19
build: use the shared action to add new items to project 23

nedbat:
The openedx org has a shared action to add items to projects. Use it to manage the credentials project.

### DIFF
--- a/.github/workflows/add-to-project.yml
+++ b/.github/workflows/add-to-project.yml
@@ -1,4 +1,4 @@
-name: Add Issues to Project
+name: Add New Items to Project
 
 on:
   issues:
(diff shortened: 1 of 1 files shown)


nedbat:
@hurtstotouchfire github board automation.

justinhynes:
@nedbat It looks like there is a conflict the PR is complaining about. Would you be able to resolve this?

Also, I'm confused by why the `add item to project` check is failing. It seems to match what is being run in `credentials-themes` and `frontend-app-learner-record`. Maybe once the conflict is resolved it will be happy??

Otherwise, I couldn't find any difference in what is proposed here compared to the other repo PRs.

nedbat:
I resolved the conflict.

justinhynes (approved):

PR #1931, opened 2023-02-10
fix: add console logging when user shares a program record via pathway

jsnwesson:
[Jira Ticket: APER-2217](https://2u-internal.atlassian.net/browse/APER-2217)
* when users send their program via email, we will now log the internal credential user id, program name + uuid, and the pathway in order to make our bug investigation easier. 
* use get_user_model to safely retrieve User data

- [ ] Run `make test-karma`
- [ ] All tests pass

### DIFF
--- a/credentials/apps/records/views.py
+++ b/credentials/apps/records/views.py
@@ -165,15 +165,21 @@ class ProgramSendView(LoginRequiredMixin, RecordsEnabledMixin, View):
         pathway_id = body["pathway_id"]
         program_uuid = kwargs["uuid"]
 
+        # need to clarify, is it safe to assume the User is the same as request.user?
+        program_details = get_program_details(
+            User,
+            request.site,
+            program_uuid,
+            bool(program_uuid),
+        )
+
         # verify that the user or an admin is making the request
         if username != request.user.get_username() and not request.user.is_staff:
             log.info(f'[Share Program Record] Request made from user {request.user.get_username()} for username {username}')
             return JsonResponse({"error": "Permission denied"}, status=403)
 
-        credential = UserCredential.objects.filter(
-            username=username, status=UserCredential.AWARDED, program_credentials__program_uuid=program_uuid
-        )
-        program = get_object_or_404(Program, uuid=program_uuid, site=request.site)
+        # TODO: get pathway, cert, public record
+        program = program_details.record.program
         pathway = get_object_or_404(
             Pathway,
             id=pathway_id,
@@ -181,11 +187,11 @@ class ProgramSendView(LoginRequiredMixin, RecordsEnabledMixin, View):
             pathway_type=PathwayType.CREDIT.value,
         )
         certificate = get_object_or_404(ProgramCertificate, program_uuid=program_uuid, site=request.site)
-        user = get_object_or_404(User, username=username)
-        preexisting_program_cert_record = ProgramCertRecord.objects.filter(user=user, program=program).exists()
+        user = program_details.record.learner
+        preexisting_program_cert_record = program_details.record.shared_program_record_uuid.exists()
         public_record, _ = ProgramCertRecord.objects.get_or_create(user=user, program=program)
 
-        record_path = reverse("records:public_programs", kwargs={"uuid": public_record.uuid.hex})
+        record_path = reverse("records:public_programs", kwargs={"uuid": public_record.hex})
         record_link = request.build_absolute_uri(record_path)
         csv_link = urllib.parse.urljoin(record_link, "csv")
 
@@ -197,7 +203,7 @@ class ProgramSendView(LoginRequiredMixin, RecordsEnabledMixin, View):
                 "program_name": program.title,
                 "record_link": record_link,
                 "user_full_name": request.user.get_full_name() or request.user.username,
-                "program_completed": credential.exists(),
+                "program_completed": program_details.record.program.completed,
                 "previously_sent": False,
                 "csv_link": csv_link,
             },
@@ -207,6 +213,7 @@ class ProgramSendView(LoginRequiredMixin, RecordsEnabledMixin, View):
         log.info(f'[Share Program Record] Pathway is {pathway.name}')
         ace.send(msg)
 
+        # Need to clarify, is it worth replacing UserCreditPathway with post/patch methods in api.py?
         # Create a record of this email
         if UserCreditPathway.objects.filter(user=user, pathway=pathway, program=program).exists():
             UserCreditPathway.objects.update_or_create(
--- a/credentials/apps/records/views.py
+++ b/credentials/apps/records/views.py
@@ -8,6 +8,7 @@ from uuid import uuid4
 from analytics.client import Client as SegmentClient
 from django import http
 from django.conf import settings
+from django.contrib.auth import get_user_model
 from django.contrib.auth.mixins import AccessMixin, LoginRequiredMixin
 from django.http import HttpResponse, HttpResponseRedirect, JsonResponse
 from django.shortcuts import get_object_or_404
@@ -19,7 +20,6 @@ from edx_ace import Recipient, ace
 from ratelimit.decorators import ratelimit
 
 from credentials.apps.catalog.models import Pathway, Program
-from credentials.apps.core.models import User
 from credentials.apps.core.views import ThemeViewMixin
 from credentials.apps.credentials.models import ProgramCertificate, UserCredential
 from credentials.apps.records.api import get_program_details, get_program_record_data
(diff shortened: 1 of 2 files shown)


credentials/apps/records/views.py
```diff
@@ -90,15 +91,6 @@ def get_context_data(self, **kwargs):
...
 class RecordsView(RecordsListBaseView):
-    # TODO: We should be able to remove this function as part of https://github.com/openedx/credentials/issues/1722
-    def _get_programs(self):
-        return get_user_program_data(
-            self.request.user.username,
-            self.request.site,
-            include_empty_programs=False,
-            include_retired_programs=True,
-        )
-
```
justinhynes, line 101: I don't think we should combine these changes together. I'd prefer a separate PR with changes specifically addressed for this issue/ticket.

jsnwesson: Makes sense! I'll make sure to revert these back.

credentials/apps/records/views.py
```diff
@@ -107,27 +99,6 @@ def get(self, request, *args, **kwargs):
 
         return super().get(request, *args, **kwargs)
 
-    # TODO: We should be able to remove this function as part of https://github.com/openedx/credentials/issues/1722
```
justinhynes, line 110: See comment for `RecordsView`.

credentials/apps/records/views.py
```diff
@@ -163,19 +134,6 @@ def dispatch(self, request, *args, **kwargs):
 class ProgramRecordView(ConditionallyRequireLoginMixin, RecordsEnabledMixin, TemplateView, ThemeViewMixin):
     template_name = "programs.html"
 
-    # TODO: We should be able to remove this function as part of https://github.com/openedx/credentials/issues/1722
```
justinhynes, line 166: See comment for `RecordsView`.

credentials/apps/records/views.py
```diff
@@ -243,26 +165,33 @@ def post(self, request, **kwargs):
...
+        program_details = get_program_details(
+            User,
+            request.site,
+            program_uuid,
+            bool(program_uuid),
+        )
+
         # verify that the user or an admin is making the request
         if username != request.user.get_username() and not request.user.is_staff:
+            log.info(f'[Share Program Record] Request made from user {request.user.get_username()} for username {username}')
```
justinhynes, line 178: The learner's username is PII. Is there anyway we can log the internal Credentials user ID instead of using their username? We can't use `LMS_USER_ID` here from the User model as it's not guaranteed that we have the `LMS_USER_ID` synced in the service. Everyone _should_ have an auto-generated integer ID from the internal `User` table though.

credentials/apps/records/views.py
```diff
@@ -274,13 +203,17 @@ def post(self, request, **kwargs):
...
                 "user_full_name": request.user.get_full_name() or request.user.username,
-                "program_completed": credential.exists(),
+                "program_completed": program_details.record.program.completed,
                 "previously_sent": False,
                 "csv_link": csv_link,
             },
         )
+        log.info(f'[Share Program Record] User information sent is either full_name or username {request.user.get_full_name() or request.user.username}')
+        log.info(f'[Share Program Record] Program Record is {program.title} with UUID {program_uuid}')
+        log.info(f'[Share Program Record] Pathway is {pathway.name}')
```
justinhynes, line 213: If possible, can we combine this into one single comprehensive log statement? Having them be three separate logs statements introduces a chance at confusion if multiple learners are invoking a pathway at once

justinhynes: Also I have a similar request around using the internal user id instead of full name or username to limit PII appearing in logs. Thanks!

credentials/apps/records/views.py
```diff
@@ -243,26 +165,33 @@ def post(self, request, **kwargs):
...
             Pathway,
             id=pathway_id,
             programs__uuid=program_uuid,
             pathway_type=PathwayType.CREDIT.value,
         )
         certificate = get_object_or_404(ProgramCertificate, program_uuid=program_uuid, site=request.site)
-        user = get_object_or_404(User, username=username)
-        preexisting_program_cert_record = ProgramCertRecord.objects.filter(user=user, program=program).exists()
+        user = program_details.record.learner
+        preexisting_program_cert_record = program_details.record.shared_program_record_uuid.exists()
```
justinhynes, line 191: `.exists()` is a model/ORM function and I'm not sure if it works in this capacity. I almost expected this to throw an exception. Did this work when you tested it?

credentials/apps/records/views.py
```diff
@@ -274,13 +203,17 @@ def post(self, request, **kwargs):
...
                 "previously_sent": False,
                 "csv_link": csv_link,
             },
         )
+        log.info(f'[Share Program Record] User information sent is either full_name or username {request.user.get_full_name() or request.user.username}')
+        log.info(f'[Share Program Record] Program Record is {program.title} with UUID {program_uuid}')
+        log.info(f'[Share Program Record] Pathway is {pathway.name}')
         ace.send(msg)
 
+        # Need to clarify, is it worth replacing UserCreditPathway with post/patch methods in api.py?
```
justinhynes, line 216: I'm replying to your comment based on an assumption around the context of the comment itself. 

The functions contained in an `api.py` file should be _Python_ APIs. Essentially, the functions contained in a Django apps `api.py` should be _public_ functions that are exposed for other Django apps to import and use. They aren't intended to have anything in relation to a _REST_ API, which would support post/patch functions related to a network endpoint. 

Hopefully that makes sense.

Additionally, it isn't a big deal to use UserCreditPathway here, as this is a model that belongs to the `records` app. What we _don't_ want to do is directly import and use models from _another_ app (which is how it exists today... the `Pathway` and `Program` models are imported from the _catalog_ app and are being used directly in the `records` app.

credentials/apps/records/views.py
```diff
@@ -243,26 +165,33 @@ def post(self, request, **kwargs):
         pathway_id = body["pathway_id"]
         program_uuid = kwargs["uuid"]
 
+        # need to clarify, is it safe to assume the User is the same as request.user?
+        program_details = get_program_details(
+            User,
+            request.site,
+            program_uuid,
+            bool(program_uuid),
+        )
```
justinhynes, line 174: A couple of points here...

> # need to clarify, is it safe to assume the User is the same as request.user?
🤔  To be safe, I think we'd likely want to just use the user specified in the request body/data. In the past, this request checks to see if it originated from the user themselves _or_ an admin/staff user from 2U/edX.

Second, this isn't going to work as you may expect. You're passing in the `User` _model_ to `get_program_details` but `get_program_details` is expecting a User _instance_. 

We need to _retrieve_ the user instance before we can call this Python function. You can retrieve the instance by using the Django ORM: 
```
user = User.objects.get(username=username)
```

Third, I'm not sure we're using `get_program_details` as intended as its coupled pretty tightly with the ProgramRecordView view/class. 
------------------------------------------

OK. So, my suggestion here...

I don't think we need to use `get_program_details` here. I'd rather see us create new _Python_ API functions in the `catalog` app that support retrieval of a Pathway or Program data as needed so we can remove the import of these models from another Django app. 

Final thought. My suggestions for creating these new public Python APIs and removing the import of cross-app models are purely a "since we'll be in here" change. It doesn't fix or solve anything other than making the codebase better than it was before and bringing it in line with best practices. 

If you think this is going to be too much work right now, we can add the logging without changing anything else. Let me know what you think

credentials/apps/records/views.py
justinhynes: Can we remove this line? I'd prefer to remove it over logging PII.

credentials/apps/records/views.py
```diff
@@ -279,6 +280,8 @@ def post(self, request, **kwargs):
                 "csv_link": csv_link,
             },
         )
+
+        log.info(f'[Share Program Record] Internal Credentials user id [{user.id}] is enrolled in program [{program.title}] with UUID [{program_uuid}]. Pathway is [{pathway.name}]')
```
justinhynes, line 284: Instead of logging the pathway name, can we log the ID/UUID? A name _could_ change, but the other values shouldn't and ID/UUID is easier to query for in our data. Additionally, we already have access to the `pathway_id` as part of the request too.

Nit: Wordcrafting suggestion...
```
log.info(f'[Share Program Record] Internal Credentials User [{user.id}] is sharing their progress in program [{program_uuid}] with pathway [{pathway_id}]')
```

credentials/apps/records/views.py
jsnwesson: Sure can! Initially I wondered if we needed to log this occurence but the JSON response providing an error message of "Permission denied" should suffice.

justinhynes (approved):

PR #1971, opened 2023-03-21
feat: upgraded to node v18, added .nvmrc and updated workflows

Mashal-m:


justinhynes:
Hi @Mashal-m -- just checking on this PR, which has been a draft since March. Is this ready for review?

PR #1981, opened 2023-04-10
fix: Don't encode to bytes for CSV filename

timmc-edx:
This was producing filenames like `b'something'.csv` rather than just `something.csv`. The test didn't catch it because it just repeated the same code it was testing. There's randomness in the test data, so this uses a regex to get as close as possible to an equality assertion.

- [x] Run `make test-karma`
- [x] All tests pass


justinhynes (approved):
Thanks for this fix! Sorry it got lost in the backlog of other PRs in this repo.

PR #1986, opened 2023-04-13
build: use GHA for running code coverage

justinhynes:
We have to change our approach for collecting code coverage statistics after the (deprecated) codecov package was yanked from pypi.

- [ ] Run `make test-karma`
- [ ] All tests pass


MaxFrank13 (approved):
LGTM

PR #1988, opened 2023-04-18
feat: Course completion endpoint

cdeery:
This is an API to show the status for a learner of their course certificates.

- [ ] Run `make test-karma`
- [ ] All tests pass


credentials/apps/credentials/rest_api/v1/permissions.py
```diff
@@ -0,0 +1,54 @@
+"""
+Custom permissions classes for use with DRF.
+"""
+from django.http import Http404
+from rest_framework import permissions
+
+
+class UserCredentialViewSetPermissions(permissions.DjangoModelPermissions):
```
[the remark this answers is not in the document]
cdeery: No, that was just a placeholder. The actual permissions will be forthcoming shortly

credentials/apps/credentials/rest_api/v1/views.py
```diff
@@ -0,0 +1,133 @@
...
+from credentials.apps.core.models import User
+from credentials.apps.records.models import UserGrade
+from credentials.apps.records.utils import get_credentials
+
+log = logging.getLogger(__name__)
+
+class LearnerCertificateStatusView(APIView):
+
+
+    permission_classes = (permissions.IsAdminUser,)
```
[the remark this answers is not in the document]
cdeery: Yes, that will be in the next push.

credentials/settings/base.py
```diff
@@ -458,6 +458,7 @@
 # END DJANGO DEBUG TOOLBAR CONFIGURATION
 
 USERNAME_REPLACEMENT_WORKER = "replace with valid username"
+LEARNER_STATUS_WORKER = "replace with valid username"
```
justinhynes, line 461: nit: Is this worker tied to an IDA by any chance? Is this request coming from the Subscriptions IDA? If so, maybe we can make a stronger association of the name of the worker to the source of the request? (`SUBSCRIPTIONS_WORKER`?)

credentials/apps/credentials/rest_api/v1/views.py
```diff
@@ -0,0 +1,129 @@
...
+
+        try:
+            if username:
+                user = User.objects.get(username=username)
+                lms_user_id = user.lms_user_id
+            else:
+                user = User.objects.get(lms_user_id=lms_user_id)
+                username = user.username
+        except User.DoesNotExist:
+            # we don't have the user in the system
```
justinhynes, line 95: Do you think it would be useful to log something here in this case?

credentials/apps/credentials/rest_api/v1/views.py
```diff
@@ -0,0 +1,129 @@
...
+        CanGetLearnerStatus,
+    )
+
+    def post(self, request):
+        """
+        **POST Parameters**
+
+        A POST request must include one of "lms_user_id" or "username",
+        and it can have one or both of a list of course
+        uuids and/or a program uuid.
```
justinhynes, line 35: General question: According to this comment we _could_ receive a Program UUID. What happens in that case? Nothing? It seems like we're only returning information around the learner's statuses in their courses. 

I see that we're getting back some program credential information from our call to `get_credentials` (on line 101) but we're not doing anything with `program_credentials` returned. Is this future work? Fast follow? Or a mistake in the docstring?

credentials/apps/credentials/rest_api/v1/views.py
```diff
@@ -0,0 +1,129 @@
...
+
+from edx_rest_framework_extensions.auth.jwt.authentication import JwtAuthentication
+from rest_framework import permissions, status
+from rest_framework.authentication import SessionAuthentication
+from rest_framework.response import Response
+from rest_framework.views import APIView
+
+from credentials.apps.core.models import User
+from credentials.apps.credentials.rest_api.v1.permissions import CanGetLearnerStatus
+from credentials.apps.records.models import UserGrade
```
justinhynes, line 11: We really shouldn't be importing models from other apps directly. If there is a requirement to retrieve grade information we should use (or create) a Python API function from the `records` Django app.

Utility functions that live inside a Django app's `api.py` file are considered public and exposed for general use across Django apps.

credentials/apps/credentials/rest_api/v1/views.py
```diff
@@ -0,0 +1,129 @@
...
+from credentials.apps.records.utils import get_credentials
```
justinhynes, line 12: Similarly, we shouldn't be importing and using utility functions from another Django app unless they are inside of the app's `api.py` file. If we need to call `get_credentials` from another Django app, we should have a "public" wrapper for the utility in the app's `api.py` file.

justinhynes (changes requested):
Could we try and reduce/remove the use of cross-app Model imports and internal utility functions? TIA! Let me know if you have any questions.

credentials/apps/credentials/rest_api/v1/views.py
cdeery: That is planned for a fast follow- I will spell it out more in the ADR, but being able to sanity check a learner's status against a particular program will help Aperture and support.

credentials/apps/credentials/rest_api/v1/views.py
justinhynes: (optionally) we could just move the function to the public `api.py` too, but I'm unsure how many references would need to be updated at that point

credentials/apps/credentials/rest_api/v1/views.py
cdeery: Yes, that's a good idea. Will do

credentials/apps/credentials/rest_api/v1/views.py
```diff
@@ -0,0 +1,129 @@
+import logging
+
+from edx_rest_framework_extensions.auth.jwt.authentication import JwtAuthentication
+from rest_framework import permissions, status
+from rest_framework.authentication import SessionAuthentication
+from rest_framework.response import Response
+from rest_framework.views import APIView
+
+from credentials.apps.core.models import User
```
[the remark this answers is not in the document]
cdeery: Thanks. Will do.

credentials/settings/base.py
cdeery: I think it's better to have the user identified by the right they have rather than the current job, as the API may be used for other use cases in the future. I'm open to hashing out better names.

credentials/apps/credentials/rest_api/v1/views.py
cdeery: Thanks. Refactored code to eliminate cross-app includes

credentials/apps/credentials/rest_api/v1/views.py
cdeery: I created an internal API to handle the private information.

cdeery:
@shafqatfarhan do you have any additional comments?

credentials/apps/credentials/rest_api/v1/views.py
justinhynes: Thanks!

justinhynes (approved):
Thanks for making the suggested changes!

PR #2002, opened 2023-05-10
fix: pinning `urllib3` package to fix update requirements workflow

justinhynes:
We need to add a new constraint as `boto` is looking for versions of `urllib3` < 2.x.

- [ ] Run `make test-karma`
- [ ] All tests pass


cdeery (approved):
Looks good

PR #2008, opened 2023-05-16
feat: Management command to sync LMS user ids

cdeery:
This PR creates a management command that loops over all of the users in Credentials that don't have LMS user ids, fetches them from the LMS, and updates them in Credentials.
This will backfill all of the missing IDs that predate the update to collect LMS ids when data is pushed to Credentials.

- [ ] Run `make test-karma`
- [ ] All tests pass

### DIFF
--- a/credentials/apps/core/management/commands/sync_ids_from_platform.py
+++ b/credentials/apps/core/management/commands/sync_ids_from_platform.py
@@ -57,10 +57,6 @@ class Command(BaseCommand):
         if 0 == limit:
             count = num_users_to_update
 
-        if count == 0:
-            logger.warning("No users to update. Stopping")
-            return
-
         logger.warning(f"Start processing {count} IDs with no lms_user_id")
 
         # loop over users in batches
@@ -72,10 +68,7 @@ class Command(BaseCommand):
             for user in curr_users:
                 # did the endpoint return a value for this user?
                 if user.username in ids:
-                    try:
-                        self.update_user(user, ids[user.username], verbosity, dry_run)
-                    except RuntimeError as err:
-                        logger.error(f"Error occurred when updating lms_user_id for user {user.username}: {err}")
+                    self.update_user(user, ids[user.username], verbosity, dry_run)
                 else:
                     logger.error(f"Could not get lms_user_id for user {user.username}")
             if x + slice_size < count:
--- a/credentials/apps/core/management/commands/sync_ids_from_platform.py
+++ b/credentials/apps/core/management/commands/sync_ids_from_platform.py
@@ -8,83 +8,90 @@ from urllib.parse import urljoin
 
 from django.contrib.auth import get_user_model
 from django.core.management.base import BaseCommand
-from credentials.apps.core.models import SiteConfiguration
 
-User = get_user_model()
+from credentials.apps.core.models import Site, SiteConfiguration
+
+
 logger = logging.getLogger(__name__)
+User = get_user_model()
 
 site_configs = SiteConfiguration.objects.first()
 
-class Command(BaseCommand):
 
+class Command(BaseCommand):
     def add_arguments(self, parser):
+        parser.add_argument("--batch_size", type=int, default=10, help="Number of IDs to process at a time. Default 10")
         parser.add_argument(
-            "--batch_size", 
-            type=int, 
-            default=10, 
-            help="Number of IDs to process at a time. Default 10")
+            "--pause_ms", type=int, default=1, help="Number of seconds to pause between calls. Default 1 sec"
+        )
         parser.add_argument(
-            "--pause_ms", 
-            type=int, 
-            default=1, 
-            help="Number of seconds to pause between calls. Default 1 sec")
-        parser.add_argument(
-            "--limit", 
-            type=int, 
-            default=0, 
-            help="Total number of IDs to update. 0 for update all, Default is 0 (all)")
+            "--limit", type=int, default=100, help="Total number of IDs to update. 0 for update all, Default is 100"
+        )
+        parser.add_argument("--verbose", action="store_true", help="Log each update")
+        parser.add_argument("--site", type=str, default=0, help="Domain of the site to query for lms_user_id's")
 
     def handle(self, *args, **options):
         """
-            Get batches of user info from accounts and update the lms_user_id
-            for users who are missing it.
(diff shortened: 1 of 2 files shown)


credentials/apps/core/management/commands/sync_ids_from_platform.py
```diff
@@ -0,0 +1,103 @@
...
+            help="Number of IDs to process at a time. Default 10")
+        parser.add_argument(
+            "--pause_ms", 
+            type=int, 
+            default=1, 
+            help="Number of seconds to pause between calls. Default 1 sec")
+        parser.add_argument(
+            "--limit", 
+            type=int, 
+            default=0,
```
justinhynes, line 34: Should we default to some small number just in case? 10? 50? 100? I'm fine with '0' defaulting to "all users" but I'm not sure if the default value should try to process them all at once.

credentials/apps/core/management/commands/sync_ids_from_platform.py
```diff
@@ -0,0 +1,103 @@
...
+        
+
+    def get_lms_ids(self, users):
+        """
+        given a list of users, query platform and get lms_user_ids
+        return a dict of lms_user_ids keyed by username
+        """
+        user_dict = dict()
+        # create a comma separated string with the usernames
+        user_name_list = ','.join(user.username for user in users)
```
justinhynes, line 86: nit: this is really building a query parameter string, right? not an actual list/collection? maybe we can reflect that in the name?

justinhynes:
Any chance we could add a unit test for the management command?

credentials/apps/core/management/commands/sync_ids_from_platform.py
```diff
@@ -0,0 +1,103 @@
...
+                if user.username in ids.keys():
+                    try:
+                        self.update_user(user, ids[user.username])
+                    except:
+                        logger.error(f"Error occurred when updating lms_user_id for user {user.username}")
+                else:
+                    logger.error(f"Could not get lms_user_id for user {user.username}")
+            if x + slice_size < count:
+                time.sleep(pause)
+        else:
```
justinhynes, line 75: I didn't know for/else was a supported thing in Python 🤯. TIL

credentials/apps/core/management/commands/sync_ids_from_platform.py
```diff
@@ -0,0 +1,103 @@
...
+            Get batches of user info from accounts and update the lms_user_id
+            for users who are missing it.
+        """
+        users_without_lms_id = User.objects.filter(lms_user_id=None)
+        num_users_to_update = users_without_lms_id.count()
+        offset = options.get("batch_size")
+        pause = options.get("pause_ms")
+        count = options.get("limit")
+        # Get the actual number of users to process, an arg of 0 means all of them
+        count = min(count, num_users_to_update)
```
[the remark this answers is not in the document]
cdeery: That immediately made it more clear. Thanks for saving myself from trying to be too clever!

credentials/apps/core/management/commands/sync_ids_from_platform.py
```diff
@@ -0,0 +1,103 @@
...
+from urllib.parse import urljoin
+
+from django.contrib.auth import get_user_model
+from django.core.management.base import BaseCommand
+from credentials.apps.core.models import SiteConfiguration
+
+User = get_user_model()
+logger = logging.getLogger(__name__)
+
+site_configs = SiteConfiguration.objects.first()
```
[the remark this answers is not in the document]
cdeery: I added in the option for specifying a site. I think this is rare, and running it would be a pain because you have to do each site separately, but it should do the trick.

credentials/apps/core/management/commands/sync_ids_from_platform.py
cdeery: Good call. I made it 100. Probably not good to run 6 million updates by accident.

credentials/apps/core/management/commands/sync_ids_from_platform.py
```diff
@@ -0,0 +1,103 @@
...
+        else:
+            logger.error(f" {user_url} returned status {user_response.status_code}")
+
+        return user_dict    
+
+    def update_user(self, user, id):
+        """ update the lms_user_id for the user """
+        if id and id > 0:
+            user.lms_user_id = id
+            user.save(update_fields=['lms_user_id'])
```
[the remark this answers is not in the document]
cdeery: I added a comment so I remember what it does the next time I look at it myself.

credentials/apps/core/management/commands/sync_ids_from_platform.py
cdeery: renamed.

credentials/apps/core/management/commands/sync_ids_from_platform.py
cdeery: obscure, but useful.

credentials/apps/core/management/commands/sync_ids_from_platform.py
```diff
@@ -0,0 +1,113 @@
...
+        limit = options.get("limit")
+        verbosity = options.get("verbose")
+        site_domain = options.get("site")
+
+        # if there are multiple sites managing different users
+        # This is likely rare
+        if site_domain:
+            logger.info(f"using {site_domain} for user queries")
+            site = Site.objects.filter(domain=site_domain)
+            globals()["site_configs"] = SiteConfiguration.objects.filter(site=site)
```
justinhynes, line 51: Why do we need to call `globals()` here? If we're passing in the ID of the site to the management command, can we just do `SiteConfiguration.objects.get(id={some-id-passed-in}`? Then we would delete line 18 as well.

justinhynes:
Just had another thought after our demo. Since we can't easily pass in arguments to the management command in production, we would likely need to use a configuration model (similar to how the `notify_credentials` mgmt command gets its arguments).

credentials/apps/core/management/commands/sync_ids_from_platform.py
cdeery: refactored/cleaned this up

justinhynes (approved):
Thanks for the updates!

justinhynes (approved):
Thanks for the updates!

PR #2014, opened 2023-05-24
fix: unpin edx-event-bus-kafka and fix breaking changes

justinhynes:
[APER-2438]

- unpin `edx-event-bus-kafka` (ebk) and update dependencies to pull in latest version of ebk
- remove the local `consume_events.py` as it is no longer needed
- adds `openedx_events` as a third party app to Credentials

- [ ] Run `make test-karma`
- [ ] All tests pass


MaxFrank13 (approved):
Thanks for taking care of the updates! 👍

PR #2019, opened 2023-06-01
feat: add course_runs option in learner status API

cdeery:
Add an optional section to the input payload for the learner status API which can provide a list of course-run UUIDS.
The API will handle either courses, course-runs, or a combination of both. (or none, but it won't do anything useful)

### DIFF
--- a/credentials/apps/credentials/rest_api/v1/views.py
+++ b/credentials/apps/credentials/rest_api/v1/views.py
@@ -30,14 +30,19 @@ class LearnerCertificateStatusView(APIView):
         **POST Parameters**
 
         A POST request must include one of "lms_user_id" or "username",
-        and a list of course uuids
+        and a list of course uuids, course_runs, or a mix of both.
         (or a program uuid, in a future version)
 
         {
             "lms_user_id": <lms_id>,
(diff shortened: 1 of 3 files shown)


credentials/apps/credentials/rest_api/v1/views.py
```diff
@@ -30,14 +30,19 @@ def post(self, request):
...
             "courses": [
-                "uuid1",
-                "uuid2"
+                "course_uuid1",
+                "course_uuid2"
+                ...
+            ],
+            "course_runs": [
+                "course_run_uuid1",
+                "course_run_uuid2",
```
justinhynes, line 45: Would we be using a course run UUID here? My assumption here is that we would be given a course-run key. In the requested update from Ponderosa, they mentioned they wanted to pass `course_run_id` which I _assume_ is a course-run key, not a UUID.

cdeery: Yes, the request was ambiguous. That makes more sense. I have updated to code to accept either a UUID or a course-run-key

shafqatfarhan (approved):
LGTM 👍

PR #2034, opened 2023-06-16
fix: correct use of `update_or_create` when updating existing credentials via event bus

justinhynes:
[APER-2505]

This PR corrects an error in the `_update_or_create_credential` function, which is invoked when processing `CERTIFICATE_CREATED` events from the event bus.

I had set the search criteria too strictly and didn't properly pass in a `defaults` dict with the data that should be updated if a match is found. This was resulting in a database IntegrityError because the ORM wasn't finding existing entities correctly and attempting to insert a new record that violated constraints on the table.

- [x] Run `make test-karma`
- [x] All tests pass


MaxFrank13 (approved):
LGTM! Thanks for the explanation

PR #2050, opened 2023-06-22
fix: broken swagger and duplicate route

deborahgu:
This API contains a non-CRUD class with no serializer, which the
swagger conversion can't handle automagically. Additionally, there was
an endpoint that had been imported twice because of a file structure
containing multiple `urls.py` files.

* specifies the swagger schema for a non-serializer class
* removes redunandant route via wrong endpoint

Note that the generated swagger document is somewhat confusing, and does not
correctly group the different APIs. This is unavoidable because of two
separate complicating factors:

* the openAPI specification, which cannot has certain assumptions about base prefix
* changes within this cluster of APIs, in which endpoint structure has had three
  distinct layouts, before we reached the current, edx-standard structure.
  (`appname/api/v#/endpoint_name`, `api/v#/endpoint_name`, and
  `api/appname/v#/endpoint_name`)

There are ways to address this, but they are all messy (e.g. creating aliases for old endpoints).

Fixes: APER-2531

- [x] Run `make test-karma`
- [x] All tests pass

### DIFF
--- a/credentials/apps/credentials/rest_api/v1/views.py
+++ b/credentials/apps/credentials/rest_api/v1/views.py
@@ -27,66 +27,60 @@ class LearnerCertificateStatusView(APIView):
         CanGetLearnerStatus,
     )
 
-    lms_user_id_schema = openapi.Schema(
-        type=openapi.TYPE_STRING,
-        description='lms_user_id as a string')
+    lms_user_id_schema = openapi.Schema(type=openapi.TYPE_STRING, description="lms_user_id as a string")
 
-    username_schema = openapi.Schema(
-        type=openapi.TYPE_STRING,
-        description='username')
+    username_schema = openapi.Schema(type=openapi.TYPE_STRING, description="username")
 
     per_course_grade_schema = openapi.Schema(
         type=openapi.TYPE_OBJECT,
         properties={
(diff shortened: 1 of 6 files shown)


credentials/apps/credentials/rest_api/v1/views.py
```diff
@@ -25,68 +27,77 @@ class LearnerCertificateStatusView(APIView):
         CanGetLearnerStatus,
     )
 
+    lms_user_id_schema = openapi.Schema(type=openapi.TYPE_STRING, description="lms_user_id as a string")
```
justinhynes, line 30: nit: I don't believe this is a string. The `lms_user_id` is an auto-generated integer. 

If I try to use this via Swagger without making `lms_user_id` a String it will throw an error. However, if you try exercising the API through DRF, this is totally valid data: 
```
{
  "lms_user_id": 3,
  "course_runs": [
    "course-v1:edX+magnets101+2T2023"
  ]
}
```

If you pass it as a String, it does the right thing though. I expect most times we won't be _receiving_ it as a String, and I have a tiny concern about confusing people who may try to interact with the API via Swagger.

justinhynes (approved):
One tiny comment but overall looks good to me! Thanks for fixing this up.

PR #2053, opened 2023-06-23
fix: pin mysqlclient version

MaxFrank13:
Pinning version of `mysqlclient` to v2.1.1. See [APER-2556](https://2u-internal.atlassian.net/browse/APER-2556) for more details and plan for follow up.

- [ ] Run `make test-karma`
- [ ] All tests pass


justinhynes (approved):
Thanks for adding the constraint!

PR #2058, opened 2023-06-30
feat: change behavior for unknown username, userID

deborahgu:
If an unknown username or lms_user_id is sent, might be real in the LMS and unknown to credentials (i.e. a student who has never received a credential), or it might be fake. Either way, return a 200 response with an empty status list, rather than reporting whether or not the user exists.

Also:

* Fixed this information in both the openAPI and the ReST docs
* improved the formatting/display of the docstring in POST so it will work better in the generated swagger
* Added some Python typing
* fixed some grammar bugs in the ReST
* clarified that username/user ID is exactly one option in the ReST
* removed some commented out code coverage experiments, with the approval of cdeery

FIXES: APER-2425

- [x] Run `make test-karma`
- [x] All tests pass


docs/credentials_api.rst
```diff
@@ -210,15 +210,18 @@ You can filter the returned list of credentials by using the ``program_uuid`` pa
 Query for a user's earned certificates for specific courses
 -----------------------------------------------------------
 
-Query for an individual's users earned certificates for a list of courses or course runs.
+Query for an individuals user's earned certificates for a list of courses or course runs.
```
justinhynes, line 213: Nit: Should this be `individual learner's`?

justinhynes (approved):
Thanks for making those changes!

PR #2060, opened 2023-07-03
fix: cache updated to pymemcache

zubairshakoorarbisoft:

- [ ] Run `make test-karma`
- [ ] All tests pass


justinhynes:
The update to `PyMemcache` has already been completed. I don't believe this PR is needed any longer and I am going to close it. Please re-open and let us know if this is incorrect.

PR #2087, opened 2023-07-18
refactor: remove learner record from Credentials

jsnwesson:
* Learner Record MFE is now public and replaces the LR functionality that existed in Credentials.

- [ ] Run `make test-karma`
- [ ] All tests pass


credentials/static/sass/_layouts.scss
```diff
@@ -148,7 +148,7 @@
 
 .record-card {
   margin-bottom: 0;
-  background: $stripe-grey;
+  background: #F9FAFC;
```
jsnwesson, line 151: I decided to replace the variables in this file with their actual value, as the variables were originally used in now obsolete files, and the remaining use of those variables occurred once each in this file.

PR #2104, opened 2023-07-26
refactor!: remove Program Record UIs frontend components

jsnwesson:
### Context
The Learner's Record view has been extracted into the Learner Record MFE, and so the record view that exists in Credentials can be safely removed.

https://github.com/openedx/credentials/issues/1722

### What this removes
Program Record
Record List
Masquerade Banner 

- [ ] Run `make test-karma`
- [ ] All tests pass


justinhynes:
Thanks so much for all your work on this! I have some questions for you, which I think may lead to some more code deletion. 

1. Can we also remove some of the sass styling for components we removed? For example, we still have `credentials/static/views/_masquerade-banner.scss` in the code. I don't think this would be used by anything anymore. I think there are other sass classes we could clean up. 
2. Are there any Django templates that can be removed now? For example, `_masquerade.html` comes to mind as a removal candidate. 
3. Have you checked if there are any other utility functions/Python code we can delete? Things like `masquerading_authorized` came to mind from `credentials/credentials/apps/records/utils.py`. 
4. Have we checked to see if there are any Python requirements that are no longer in use? (`django-hijack` comes to mind)
5. Have we checked to see if there are any JS requirements that are no longer in use? (`react` comes to mind?)

jsnwesson:
Ach, yeah, a lot of those are going to take place in the follow-up PR that's pretty much ready to be shared out once this is approved. https://github.com/openedx/credentials/pull/2106

Originally, I wanted this PR to really only focus on React components and their test counterparts because that was already a lot of files to look over (but really the majority of files were translations). If it makes more sense to move some of the code deletions into this PR (like the SCSS, HTML, and React deletion), then I can do that!

justinhynes (approved):

justinhynes:
Thanks, @jsnwesson. Makes sense. Glad to see that we're going to continue to delete code after this one lands :D

PR #2106, opened 2023-07-27
Jwesson/remove remaining program record UI files

jsnwesson:
### Context
**This is Part 2 of an effort to remove Learner Record. See below for Part 1** 

The Learner's Record view has been extracted into the Learner Record MFE, and so the record view that exists in Credentials can be safely removed.

Issue that this came from: #1722
Part 1 for this ticket: #2104

### What this removes
Django configurations for Program Record
React setup
Jest setup

- [ ] Run `make test-karma`
- [ ] All tests pass


credentials/static/sass/_layouts.scss
```diff
@@ -148,7 +148,7 @@
 
 .record-card {
   margin-bottom: 0;
-  background: $stripe-grey;
+  background: #F9FAFC;
```
justinhynes, line 151: I'm curious what lead to these updates?

justinhynes:
Since we have removed the `django-hijack` package, can we also ensure to run `make upgrade` to regenerate the requirements files? This should remove the package from our autogenerated *.txt requirements files.

Also, could we remove the constraint for `django-hijack` in `constraints.txt`?

credentials/static/sass/_layouts.scss
jsnwesson: There was a handful of variables stored in `credentials/static/sass/views/_records-list.scss` that were referenced in this file, and rather than keeping those variable names -- as they would've only be used once here -- I used the original value.

justinhynes: Oh, nice! Thanks for the update :)

justinhynes (approved):
Thanks for all the work on this! Can we just ensure that we squash before merging? :joy:

jsnwesson:
Haha! Absolutely. So much to clean.

PR #2107, opened 2023-07-28
chore: unpin urllib3

justinhynes:
[APER-2422]

Unpin `urllib3` now that there are no conflicts with boto.

- [ ] Run `make test-karma`
- [ ] All tests pass


justinhynes:
I was surprised that unpinning a constraint caused us to pull an older version of our requirements. I am going to recheck things on my end before putting this up for review.

PR #2167, opened 2023-09-06
fix: unpin social auth

justinhynes:
Temporarily unpinning `social-auth-app-django` so we can attempt to reverse the migration adding the `extra_data_new` column.

- [ ] Run `make test-karma`
- [ ] All tests pass


matthugs (approved):
w/e my way didn't immediately work, this seems fine

PR #2177, opened 2023-09-12
Add documentation for Event Bus capabilities in Credentials

justinhynes:
[APER-2632]

This PR adds some documentation and diagrams for the event bus functionality that has been added to the Credentials IDA over the past few months.

- [x] Run `make test-karma`
- [x] All tests pass


docs/event_bus.rst
```diff
@@ -0,0 +1,46 @@
+Event Bus
```
justinhynes, line 1: To view as it would be rendered in a browser, see: https://github.com/openedx/credentials/blob/f843a9faebdb5d11f6d04c4077c1d810f9517c75/docs/event_bus.rst

docs/credentials_api.rst
```diff
@@ -42,7 +42,7 @@ Create a New Credential
 * The ``username`` and ``credential`` parameters are required.
```
justinhynes, line 42: VS code complained that this file had non-standard line terminators, so I let it do its thing.

deborahgu (approved):

PR #2183, opened 2023-09-13
feat: make catalog model fields read only in Django Admin

justinhynes:
[APER-2841](https://2u-internal.atlassian.net/browse/APER-2841)

This PR updates the Catalog app's models available via Django Admin of the Credentials IDA to be read-only. We are making this change becuase Credentials is not the source of truth for this data, it is synced from the Course-Discovery IDA.

If the data needs to be updated, it should be done in Discovery and then admins should use the `copy-catalog` management command to sync the data to Credentials.

This is being done to ensure data integrity across services.

- [ ] Run `make test-karma`
- [ ] All tests pass


jsnwesson (approved):

jsnwesson:
LGTM!

PR #2185, opened 2023-09-14
feat: enforce CourseCertificate course_run, course_id presence

deborahgu:
Will require thorough manual QA.

* on CourseCertificate save:
    * create course_run from course ID if needed
    * create course ID from course_run if needed
* refactored the CredentialField serializer to be more efficient given the change
* In most cases of direct object reference, switched from course_id to course_run.key
* Fixed a couple of bad tests that confused thought course_id should reference course_run.course.id (alas for legacy name issues)
* Cleaned up a few log messages (log levels, text, clarity, string formatting)
* marked some tests as skip
    * Given the enforced save method, making objects broken enough to prove the test cases would require mocking so extreme as to make the tests practically useless
    * The enforced data validation is tested elsewhere, by the rest of the test suite

Additionally:

* Added type checking, not ubiquitously, but wherever it helped me debug
* Added a few explanatory comments where it helped me debug
* let black reformat a couple of files that I touched but otherwise didn't change

Tickets to be created for future work:

* Make course_run non-nullable
    * data migration to fix missing course_run for anyone who hasn't run populate_missing_courserun_info
    * schema migration
    * raise helpful error on illegal creation
* (Optional) completely strip out course_id
    * this one may not be worth it; as long as the fields are forced to be identical and present, it's fine to have a convenience property

FIXES: APER-1136

- [ ] Run `make test-karma`
- [ ] All tests pass

### DIFF
--- a/credentials/apps/api/v2/serializers.py
+++ b/credentials/apps/api/v2/serializers.py
@@ -78,8 +78,8 @@ class CredentialField(serializers.Field):
             raise ValidationError({"course_run_key": msg})
         else:
             msg = (
-                "CourseCertificate failed to create because the CourseRun %s doesn't exist in the catalog",
-                course_run_key,
+                "CourseCertificate failed to create because the CourseRun %s doesn't exist in the catalog"
+                % course_run_key
             )
             raise ValidationError({"course_run_key": msg})
 
--- a/credentials/apps/api/v2/serializers.py
+++ b/credentials/apps/api/v2/serializers.py
@@ -2,12 +2,14 @@
 Serializers for data manipulated by the credentials service APIs.
 """
 import logging
+from typing import TYPE_CHECKING
 
 from django.core.exceptions import ObjectDoesNotExist
 from rest_framework import serializers
 from rest_framework.exceptions import ValidationError
 from rest_framework.reverse import reverse
 
+import credentials.apps.catalog.api
 from credentials.apps.api.accreditors import Accreditor
 from credentials.apps.catalog.models import CourseRun
 from credentials.apps.credentials.constants import UserCredentialStatus
@@ -21,13 +23,17 @@ from credentials.apps.credentials.models import (
 from credentials.apps.records.models import UserGrade
 
 
+if TYPE_CHECKING:
+    from credentials.apps.credentials.models import AbstractCertificate
+
+
 logger = logging.getLogger(__name__)
 
 
 class CredentialField(serializers.Field):
     """Field identifying the credential type and identifiers."""
 
-    def to_internal_value(self, data):
+    def to_internal_value(self, data) -> "AbstractCertificate":
         site = self.context["request"].site
 
         program_uuid = data.get("program_uuid")
@@ -36,27 +42,23 @@ class CredentialField(serializers.Field):
                 return ProgramCertificate.objects.get(program_uuid=program_uuid, is_active=True)
             except ObjectDoesNotExist:
                 msg = f"No active ProgramCertificate exists for program [{program_uuid}]"
-                logger.exception(msg)
+                logger.error(msg)
                 raise ValidationError({"program_uuid": msg})
 
         course_run_key = data.get("course_run_key")
-        if course_run_key:
+        if not course_run_key:
+            raise ValidationError("Credential identifier is missing.")
+
+        # If we get this far, we necessarily have a course_run_key
+        course_runs = credentials.apps.catalog.api.get_course_runs_by_course_run_keys([course_run_key])
+        if course_runs:
+            course_run = course_runs[0]
             if self.read_only:
                 try:
-                    cert = CourseCertificate.objects.get(course_id=course_run_key, site=site)
+                    cert = CourseCertificate.objects.get(course_run=course_run, site=site)
                 except ObjectDoesNotExist:
                     cert = None
             else:
-                try:
-                    course_run = CourseRun.objects.get(key=course_run_key)
-                except ObjectDoesNotExist:
-                    logger.exception(
-                        "Course run certificate failed to create "
-                        f"because the course run [{course_run_key}] doesn't exist in the catalog"
-                    )
-                    # TODO: course_run is still nullable while we validate the data. Re raise the exception once the
-                    # field is no longer nullable
-                    course_run = None
                 # Create course cert on the fly, but don't upgrade it to active if it's manually been turned off
                 cert, _ = CourseCertificate.objects.get_or_create(
                     course_id=course_run_key,
@@ -67,15 +69,19 @@ class CredentialField(serializers.Field):
                         "certificate_type": data.get("mode"),
                     },
                 )
-
             if cert is None or not cert.is_active:
-                msg = f"No active CourseCertificate exists for course run [{course_run_key}]"
-                logger.exception(msg)
+                msg = "No active CourseCertificate exists for course run [%s]" % (course_run_key)
                 raise ValidationError({"course_run_key": msg})
-
             return cert
-
-        raise ValidationError("Credential identifier is missing.")
+        elif self.read_only:
+            msg = "No active CourseCertificate exists for course run [%s]" % (course_run_key)
+            raise ValidationError({"course_run_key": msg})
+        else:
+            msg = (
+                "CourseCertificate failed to create because the CourseRun %s doesn't exist in the catalog",
+                course_run_key,
+            )
+            raise ValidationError({"course_run_key": msg})
 
     def to_representation(self, value):
         """Serialize objects to a according to model content-type."""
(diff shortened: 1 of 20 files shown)


credentials/apps/catalog/tests/test_api.py
```diff
@@ -19,6 +20,10 @@
 from credentials.apps.credentials.tests.factories import CourseCertificateFactory
 
 
+if TYPE_CHECKING:
+    from credentials.apps.credentials.models import CourseCertificate
```
deborahgu, line 24: I don't really see how you can add code coverage for `if TYPE_CHECKING`

credentials/apps/api/v2/serializers.py
```diff
@@ -2,12 +2,14 @@
...
 """
 import logging
+from typing import TYPE_CHECKING
 
 from django.core.exceptions import ObjectDoesNotExist
 from rest_framework import serializers
 from rest_framework.exceptions import ValidationError
 from rest_framework.reverse import reverse
 
+import credentials.apps.catalog.api
```
justinhynes, line 12: Can we update this import to just import the function we need from the Python API (`get_course_runs_by_course_run_keys`)

credentials/apps/api/v2/serializers.py
```diff
@@ -67,15 +69,19 @@ def to_internal_value(self, data):
                         "certificate_type": data.get("mode"),
                     },
                 )
-
             if cert is None or not cert.is_active:
-                msg = f"No active CourseCertificate exists for course run [{course_run_key}]"
-                logger.exception(msg)
+                msg = "No active CourseCertificate exists for course run [%s]" % (course_run_key)
```
justinhynes, line 73: Can we update `msg` to use f-strings for formatting? We've been moving in the direction of f-strings as they are slightly more performant than other logging options.

credentials/apps/api/v2/serializers.py
```diff
@@ -67,15 +69,19 @@ def to_internal_value(self, data):
...
-
-        raise ValidationError("Credential identifier is missing.")
+        elif self.read_only:
+            msg = "No active CourseCertificate exists for course run [%s]" % (course_run_key)
+            raise ValidationError({"course_run_key": msg})
+        else:
+            msg = (
+                "CourseCertificate failed to create because the CourseRun %s doesn't exist in the catalog"
+                % course_run_key
+            )
```
justinhynes, line 83: See f-string comment above :)

credentials/apps/api/v2/serializers.py
```diff
@@ -315,6 +322,7 @@ def create(self, validated_data):
         course_id = validated_data["course_id"]
         course_run = None
         try:
+            # TODO use the catalog API, not direct reference
```
justinhynes, line 325: Thanks for adding a TODO for this :D

credentials/apps/api/v2/serializers.py
```diff
@@ -21,13 +23,17 @@
 from credentials.apps.records.models import UserGrade
 
 
+if TYPE_CHECKING:
+    from credentials.apps.credentials.models import AbstractCertificate
```
justinhynes, line 27: Why would we need this import behind a conditional?

credentials/apps/records/utils.py
```diff
@@ -21,6 +22,11 @@
 from credentials.apps.records.models import ProgramCertRecord, UserCreditPathway
 
 
+if TYPE_CHECKING:
+    from django.contrib.sites.models import Site
+
+    from credentials.apps.credentials.models import UserCredential
```
justinhynes, line 28: Hm. I think importing this model primarily for type checking, right? However, it's going against our best practices of not importing models directly across Django apps.

It makes me think... is the utility function `get_credentials` not located in the right place?

justinhynes: (I know that there are a lot of previous violations here that weren't made by you but I was wondering if there is a path forward without continuing to violate our best practices)

credentials/apps/records/management/commands/seed-records.py
```diff
@@ -99,7 +99,11 @@ def seed_organizations(site, faker):
 
     @staticmethod
     def seed_courses(site, organizations, faker):
-        """Seed two courses per organization"""
+        """Seed two courses per organization
+
+        Unlike in the context of a UserCredential.course_id, course_id here does literally mean
+        the course.uuid, not course_run.key.
```
justinhynes, line 105: Woof. Thanks for the comment. I'm sure it will save someone time in the future from confusion.

credentials/apps/records/api.py
```diff
@@ -16,6 +17,9 @@
 from credentials.apps.records.utils import get_credentials
 
 
+if TYPE_CHECKING:
+    from credentials.apps.credentials.models import CourseRun
```
justinhynes, line 21: (See comments below about adding new cross-app model imports)

credentials/apps/credentials/management/commands/tests/test_populate_missing_courserun_info.py
```diff
@@ -26,6 +27,9 @@ def setUp(self):
         self.course_run = CourseRunFactory()
         self.course_certificate = CourseCertificateFactory(course_id=self.course_run.key)
 
+    @pytest.mark.skip(
+        "It's now impossible to create a test object in this broken state, so no longer usefully testable"
+    )
```
justinhynes, line 32: Will the skips be temporary? Will we come back to remove the tests at a later date?

credentials/apps/credentials/models.py
```diff
@@ -198,8 +198,8 @@ class CourseCertificate(AbstractCertificate):
     .. no_pii: This model has no PII.
     """
 
-    # course_id, despite the name, is a course run key. It is a deprecated
-    # property and will be removed. If you need the value stored in course_id,
+    # course_id is identical to course_run.key. It is a deprecated legacy property.  # For now it is still used as a
```
justinhynes, line 201: nit: did your comment get auto-formatted here?

credentials/apps/credentials/models.py
```diff
@@ -234,7 +234,23 @@ class Meta:
 
     @cached_property
     def course_key(self):
-        return CourseKey.from_string(self.course_id)
+        return CourseKey.from_string(self.course_run.key)
+
+    def save(self, *args, **kwargs):
+        """Force-create the course_run or sync course_id if missing.
```
justinhynes, line 240: Is this comment correct? How would we be forcing the creation of a course run?

credentials/apps/api/v2/serializers.py
deborahgu: I could *not* get monkeypatch to work any other way, I banged my head against this one for ages.

credentials/apps/api/v2/serializers.py
deborahgu: Ah, see, I explicitly recommend against f-strings for logging messages, because they're _less_ performant. (They automatically resolve interpolation, even it the log level isn't called.) [This was my OEP draft](https://gist.github.com/deborahgu/34633ce9978a869bd3982438a5f2515e)

credentials/apps/api/v2/serializers.py
deborahgu: Because there's no reason to import at runtime; it's just [a helpful conditional for static type checkers](https://docs.python.org/3/library/typing.html#constant). It helps the IDE help us, but shouldn't be consumed at runtime.

credentials/apps/records/utils.py
deborahgu: `if TYPE_CHECKING` isn't a runtime import, and it's specifically considered safe to avoid import cycles. It's really only there for type checkers, and is by design safe for cross-module imports.

credentials/apps/credentials/management/commands/tests/test_populate_missing_courserun_info.py
deborahgu: yeah honestly I waffled about straight up removing the tests but I felt nervous about it.

credentials/apps/credentials/models.py
deborahgu: why yes it did :D

credentials/apps/api/v2/serializers.py
justinhynes: OK! Thanks for the heads up and I don't think it's a big enough issue to block the PR over it. 👍

credentials/apps/api/v2/serializers.py
justinhynes: Thank you for this explanation. Super helpful!

justinhynes (approved):
LGTM! Thanks for answering the questions and offline conversation.

PR #2221, opened 2023-10-05
chore: using new version of i18.

awais786:
Upgrading to edx-i18n-tools==1.3.0

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,7 +96,7 @@ format: ## format code
 	black .
 
 format_check: ## check that code is formatted correctly
-	black --check .
+	black --check credentials/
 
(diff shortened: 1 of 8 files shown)


awais786:
I have removed black check temporarily.

Makefile
```diff
@@ -96,7 +96,7 @@ format: ## format code
 	black .
 
 format_check: ## check that code is formatted correctly
-	black --check .
+	black --check credentials/
```
awais786, line 99: Not sure its ideal way

zubairshakoorarbisoft (approved):
LGTM

PR #2244, opened 2023-10-16
feat: add indexes to speed up django admin

deborahgu:
This adds indexes for the Boolean fields inherited from the abstract models, and also for full name. This is about trying to improve the speed of reads, but it might give us a penalty on writes, so we will need to keep an eye on alerts from new relic.

Note that application of the migration might lock the table for a bit. https://realpython.com/create-django-index-without-downtime/ has ways to work around that but I don't think they are necessary. Brief locks to credentials shouldn't be catastrophic. (When I tested it in a low-powered development environment with about 5 million rows it took approximately half a minute.)

FIXES: APER-2970

- [ ] Run `make test-karma`
- [ ] All tests pass


credentials/apps/core/migrations/0023_add_core_user_indices.py
```diff
@@ -0,0 +1,22 @@
+# Generated by Django 3.2.21 on 2023-10-16 15:44
```
justinhynes, line 1: Hm. I'm not sure if this is a real concern but figured I would bring it up. 

The Credentials IDA has already been upgraded to Django 4.2 in Stage and Prod (and devstack). 

Since this migration was generated in 3.2.x, can I request that we regenerate this migration with a system running Django 4.2.x? 

I feel that it is _unlikely_ that the migration is different between the two versions, but I think I would feel better knowing it was generated using a version of Django that we're currently using.

justinhynes (approved):
LGTM!

PR #2250, opened 2023-10-18
feat: publish program cert events via config

justinhynes:
[APER-2873]

This PR converts Credentials to use a publish-by-config model when sendind events to the Event Bus. The original code added to publish events by manual push will be removed at a later date.

These changes include a slight refactor of the event publishing code to ensure that we can move to a publish-by-config model without duplicating events.

A new setting `SEND_LEARNING_PROGRAM_CERTIFICATE_LIFECYCLE_EVENTS_TO_BUS` is introduced, which will control whether _all_ program certificate events are sent to the event bus. Previously, there were two separate settings controlling this behavior.

- [ ] Run `make test-karma`
- [ ] All tests pass


credentials/apps/core/api.py
```diff
@@ -11,6 +11,19 @@
 logger = logging.getLogger(__name__)
 
 
+def get_user_by_username(username):
```
justinhynes, line 14: Adding a simple public Python utility function to the `core` app's `api.py` for retrieving a user via username. I wanted to try and provide a way to eliminate some boiler plate User lookup code that appears all throughout the codebase.

credentials/apps/credentials/events.py
```diff
@@ -0,0 +1,25 @@
+"""
```
justinhynes, line 1: This is lifted directly from some code Arch-BOM added for this same purpose. It is temporary code and will be removed in another commit after we officially move to a publish-by-config configuration.

credentials/apps/credentials/issuers.py
```diff
@@ -231,6 +228,32 @@ def _send_program_completion_email(self, username, credential, created, lms_user
         if created and getattr(settings, "SEND_EMAIL_ON_PROGRAM_COMPLETION", False):
             send_program_certificate_created_message(username, credential, lms_user_id)
 
+    def _generate_program_certificate_event_data(self, username, user_credential, credential):
```
justinhynes, line 231: I split this code out into a separate function to try and reduce cognitive load in the `_emit_program_certificate_signal` function.

credentials/apps/credentials/issuers.py
```diff
@@ -247,29 +244,18 @@ def _emit_program_certificate_signal(self, username, user_credential, status, cr
              credential
             date_override (DateTime): A DateTime describing when the learner's credential is available to view
         """
-        program_cert_awarded_signal_enabled = SEND_PROGRAM_CERTIFICATE_AWARDED_SIGNAL.is_enabled()
-        program_cert_revoked_signal_enabled = SEND_PROGRAM_CERTIFICATE_REVOKED_SIGNAL.is_enabled()
-        if not program_cert_awarded_signal_enabled and not program_cert_revoked_signal_enabled:
-            # if neither of the signals are enabled we have nothing to do here
-            return
```
justinhynes, line 254: Originally, I thought that we would check to see if event bus functionality was enabled. If not, I wouldn't even bother calling emitting the signal. 

Having the code written this way was a pre-optimization that wasn't required (and I can hear a previous mentor saying "pre-optimization is the root of all evil" in my head). 

It made it very hard to ensure we weren't emitting duplicate events when adding the publish-by-config functionality. 

Checking if the currently existing feature flags are enabled will happen in `signals.py`.

credentials/apps/credentials/signals.py
```diff
@@ -66,12 +69,28 @@ def listen_for_program_certificate_events(sender, signal, **kwargs):  # pylint:
...
+        # temporary: defer to EVENT_BUS_PRODUCER_CONFIG if present
+        producer_config_setting = determine_producer_config_for_signal_and_topic(
+            program_credential_event_type, getattr(settings, "PROGRAM_CERTIFICATE_EVENTS_KAFKA_TOPIC_NAME", "")
+        )
+
+        if producer_config_setting is True:
+            logger.info(f"Producing [{program_credential_event_type.event_type}] event via config")
+            return
+
+        logger.info(f"Producing [{program_credential_event_type.event_type}] event via manual send")
```
justinhynes, line 93: If the event bus functionality is enabled, we then determine if we should publish the event by configuration (if configured and enabled, and it will be disabled by default).

This function goes away later once we enable publish-by-config in our settings.

credentials/apps/credentials/tests/test_issuer.py
```diff
@@ -281,23 +281,6 @@ def test_emit_program_certificate_signal_certificate_awarded(self, mock_send):
...
-        """
-        self.site_config.records_enabled = False
-        self.site_config.save()
-
-        user = UserFactory()
-        self.issuer.issue_credential(self.certificate, user.username)
-        user_credential = UserCredential.objects.get(username=user.username, credential_id=self.certificate.id)
-
-        assert user_credential
-        assert mock_send.call_count == 0
```
justinhynes, line 299: This test is no longer needed/valid after the refactor.

credentials/apps/core/api.py
```diff
@@ -11,6 +11,19 @@
...
+    """
+    Utility function that retrieves a returns a User instance based on a given username. Returns None if a user cannot
+    be found with given username.
+    """
+    logger.info(f"Retrieving details for user with username [{username}]")
+    try:
+        return User.objects.get(username=username)
+    except User.DoesNotExist:
+        logger.warning(f"A user with username [{username}] could not be found.")
```
deborahgu, line 23: [nit(? what's a good tag for "this makes me think of a thing")] 

One thing I've been thinking about is what logging messages should be `WARN` versus `DEBUG`, thinking about a noise/cost issue with messages that are emitted for frequent issues that are non-problems. In this case, looking at the calling code, It probably makes sense to admit the log message under normal use. It's just a thing I've been mulling over in my own head at the moment.

that being said, I'm not sure I agree that the initial `INFO` message is the kind I see that much utility in; that's the kind of thing I might flag for `DEBUG`. But I don't know, how much is that the kind of thing we would want to see in splunk? 

(There's a reason I'm not calling this a nit; it's because I'm so wishy-washy about it. There's a cost to the logging noise and a cost to not having the logging information, you know?)

credentials/apps/core/api.py
deborahgu: would your plan be to replace the code in other places? And not just in the one place where you do here?

credentials/apps/credentials/issuers.py
```diff
@@ -247,29 +244,18 @@ def _emit_program_certificate_signal(self, username, user_credential, status, cr
...
-
-        try:
-            # need to retrieve the learner's User instance in order to retrieve PII data for the event
-            user = User.objects.get(username=username)
```
deborahgu, line 258: does this mean you can remove the import of / call to `get_user_model`?

credentials/apps/credentials/issuers.py
```diff
@@ -247,29 +244,18 @@ def _emit_program_certificate_signal(self, username, user_credential, status, cr
...
-        except User.DoesNotExist:
-            logger.exception(
+        user = get_user_by_username(username)
+        if not user:
+            logger.warning(
```
deborahgu, line 249: [NIT] This is a place where I start wondering about warnings. The way this code is written you are going to send two `WARN`s in this circumstance, and I don't really see an advantage of that. 

I understand that the goal of the utility method is that it might be used more than once, but right now it is used only once. So what you've got is a utility method that's only being used in one place (which seems like pre-optimizing), and is generalized enough that you are sending two logging messages for every time this one error comes up (which is noise).

(If it were me I might try to make the point of the code more obvious by finding one or two other places where you could call `get_user_by_username` and pop it in there. YMMV.)

credentials/apps/core/api.py
justinhynes: Yes, I plan on doing this as a separate PR to try and keep this one as focused on the current task.

credentials/apps/credentials/issuers.py
justinhynes: Yes! Good eye, thanks. I'll remove this now.

credentials/apps/credentials/issuers.py
justinhynes: This totally makes sense to me. This is helping me realize I am overly verbose with logging messages and I'll try to take more care/thought going forward. I will remove the info/warn from the utility function as I believe that those logs are 99.99% noise when the system is working as expected.

This warning is the more important/critical piece to know (and have part of the historical logs) if we have to investigate an issue.

PR #2276, opened 2023-11-02
feat: emit Segment event when creating or updating a program credential

justinhynes:
[APER-3017]

Adds a feature to Credentials where the system will emit a Segment event (if configured) when a program credential is awarded or updated.

New events added with this PR:
* `edx.bi.credentials.credential_issuers.program_certificate_updated`
* `edx.bi.credentials.credential_issuers.program_certificate_created`

Additionally, an example of the properties sent with the event:
```
{
    "category": "credentials",
    "user":
    {
        "username": "edx",
        "lms_user_id": 3
    },
    "program":
    {
        "uuid": "83c7ae34-5e74-44f1-a837-7eb3426b3478",
        "title": "Best Program Ever",
        "program_type": "microbachelors"
    },
    "uuid": "bd4ffd35-8c68-45c0-9eb8-9b8652411ba4",
    "status": "awarded",
    "url": "{blah}/credentials/bd4ffd358c6845c09eb89b8652411ba4/",
    "timestamp": "2023-11-01T19:21:08.300353+00:00"
}
```

- [ ] Run `make test-karma`
- [ ] All tests pass

### DIFF
--- a/credentials/apps/api/v2/views.py
+++ b/credentials/apps/api/v2/views.py
@@ -263,7 +263,7 @@ class UsernameReplacementView(APIView):
                     num_rows_changed += model.objects.filter(**{column: current_username}).update(
                         **{column: new_username}
                     )
-        except Exception as exc:  # pylint: disable=broad-except
+        except Exception as exc:
             log.exception(
                 "Unable to change username from %s to %s. Failed on table %s because %s",
                 current_username,
--- a/credentials/apps/core/api.py
+++ b/credentials/apps/core/api.py
@@ -49,7 +49,7 @@ def get_or_create_user_from_event_data(user_data):
         # create the user if they don't exist, this follows similar behavior that our JWT authentication implements if
         # a user doesn't exist when we're making network calls across services
         user, created = User.objects.get_or_create(username=user_data.pii.username)
-    except Exception:  # pylint: disable=broad-exception-caught
+    except Exception:
         logger.exception("Error occurred retrieving or creating a user")
         return None, None
 
--- a/credentials/apps/credentials/api.py
+++ b/credentials/apps/credentials/api.py
@@ -73,7 +73,7 @@ def _update_or_create_credential(username, credential_type, credential_id, statu
             credential_id=credential_id,
             defaults={"status": status},
         )
-    except Exception:  # pylint: disable=broad-exception-caught
+    except Exception:
         logger.exception(
             f"Error occurred processing a credential with credential_id [{credential_id}] for user [{username}]"
         )
@@ -131,7 +131,7 @@ def create_course_cert_config(course_run: "CourseRun", site: "Site", mode: str):
         course_cert_config = _CourseCertificate.objects.create(
             course_id=course_run.key, course_run=course_run, site=site, is_active=True, certificate_type=mode
         )
-    except Exception:  # pylint: disable=broad-exception-caught
+    except Exception:
         logger.exception(f"Error occurred creating a CourseCertificate configuration for course run [{course_run.key}]")
 
     return course_cert_config
--- a/credentials/apps/credentials/issuers.py
+++ b/credentials/apps/credentials/issuers.py
@@ -286,8 +286,8 @@ class ProgramCertificateIssuer(AbstractCredentialIssuer):
 
     def _emit_program_certificate_segment_event(self, request, site_config, user, user_credential, credential, created):
         """
-        A utility function used to dispatch a Segment event when a program certificate has been updated. This includes
-        upon initial creation or an update later (revocation, re-awarding).
+        A utility function used to dispatch a Segment event when a program certificate record has been created or
+        updated.
 
         Args:
             request (HttpRequest): The original request object from the credential update request
@@ -310,18 +310,11 @@ class ProgramCertificateIssuer(AbstractCredentialIssuer):
             segment_client = SegmentClient(write_key=site_config.segment_key)
 
         if segment_client:
-            anonymous_id = None
-            if request and request.COOKIES:
-                anonymous_id = request.COOKIES.get("ajs_anonymous_id", str(uuid4()))
-            else:
-                anonymous_id = str(uuid4())
-
             event_name = "edx.bi.credentials.credential_issuers.program_certificate_updated"
             if created:
                 event_name = "edx.bi.credentials.credential_issuers.program_certificate_created"
 
             program = credential.program
-
             event_properties = {
                 "category": "credentials",
                 "user": {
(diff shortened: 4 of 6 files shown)


credentials/apps/api/v2/views.py
```diff
@@ -263,7 +263,7 @@ def _replace_username_for_all_models(self, current_username, new_username, repla
                     num_rows_changed += model.objects.filter(**{column: current_username}).update(
                         **{column: new_username}
                     )
-        except Exception as exc:  # pylint: disable=broad-except
```
justinhynes, line 266: Not related to my changes, but was flagged by CI.

credentials/apps/core/api.py
```diff
@@ -49,7 +49,7 @@ def get_or_create_user_from_event_data(user_data):
         # create the user if they don't exist, this follows similar behavior that our JWT authentication implements if
         # a user doesn't exist when we're making network calls across services
         user, created = User.objects.get_or_create(username=user_data.pii.username)
-    except Exception:  # pylint: disable=broad-exception-caught
+    except Exception:
```
justinhynes, line 52: Not related to my changes, but was flagged by CI.

credentials/apps/credentials/api.py
```diff
@@ -73,7 +73,7 @@ def _update_or_create_credential(username, credential_type, credential_id, statu
             credential_id=credential_id,
             defaults={"status": status},
         )
-    except Exception:  # pylint: disable=broad-exception-caught
```
justinhynes, line 76: Not related to my changes, but was flagged by CI.

credentials/apps/credentials/api.py
```diff
@@ -131,7 +131,7 @@ def create_course_cert_config(course_run: "CourseRun", site: "Site", mode: str):
         course_cert_config = _CourseCertificate.objects.create(
             course_id=course_run.key, course_run=course_run, site=site, is_active=True, certificate_type=mode
         )
-    except Exception:  # pylint: disable=broad-exception-caught
+    except Exception:
```
justinhynes, line 134: Not related to my changes, but was flagged by CI.

credentials/apps/credentials/issuers.py
```diff
@@ -278,6 +284,67 @@ def _emit_program_certificate_signal(self, username, user_credential, status, cr
...
+                "status": user_credential.status,
+                "url": f"https://{credential.site.domain}/credentials/{str(user_credential.uuid).replace('-', '')}/",
+                "timestamp": user_credential.modified,
+            }
+
+            segment_client.track(
+                anonymous_id=anonymous_id,
+                event=event_name,
+                properties=event_properties,
+            )
```
justinhynes, line 346: "Thinking out loud": Maybe I should wrap this in a try/except? This is part of an atomic transaction, seems like this would be bad if it throws an error and causes us to roll back awarding or update a credential?

credentials/apps/credentials/tests/test_issuer.py
```diff
@@ -20,7 +20,11 @@
     UserCredential,
     UserCredentialAttribute,
 )
-from credentials.apps.credentials.tests.factories import CourseCertificateFactory, ProgramCertificateFactory
+from credentials.apps.credentials.tests.factories import (
```
justinhynes, line 23: Sorry ahead of time. I refactored a number of tests because the tests were hard to debug. 

We were using code under test to create test data, which caused a ton of confusion when I was trying to write/debug tests that were making changes to this functionality.

There may be more code duplication now, but we shouldn't be calling `issue_credential` from any `setUp` functions any longer.

credentials/apps/credentials/issuers.py
```diff
@@ -278,6 +284,67 @@ def _emit_program_certificate_signal(self, username, user_credential, status, cr
...
+                    "program_type": program.type_slug,
+                },
+                "uuid": str(user_credential.uuid),
+                "status": user_credential.status,
+                "url": f"https://{credential.site.domain}/credentials/{str(user_credential.uuid).replace('-', '')}/",
+                "timestamp": user_credential.modified,
+            }
+
+            segment_client.track(
+                anonymous_id=anonymous_id,
```
muhammad-ammar, line 343: Shouldn't this be lms user id?

credentials/apps/credentials/issuers.py
```diff
@@ -278,6 +284,67 @@ def _emit_program_certificate_signal(self, username, user_credential, status, cr
             # .. event_implemented_name: PROGRAM_CERTIFICATE_REVOKED
             PROGRAM_CERTIFICATE_REVOKED.send_event(time=time, program_certificate=program_certificate_data)
 
+    def _emit_program_certificate_segment_event(self, request, site_config, user, user_credential, credential, created):
```
muhammad-ammar, line 287: Are there any plans to extract the common code from this method into a utility function so that others can use it?

credentials/apps/credentials/issuers.py
justinhynes: AFAIK, this is a separate ID required for tracking on Segment's side. https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/identity/#anonymous-ids. 

We check the request cookies and we'll pull out and use the `ajs_anonmyous_id` if found, otherwise we just generate a new one. The POST will fail if I don't include it. 

I _am_ including `lms_user_id` but it will only be accessible in the event properties.

credentials/apps/credentials/issuers.py
justinhynes: A good question! This was something we were doing on a short timeframe because of the Nov 6th deadline. There were other options we wanted to explore but the timeline wouldn't work.

I'd expect us to circle back and make this a bit more general/reusable in the future.

credentials/apps/credentials/issuers.py
justinhynes: After doing a bit more reading and poking about their source code (the client is open source) I think we're okay as it is.

credentials/apps/credentials/issuers.py
muhammad-ammar: If we see the segment events triggered in edx-platform and other places like license-manager, we send the lms user id, for example, 
* https://github.com/openedx/edx-platform/blob/9ba9935fb613690f18126874fc13ba2e1b3420be/common/djangoapps/track/segment.py#L65
* https://github.com/openedx/license-manager/blob/fccf0c3ac4ec2c85315067d7b350840a7b959c46/license_manager/apps/subscriptions/event_utils.py#L170

justinhynes: Oh! Interesting. I had been using different examples but I can see that this is clearly a pattern. I'l update it to be the LMS User Id.

jsnwesson (approved):

credentials/apps/credentials/issuers.py
muhammad-ammar: @justinhynes Any idea how can we test that the new event is triggering and see it in segment or Braze?

PR #2280, opened 2023-11-06
fix: show earned credentials even if no grade is available in Credentials

justinhynes:
There are well documented communication issues between the LMS and Credentials systems. There are times that Credentials will just miss a grade update from the LMS. The grades stored in Credentials are used to render a learner's progress in their courses through the Learner Record UI.

Now, if the system is aware of an earned Credential, we will populate the `issue_date` field (which the frontend uses to determine if a certificate is earned or not). This way we can reflect the learner record as accurately as possible for a learner.

- [ ] Run `make test-karma`
- [ ] All tests pass


justinhynes:
Closing, found a bug in this change and made more changes. A new PR will be opened shortly.

PR #2287, opened 2023-11-13
fix: display earned course credentials even if the grade is missing

justinhynes:
There are well documented communication issues between the LMS and Credentials systems. There are times that Credentials will just miss a grade update from the LMS. The grades stored in Credentials are used to render a learner's progress in their courses through the Learner Record UI.

Now, if the system is aware of an earned Credential, we will populate the issue_date field (which the frontend uses to determine if a certificate is earned or not). This way we can reflect the learner record as accurately as possible for a learner.

**Screenshots**
_Current, a learner who has earned a credential in a course, but the grade is missing, will see the credential as "Not Earned"_
image

_After change, a learner who has earned a credential in a course, but the grade is missing, will see the credential as earned (albeit missing grade information)_
image

- [x] Run `make test-karma`
- [x] All tests pass

### DIFF
--- a/credentials/apps/records/api.py
+++ b/credentials/apps/records/api.py
@@ -117,7 +117,7 @@ def _get_transformed_pathway_data(program, user):
     return pathway_data
 
 
-def _get_transformed_grade_data(program, user):
+def _get_transformed_grade_data(program, user):  # pylint: disable=too-many-statements
     """
     A utility function that gathers and transforms a learner's grade data for course-runs that are part of a Program.
     This data is used to render a learner's Program Record page.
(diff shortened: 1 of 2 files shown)


credentials/apps/records/tests/test_api.py
```diff
@@ -184,26 +207,236 @@ def test_get_transformed_pathway_data(self):
...
+        self._assert_results(expected_highest_attempt_dict, highest_attempt_dict)
+        assert (
+            float(highest_attempt_dict.get(self.course1).percent_grade) == self.course1_courserun2_grade.percent_grade
+        )
+        assert (
+            float(highest_attempt_dict.get(self.course2).percent_grade) == self.course2_courserun1_grade.percent_grade
+        )
+        assert expected_issue_date_course2 == last_updated
+
+    def test_get_transformed_grade_data_has_credential_no_grade(self):
```
justinhynes, line 252: This test (and the test following this one) are not directly related to the changes in this PR. 

They _are_ important to the functionality of the feature/component as a whole. Since I was already adding/changing existing tests, I spent some time to add additional tests for edge case functionality that want to keep tabs on.

credentials/apps/records/tests/test_api.py
```diff
@@ -184,26 +207,236 @@ def test_get_transformed_pathway_data(self):
...
+        self._assert_results(expected_result_course2, result[1])
         self._assert_results(expected_highest_attempt_dict, highest_attempt_dict)
-        assert float(highest_attempt_dict.get(self.course).percent_grade) == self.grade_high.percent_grade
-        assert expected_issue_date == last_updated
+        assert (
+            float(highest_attempt_dict.get(self.course1).percent_grade) == self.course1_courserun2_grade.percent_grade
+        )
+        assert expected_issue_date_course1 == last_updated
+
+    def test_get_transformed_grade_data_earned_credential_with_visible_date(self):
```
justinhynes, line 341: This test (and the test following this one) are not directly related to the changes in this PR. 

They _are_ important to the functionality of the feature/component as a whole. Since I was already adding/changing existing tests, I spent some time to add additional tests for edge case functionality that we want to keep tabs on.

credentials/apps/records/api.py
```diff
@@ -185,37 +185,73 @@ def _get_transformed_grade_data(program, user):
...
+                    "issue_date": issue_date.isoformat(),
+                    "percent_grade": float(grade.percent_grade),
+                    "letter_grade": grade.letter_grade or _("N/A"),
                 }
             )
             added_courses.add(course)
-        elif grade is not None and grade.course_run == course_run:
-            user_credential = user_credential_dict.get(course_run.key)
-            issue_date = get_credential_dates(user_credential, False)
+        # Case 2: Learner has earned a credential in, but we have no record of a grade for, a course-run of a course
```
deborahgu, line 221: does it actually make sense to build these three dicts separately? They seem similar enough that it seems like it would make more sense to create a variable for `attempts`, `letter_grade` etc. for the cases and just build the  dict once.

justinhynes: OK! I can try and make it a bit more DRY :)

credentials/apps/records/api.py
```diff
@@ -117,7 +117,7 @@ def _get_transformed_pathway_data(program, user):
     return pathway_data
 
 
-def _get_transformed_grade_data(program, user):
+def _get_transformed_grade_data(program, user):  # pylint: disable=too-many-statements
```
deborahgu, line 120: oof. I'm reading this and it looks clearer to me than the other way, but arguably anything that triggers movement into `too-many-statements` is a red flag, you know? So the DRY improvements are balanced out by the `too-many-statements` trigger. I'm going to go ahead and approve this, but it sounded like you thought the other one was more clear to read, so given the above if you want to roll back to the old way I will go ahead and approve that instead.

deborahgu (approved):

credentials/apps/records/api.py
justinhynes: The linting warning came after the most recent changes. 

I think this code is terribly overcomplicated either way, and it really requires a more extensive re-write, so I'm trying to pick the least horrible improvement (for now)

PR #2315, opened 2023-12-04
fix: display earned course credentials even if the grade is missing

justinhynes:
[APER-2994]

There are well documented communication issues between the LMS and Credentials systems. There are times that Credentials will just miss a grade update from the LMS. The grades stored in Credentials are used to render a learner's progress in their courses through the Learner Record UI.

Now, if the system is aware of an earned Credential, we will populate the issue_date field (which the frontend uses to determine if a certificate is earned or not). This way we can reflect the learner record as accurately as possible for a learner.

- [ ] Run `make test-karma`
- [ ] All tests pass

### DIFF
--- a/credentials/apps/records/api.py
+++ b/credentials/apps/records/api.py
@@ -117,7 +117,7 @@ def _get_transformed_pathway_data(program, user):
     return pathway_data
 
 
-def _get_transformed_grade_data(program, user):
+def _get_transformed_grade_data(program, user):  # pylint: disable=too-many-statements
     """
     A utility function that gathers and transforms a learner's grade data for course-runs that are part of a Program.
     This data is used to render a learner's Program Record page.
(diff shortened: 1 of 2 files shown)


credentials/apps/records/api.py
```diff
@@ -185,37 +185,78 @@ def _get_transformed_grade_data(program, user):
...
 
-    # add the course-run grade data to the response in the order that is maintained by the Program's sorted field
+    # create a collection of awarded credentials mapped to a *course*, this will help us build responses below and we
+    # need an easy way to know if a learner has earned a course credential in any eligible course runs of a specific
+    # course
+    awarded_course_credential_dict = {
+        user_credential.credential.course_run.course: user_credential
+        for user_credential in course_user_credentials
+        if user_credential.status == UserCredentialStatus.AWARDED.value
+        and user_credential.credential.course_run in program_course_runs
```
justinhynes, line 195: Change from original: Adding a second conditional here to filter for awarded credentials in course-runs _that are part of the program_.

deborahgu, line 195: non-blocking: it is about three times as fast to search in the frozenset as in the list, as long as it exists (as far as I can tell it has identical contents at this point). They are both incredibly fast operations, so it's not a necessary optimization in this case, but as long as you have taken the effort to build the frozenset (which does take time), there's no reason not to optimize. Set operations are usually really fast, and I like to be in the habit of remembering to use them if I have a set available.

credentials/apps/records/tests/test_api.py
```diff
@@ -184,26 +207,320 @@ def test_get_transformed_pathway_data(self):
...
+        self._assert_results(expected_highest_attempt_dict, highest_attempt_dict)
+        assert (
+            float(highest_attempt_dict.get(self.course1).percent_grade) == self.course1_courserun1_grade.percent_grade
+        )
+        assert (
+            float(highest_attempt_dict.get(self.course2).percent_grade) == self.course2_courserun1_grade.percent_grade
+        )
+        assert expected_issue_date_course2 == last_updated
+
+    def test_get_transformed_grade_data_program_with_excluded_course_run(self):
```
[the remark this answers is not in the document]
deborahgu: just out of curiosity, did you add this test without the fixed code first, so that you could make sure it triggered under the original conditions?

deborahgu (approved):
I left nonblocking changes and a question – if you want to go ahead and merge like this go for it. Although because the added test is complex enough (it's hard to keep straight with all of the `courseX_courserunY`), if you haven't run it on the buggy version to make sure it fails there, I encourage you to.

The comments are more on the order of pedantry, things I would have done differently, just because I read the PR more carefully than usual.

credentials/apps/records/api.py
justinhynes: Seems totally reasonable to make this change. Thanks!

credentials/apps/records/api.py
```diff
@@ -185,37 +185,78 @@ def _get_transformed_grade_data(program, user):
...
+    }
```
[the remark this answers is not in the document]
justinhynes: The part about debugging hits home with me. I've converted this to a for loop over a comprehension. :)

credentials/apps/records/tests/test_api.py
```diff
@@ -179,37 +228,431 @@ def test_get_transformed_pathway_data(self):
...
+        self._assert_results(expected_result[0], result[0])
+        self._assert_results(expected_result[1], result[1])
         self._assert_results(expected_highest_attempt_dict, highest_attempt_dict)
-        assert float(highest_attempt_dict.get(self.course).percent_grade) == self.grade_high.percent_grade
-        assert expected_issue_date == last_updated
+        # in this case, because of how the logic currently is implemented, we should expect the "last updated" date to
+        # be associated with available grades
+        assert expected_issue_date_course1 == last_updated
+
+    def test_get_transformed_grade_data_no_progress(self):
```
justinhynes, line 317: (apologies in advance, I attempted to make the tests easier to read, not sure if I succeeded in that, but I'll stop messing with them haha)

@deborahgu -- this is one of two new tests I added after your initial approval.

credentials/apps/records/tests/test_api.py
```diff
@@ -179,37 +228,431 @@ def test_get_transformed_pathway_data(self):
...
+            self.course2: self.course2_courserunA_grade,
+        }
+
+        result, highest_attempt_dict, last_updated = _get_transformed_grade_data(self.program1, self.user)
+        self._assert_results(expected_result[0], result[0])
+        self._assert_results(expected_result[1], result[1])
+        self._assert_results(expected_highest_attempt_dict, highest_attempt_dict)
+        assert expected_issue_date_course2 == last_updated
+
+    def test_get_transformed_grade_data_program_with_excluded_course_run(self):
```
justinhynes, line 489: This is the second of the new tests (but retains the original tests name, the former test is below). I split them into "excluded course runs" and "excluded course run with missing grade".

deborahgu (approved):
The changes do make the tests easier to read. The noisiest things for tests so often setting up the broken test cases and explaining why they are broken 😆

PR #2322, opened 2023-12-14
feat:  bulk learner cert status and point

deborahgu:
Adds a bulk version of the learner_cert_status endpoint, following the pattern we use elsewhere for bulk variants: `bulk_learner_cert_status`.

Extrapolating the logic for a single learner/course combo into the API. Whether that particular structure belonged in api.py or in the specific API view is a matter of art rather than science; it felt like a more generalized function for the Python API rather than something for the view. (Reasonable developers could disagree; the specifics of the structure here are designed around the API view so you could go either way. To me it felt cleaner to leave it in `api.py`.)

This does have some repeated code as a side effect of not wanting to change the contract for the previously-existing endpoint, but it's kept to a minimum..

FIXES: APER-3102

- [ ] Run `make test-karma`
- [ ] All tests pass

### DIFF
--- a/credentials/apps/records/api.py
+++ b/credentials/apps/records/api.py
@@ -1,12 +1,15 @@
 import datetime
 from collections import defaultdict
-from typing import TYPE_CHECKING, List
+from typing import TYPE_CHECKING, Any, Dict, List, Optional
 
+from django.contrib.auth import get_user_model
 from django.contrib.contenttypes.models import ContentType
+from django.core.exceptions import BadRequest
 from django.template.defaultfilters import slugify
 from django.utils.translation import gettext as _
 
 from credentials.apps.catalog.api import get_program_and_course_details
+from credentials.apps.core.api import get_user_by_username
 from credentials.apps.credentials.api import (
     get_credential_dates,
     get_user_credentials_by_content_type,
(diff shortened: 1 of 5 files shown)


docs/credentials_api.rst
```diff
@@ -268,3 +268,81 @@ In this example, this user has earned a certificate in only one of the courses r
...
+                "certificate_available_date": null,
+                "grade": {
+                    "letter_grade": "Pass",
+                    "percent_grade": 1,
+                    "verified": true
+                    }
+                }
+            ]
+        }
+    ]
```
justinhynes, line 348: Is this correct? I assumed that there would be a placeholder dict/JSON object for the learner that doesn't have the credential? I gave this a test drive locally, I made a request to the bulk endpoint with one "real" user and one user that I know DNE. I got two results in my response.

Request:
```
[
    {
        "username": "sample_user",
        "course_runs":
        [
            "course-v1:edX+hynes101+1T2023"
        ]
    },
    {
        "username": "edx",
        "course_runs":
        [
            "course-v1:edX+hynes101+1T2023"
        ]
    }
]
```

Response:
```
[
    {
        "lms_user_id": null,
        "username": "sample_user",
        "status": []
    },
    {
        "lms_user_id": 3,
        "username": "edx",
        "status": [
            {
                "course_uuid": "41808971-83b2-4ad5-958b-0fec6aa4f032",
                "course_run": {
                    "uuid": "2a35752d-958f-47cd-aabc-ce2b1c91ab96",
                    "key": "course-v1:edX+hynes101+1T2023"
                },
                "status": "awarded",
                "type": "verified",
                "certificate_available_date": null,
                "grade": {
                    "letter_grade": "Pass",
                    "percent_grade": 1.0,
                    "verified": true
                }
            }
        ]
    }
]
```

Response:

justinhynes: Also, should line 342 in the example show as a decimal over an integer?

credentials/apps/records/api.py
```diff
@@ -407,3 +410,48 @@ def get_learner_course_run_status(username: str, course_ids: List[str], course_r
...
+
+    * You must include _exactly one_ of `lms_user_id` or `username`.
+    * You must include at least one of `courses` and `course_runs`, and you may include a mix of both.
+        * The `courses` list should contain a list of course UUIDs.
+        * The `course_runs` list should contain a list of course run keys.
+
+    Returns:
+        A dict of an individuals user's earned certificates for a list of courses or course runs.
+
+        If the userhas not earned any certificates, the `status` object will be empty.
```
justinhynes, line 431: nit: need a space between "user" and "has"

credentials/apps/records/api.py
```diff
@@ -407,3 +410,48 @@ def get_learner_course_run_status(username: str, course_ids: List[str], course_r
...
+) -> Dict[str, Any]:
+    """Fetches details for earned certificates for a single user.
+
+    * You must include _exactly one_ of `lms_user_id` or `username`.
+    * You must include at least one of `courses` and `course_runs`, and you may include a mix of both.
+        * The `courses` list should contain a list of course UUIDs.
+        * The `course_runs` list should contain a list of course run keys.
+
+    Returns:
+        A dict of an individuals user's earned certificates for a list of courses or course runs.
```
justinhynes, line 429: Nit: does `individuals` need to be plural?

justinhynes (approved):
Couple of nits and a question, but no blocking criticism.

PR #2360, opened 2024-01-24
feat: allow download of program record even when segment is blocked

deborahgu:
we were assuming that segment was always being permitted from the browser side, but segment is blocked by every single privacy built in, including strict tracking protections in the browser. Allow download of the record in that case.

FIXES: APER-3164

- [x] Run `make test-karma`
- [x] All tests pass

### DIFF
--- a/credentials/apps/records/views.py
+++ b/credentials/apps/records/views.py
@@ -316,10 +316,10 @@ class ProgramRecordCsvView(RecordsEnabledMixin, View):
                 request.site,
                 platform_name=site_configuration.platform_name,
             )
-        except Exception as e:
+        except Exception:
             log.error(
                 "get_program_record failed for:\n"
-                "   user: {userid}\n"
+                "   user: {user_id}\n"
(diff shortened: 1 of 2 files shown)


credentials/apps/records/views.py
```diff
@@ -300,26 +304,47 @@ def close(self):
...
 
         user_metadata = [
-            ["Program Name", record["program"]["name"]],
-            ["Program Type", record["program"]["type_name"]],
-            ["Platform Provider", record["platform_name"]],
-            ["Authoring Organization(s)", record["program"]["school"]],
-            ["Learner Name", record["learner"]["full_name"]],
-            ["Username", record["learner"]["username"]],
-            ["Email", record["learner"]["email"]],
+            ["Program Name", program.get("name", None)],
```
justinhynes, line 341: Thank you for this. Definitely should make it more robust/resilient. 👍

justinhynes (approved):
Looks like a good change! Thanks for your work on this.

PR #2364, opened 2024-01-26
feat: add decorator that can be used to log incoming REST request data

justinhynes:
[APER-3122]

Adds a decorator that can be used to log incoming REST request data. This will be used to get more visibility into the grade updates flowing from the LMS to Credentials.

This functionality will be gated using a WaffleSwitch so that we don't cause a massive spike in log size. This is intended to be used for debugging when needed.

- [ ] Run `make test-karma`
- [ ] All tests pass

### DIFF
--- a/credentials/apps/api/v2/decorators.py
+++ b/credentials/apps/api/v2/decorators.py
@@ -1,0 +1,39 @@
+"""
+Decorators for the API app of Credentials.
+"""
+
+import logging
+
+from django.conf import settings
+
+
+logger = logging.getLogger(__name__)
+
+
+def log_incoming_request(func):
+    """
+    A decorator that can be used to log information from an incoming REST request. Please take care if using this
+    decorator, we don't want to accidentally log any secrets or sensitive PII.
+
+    Use of this decorator is also gated by the `LOG_INCOMING_REQUESTS` waffle switch. This is to ensure that we can
+    toggle this functionality as needed and keep log sizes reasonable.
+    """
+
+    def wrapper(*args, **kwargs):
+        if settings.LOG_INCOMING_REQUESTS.is_enabled():
+            try:
+                request = args[1]
+                data = request.body
+                logger.info(
+                    f"{request.method} request received to endpoint [{request.get_full_path()}] from user "
+                    f"[{request.user.username}] originating from [{request.META.get('HTTP_HOST', 'Unknown')}] "
+                    f"with data: [{data}]"
+                )
+            except Exception as exc:
+                # catch overly broad exception to try to ensure if logging doesn't work the request will still be
+                # processed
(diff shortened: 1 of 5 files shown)


credentials/apps/api/v2/decorators.py
```diff
@@ -0,0 +1,39 @@
+"""
+TODO
```
justinhynes, line 2: YEESH. I forgot to fill this in. I'll take care of this before a final merge.

deborahgu, line 2: probably you intend to change that

credentials/settings/base.py
```diff
@@ -584,3 +584,12 @@
         },
     },
 }
+
+# .. toggle_name: LOG_INCOMING_REQUESTS
+# .. toggle_implementation: WaffleSwitch
+# .. toggle_default: False
+# .. toggle_description: Toggle to control whether we log incoming REST requests through the use of the `log_request`
```
justinhynes, line 591: Just realized the name of the decorator here is incorrect. Will fix.

credentials/apps/api/v2/decorators.py
```diff
@@ -0,0 +1,39 @@
...
+"""
+
+import logging
+
+from django.conf import settings
+
+
+LOGGER = logging.getLogger(__name__)
```
deborahgu, line 10: I've noticed this in some of our older code, but we shouldn't be allcapsing python variables, unless there is some style guide you are following here that I haven't seen or have expunged from my memory to save sanity points.

credentials/apps/api/v2/decorators.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def log_incoming_request(func):
+    """
+    A decorator that can be used to log information from an incoming REST request. Please take care if using this
+    decorator, we don't want to accidentally log any secrets or sensitive PII.
```
deborahgu, line 16: 🎉

deborahgu (changes requested):
a couple of minor things, plus the code coverage report, but looks really nice. I love the waffle switch. I really do want to raise a greater conversation about putting debug logging  behind a waffle switch; I think that could be super valuable, at least in a smaller IDA like credentials. In edx-platform we would probably drown, lol.

credentials/apps/api/v2/decorators.py
justinhynes: Fixed!

credentials/settings/base.py
justinhynes: Fixed!

credentials/apps/api/v2/decorators.py
justinhynes: Yepppp! Thanks, fixed in an update being pushed shortly.

PR #2376, opened 2024-02-02
feat: management command to truncate social auth

deborahgu:
truncates all entries in this table that were modified outside of the last 90 days. Doesn't take any arguments because YAGNI; can easily be modified in the future if the situation changes.

FIXES: APER-3160


justinhynes (approved):
I like the approach!

PR #2446, opened 2024-04-16
fix: fix social media and sharing icons

justinhynes:
The Credentials IDA gives our users the ability to share or print their program certificates if they wish. This functionality is enabled by rendering icons (provided via a third party package called `fontawesome`).

Recently, something caused this functionality to break and the icons are rendering as squares.

This PR attempts to fix this issue. Initially, we attempted to troubleshoot the original issue with the old `fontawesome` package (which was released 7 years ago). We were unable to do that. Then, we tried to update to the latest version of fontawesome via NPM. This caused additional issues that weren't well understood. Finally, I have decided to remove the npm package in favor of a pip package that is designed to work specifically with Django.

Thus, this PR adds a new default installed app to the IDA's `installed_apps` setting -- `fontawesomefree`.

This approach worked and has fixed the display of the icons used in the sharing options for a program certificate rendered by the Credentials IDA.

Additionally, an update was made to the `base.html` template to import the fontawesome icons, as well as an update to the `_actions.html` template to update the class names appropriate to the latest version of `fontawesome`.

Before:
image

After:
image

- [x] Run `make test-karma`
- [x] All tests pass

### DIFF
--- a/credentials/settings/base.py
+++ b/credentials/settings/base.py
@@ -63,6 +63,7 @@ THIRD_PARTY_APPS = [
     "drf_yasg",
     "xss_utils",
     "openedx_events",
+    "fontawesomefree",
 ]
 
 PROJECT_APPS = [
--- a/credentials/templates/_actions.html
+++ b/credentials/templates/_actions.html
@@ -12,7 +12,7 @@
               data-track-event-property-category="certificates"
               data-track-event-property-credential-uuid="{{ user_credential.uuid }}"
               data-track-event-property-program-uuid="{{ user_credential.credential.program_uuid }}">
-        <span class="fa fa-facebook" aria-hidden="true"></span>
+        <span class="fa-brands fa-facebook" aria-hidden="true"></span>
         <span class="action-label">{% trans 'Share this certificate via Facebook' as tmsg %}{{ tmsg | force_escape }}</span>
       </button>
     {% endif %}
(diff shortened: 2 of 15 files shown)


deborahgu (approved):
this looks good to me once the conflicts resolve.

credentials/templates/_actions.html
```diff
@@ -25,7 +25,7 @@ <h3 class="sr-only">{% trans 'Print or share your certificate' as tmsg %}{{ tmsg
               data-track-event-property-program-uuid="{{ user_credential.credential.program_uuid }}">
         <a class="share-link" target="_blank"
            href="https://twitter.com/intent/tweet?text={{ tweet_text|urlencode }}&url={{ share_url|urlencode }}{% if twitter_username %}&via={{ twitter_username }}{% endif %}">
-          <span class="fa fa-twitter" aria-hidden="true"></span>
+          <span class="fa-brands fa-twitter" aria-hidden="true"></span>
```
justinhynes, line 28: classnames changed from fontawesome v4.x --> v.6x, brand icons require the `fa-brands` classname now.

credentials/settings/base.py
```diff
@@ -63,6 +63,7 @@
     "drf_yasg",
     "xss_utils",
     "openedx_events",
+    "fontawesomefree",
```
justinhynes, line 66: adding `fontawesomefree` as a new third party app for the Credentials IDA

requirements/base.txt
```diff
@@ -174,7 +176,7 @@ jinja2==3.1.3
     # via
     #   code-annotations
     #   coreschema
-lxml[html-clean,html_clean]==5.1.1
+lxml[html-clean]==5.1.1
```
justinhynes, line 179: this duplication looked fishy, so I removed one of them

deborahgu: as I recall, when I added the first one to `base.in`, the make command that builds requirements added the 2nd.  Should we be manually editing `base.txt`?  I thought it was always autogenerated.

justinhynes: I may have made a bad assumption. I was fixing up some conflicts and thought this was part of my issues. Generally I'm against modifying the generated files, but maybe I'll just run through another `make upgrade` and use what it gives me.

justinhynes: running `make upgrade` reverted my change back to the original version, so I'll roll with that.

PR #2449, opened 2024-04-16
feat: django admin action to mass toggle User attributes

deborahgu:
As part of a cleanup, we need to deactivate and de-privilege a very large number of long-gone privileged django admin users. Luckily, Django admin  has functionality for this, via the `actions` mechanism.

This PR adds the ability to do a mass change on checked users on the following attributes:

* is_active
* is_staff
* is_superuser

FIXES: APER-3004

- [ ] Run `make test-karma`
- [ ] All tests pass


credentials/apps/core/admin.py
```diff
@@ -1,17 +1,125 @@
...
+
+    @admin.action(description=_("Activate selected entries"))
+    def activate_selected(self: admin.ModelAdmin, request, queryset):
+        """Activate the selected entries."""
+        count = queryset.count()
+        queryset.update(is_active=True)
+        model_name = self.__class__.__name__
+
+        if count == 1:
+            message = _("1 {model_name} entry was successfully activated.")
```
justinhynes, line 28: These look like f-strings, but I don't see the `f` before the quotes. Am I missing something?

justinhynes: Blurgh..  nevermind, I didn't go down far enough 😅. Ignore me.

justinhynes (approved):
This is pretty slick. I'll have to remember it in the future. LGTM!

PR #2503, opened 2024-06-17
feat!: removing the never-used Dockerfile-testing

deborahgu:
* remove the dockerfile
* remove the appropriate makefile targets
* remove information about the unused testing docker from testing.rst

FIXES: APER-3434

- [ ] Run `make test-karma`
- [ ] All tests pass


justinhynes (approved):
Awesome, thanks for doing this and it looks good to me. This was all I thought needed to be deleted (thanks for catching the docs piece :D).

justinhynes:
One _tiny_ meganit, based on the conventional commits used, I think this should be something like `feat!` or `!feat`? To denote that a feature is being removed, which might be helpful when the Sumac release notes are being compiled?

PR #2511, opened 2024-06-27
feat: removing visible_date attribute

deborahgu:
removing the reliance on the visible_date attribute

* re-factoring some tests, not just to make them work with the new code, but also because some of these tests were never ideal. They weren't all testing the right things, some of them never tested the code path that used certificate available date, and definitely a bunch have non-useful names and no comments. Several were testing impossible (or, definitionally broken) situations like program certificates with no corresponding course certificates.
* added django stubs for type checking
* added some type checking

FIXES: APER-3238

- [ ] Run `make test-karma`
- [ ] All tests pass


credentials/apps/credentials/api.py
```diff
@@ -235,7 +232,7 @@ def get_credential_dates(user_credentials, many):
         return get_credential_visible_date(user_credentials, use_date_override=True)
 
 
-def process_course_credential_update(user: "User", course_run_key: str, mode: str, credential_status: str) -> None:
```
deborahgu, line 238: once I actually turned on mypy I realized that this was an illegal way to do type check, so removing it.

credentials/apps/credentials/tests/test_api.py
```diff
@@ -1,9 +1,10 @@
 from unittest.mock import patch
 
 import ddt
-from ddt import unpack
```
deborahgu, line 4: Consistency with imports from  `ddt` .

credentials/apps/credentials/tests/test_api.py
```diff
@@ -156,7 +158,9 @@ def setUp(self):
             )
             for course_run in self.course_runs
         ]
-        self.program_cert = ProgramCertificateFactory.create(program_uuid=self.program.uuid, site=self.site)
```
deborahgu, line 159: the program certificate factory doesn't actually create the foreign key when you add the UUID, and the `filter_visible`  method that the tests are all being driven down now that the `visible_date`  circuit breaker is gone required the presence of the foreign key. This would have been discovered earlier if all of the tests had `USE_CERTIFICATE_AVAILABLE_DATE` variants but such is life.

credentials/apps/credentials/tests/test_api.py
```diff
@@ -139,6 +140,7 @@ def test_get_program_certificates_with_ids_multiple(self):
         assert result[1] == self.program_cert2
 
 
+@override_waffle_switch(settings.USE_CERTIFICATE_AVAILABLE_DATE, active=True)
```
deborahgu, line 143: I believe I can take all of these out before I make this into a real PR.

.github/workflows/ci.yml
```diff
@@ -106,7 +106,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
         with:
-          python-version: "3.8"
+          python-version: "3.11"
```
justinhynes, line 109: Good catch. I wonder if we should update the `code_coverage` action too, which has Python 3.8 defined (line 76 above).

.coveragerc
```diff
@@ -10,3 +10,7 @@ omit =
     *admin.py
     *static*
     *templates*
+[report]
+exclude_lines = 
+    pragma: no cover
+    if TYPE_CHECKING:
```
justinhynes, line 16: Is this incomplete? I have been trained to expect something after an `if` statement or a colon. 😂

deborahgu: yeah but I feel kind of guilty because, well, look at [the source of where I learned how to do this](https://github.com/nedbat/coveragepy/issues/831#issuecomment-517778185).

credentials/apps/credentials/tests/test_api.py
justinhynes: Thanks for making these make a bit more sense!

justinhynes (approved):
LGTM. Thanks for making this dream a bit closer to being a reality.

credentials/apps/records/tests/test_utils.py
```diff
@@ -251,20 +265,13 @@ def test_get_credentials_both_empty(self):
...
         assert program_results == []
 
-    def test_get_credentials_program_only(self):
-        for course_cert in self.course_certs:
-            course_cert.delete()
-        for course_run in self.course_runs:
-            course_run.delete()
-        course_results, program_results = get_credentials(self.user.username)
-        assert course_results == []
-        assert program_results[0] == self.program_user_credential
```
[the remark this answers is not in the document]
deborahgu: This is mostly what I was looking at, and also, the circumstance where there is a program certificate and yet there are no corresponding courses… What's the correct behavior even supposed to be in that circumstance? I assume both that that is bad data, but also possibly bad data that can happen? But I'd also assume that this test result in the deleted test is _wrong_ -- if you do have a program credential and no corresponding course credentials, I don't believe a program credential _should_ display. Am I wrong?

PR #2515, opened 2024-07-09
feat: the signatory is not mandatory on coursecertificate configurations

deborahgu:
Credentials IDA requires signatories on `ProgramCertificate` configurations.  However, on `CourseCertificate` configurations, not all sites will care  if a signatory is present. The certificates themselves don't render in the credentials IDA, and this is cloned data from the system of record. Although this has always been nullable in the ORM because of the way django handles nullability for `ManyToManyField`,  the django admin enforces the presence of a signatory.

This keeps the  signatory required for `ProgramCertificate`, but makes it blankable for `CourseCertificate`.

FIXES: APER-3494

- [ ] Run `make test-karma`
- [ ] All tests pass


credentials/apps/credentials/models.py
```diff
@@ -204,6 +204,11 @@ class CourseCertificate(AbstractCertificate):
     .. no_pii: This model has no PII.
     """
 
+    # We require signatories on program certificates, However, on course certificate configurations,
+    # not all sites will care. The certificates themselves don't render in the credentials IDA,
+    # and this is cloned data from the system of record.
```
jsnwesson, line 209: The comment below is kinda hard to parse. My suggestion is really just a restructuring of what you wrote if it helps:

```suggestion
    # Not all sites will require signatories in course certificate configurations, but signatories are required in program certificates configurations.
    # This is cloned data from the system of record, as the course certificates don't render in the Credentials IDA.
```

deborahgu: this makes sense to me, I committed your suggestion.

credentials/apps/credentials/tests/test_admin.py
```diff
@@ -0,0 +1,122 @@
...
+        self.data["course_run"] = self.course_run.id
+        self.data["course_id"] = self.course_run.key
+        self.sc = SiteConfigurationFactory()
+        self.data["site"] = self.sc.site.id
+
+    def test_save_course_certificate_configuration_without_signatory_passes(self):
+        """
+        a CourseCertificate configuration without a signatory passes validation
+        """
+        expected = 302
```
justinhynes, line 110: Curious as to why we expect a 302. Just a function of how the call works after saving an entry? A redirect to the list of course cert configs in Django admin? 

Also, I know the method for saving is different than the Program cert test, but would it be easier/better to check that the admin form has no errors?

justinhynes (approved):

credentials/apps/credentials/tests/test_admin.py
deborahgu: apparently saving successfully gives a redirect to the list page, I verified this is normal and expected in both reality and django tests.

PR #2539, opened 2024-08-01
feat: management command to synchronize is_active

deborahgu:
this creates a management command to synchronize is_active between LMS and credentials. It only goes in one direction (the learner on credentials to is_active = True if that is the case on LMS), because the alternative is not a problem and is much more costly to update.

FIXES: APER-3346

### DIFF
--- a/credentials/apps/core/management/commands/make_is_active_match_platform.py
+++ b/credentials/apps/core/management/commands/make_is_active_match_platform.py
@@ -1,6 +1,6 @@
 """
-Django managment command to make the is_active status on a credentials user
-be True if it the user's LMS is_active status is True.
+Django management command to make the is_active status on a credentials user
+be True if the user's LMS is_active status is True.
 
 If the credentials user is_active=True we don't need to synchronize because that
 inconsistency is fine. But there can be user problems if they can log in to the
--- a/credentials/apps/core/management/commands/make_is_active_match_platform.py
+++ b/credentials/apps/core/management/commands/make_is_active_match_platform.py
@@ -2,9 +2,9 @@
 Django management command to make the is_active status on a credentials user
 be True if the user's LMS is_active status is True.
 
-If the credentials user is_active=True we don't need to synchronize because that
+If the credentials user is_active=True but the LMS user is_active=False, we don't need to synchronize because that
 inconsistency is fine. But there can be user problems if they can log in to the
-LMS but the credentials user has been set inactive, so we fix inconsistencies in
+LMS but the credentials user has been set inactive, so we have to fix inconsistencies in
 that one direction.
 """
 
--- a/credentials/apps/core/management/commands/make_is_active_match_platform.py
+++ b/credentials/apps/core/management/commands/make_is_active_match_platform.py
@@ -62,7 +62,7 @@ class Command(BaseCommand):
         logger.info(f"using {site_configs.site.domain} for user queries")
 
         # loop over users in batches
-        logger.warning(f"Start processing {count_users} IDs with is_active=True")
+        logger.warning(f"Start processing {count_users} inactive Credentials user accounts")
         for x in range(0, count_users, offset):
             slice_size = min(count_users, x + offset)
             curr_users = inactive_credentials_users[x:slice_size]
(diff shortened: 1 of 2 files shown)


credentials/apps/core/management/commands/make_is_active_match_platform.py
```diff
@@ -0,0 +1,123 @@
+"""
+Django managment command to make the is_active status on a credentials user
+be True if it the user's LMS is_active status is True.
```
jsnwesson, line 3:
```suggestion
Django management command to make the is_active status on a credentials user
be True if the user's LMS is_active status is True.
```

credentials/apps/core/management/commands/make_is_active_match_platform.py
```diff
@@ -0,0 +1,123 @@
...
+
+If the credentials user is_active=True we don't need to synchronize because that
+inconsistency is fine. But there can be user problems if they can log in to the
+LMS but the credentials user has been set inactive, so we fix inconsistencies in
+that one direction.
```
jsnwesson, line 8:
```suggestion
If the credentials user is_active=True but the LMS user is_active=False, we don't need to synchronize because that
inconsistency is fine. But there can be user problems if they can log in to the
LMS but the credentials user has been set inactive, so we have to fix inconsistencies in
that one direction.
```

credentials/apps/core/management/commands/make_is_active_match_platform.py
```diff
@@ -0,0 +1,123 @@
...
+
+        site_configs = SiteConfiguration.objects.first()
+        if site_id:
+            # multiple are sites managing different users
+            logger.info(f"using site id {site_id} for user queries")
+            site_configs = SiteConfiguration.objects.get(site__id=site_id)
+        logger.info(f"using {site_configs.site.domain} for user queries")
+
+        # loop over users in batches
+        logger.warning(f"Start processing {count_users} IDs with is_active=True")
```
justinhynes, line 65: For this particular log message, should this say `Start processing (num users) IDs with is_active=False`? We're logging the number of users with inactive accounts in Credentials at this point, right?

credentials/apps/core/management/commands/make_is_active_match_platform.py
```diff
@@ -0,0 +1,123 @@
...
+                time.sleep(pause)
+
+        logger.info("Finished is_active matching.")
+
+    def get_is_active(
+        self,
+        users: List["AbstractUser"],
+        site_config: SiteConfiguration,
+    ) -> Dict[str, bool]:
+        """Get is_active status for a list of users.
```
justinhynes, line 85: Nit: Could this be a tiny bit more specific? "Get is_active status from the LMS for a list of users"?

credentials/apps/core/management/commands/make_is_active_match_platform.py
```diff
@@ -0,0 +1,123 @@
...
+        logger.warning(f"Start processing {count_users} IDs with is_active=True")
+        for x in range(0, count_users, offset):
+            slice_size = min(count_users, x + offset)
+            curr_users = inactive_credentials_users[x:slice_size]
+            is_active_statuses = self.get_is_active(curr_users, site_configs)
+            for user in curr_users:
+                if user.username in is_active_statuses and user.is_active != is_active_statuses[user.username]:
+                    self.enable_user(user, verbosity, dry_run)
+                else:
+                    logger.error(f"Could not get is_active for user {user.username}")
```
justinhynes, line 74: Do you think we should log the internal Credentials ID here over username?

(I would vastly prefer username here as its a way more usable log message, but I know we try not to log PII wherever possible)

credentials/apps/core/management/commands/make_is_active_match_platform.py
deborahgu: yes, and actually I'm rephrasing the sentence to make it less ambiguous

credentials/apps/core/management/commands/make_is_active_match_platform.py
deborahgu: nice catch

justinhynes (approved):
LGTM! Thanks for the adjustments :)

PR #2543, opened 2024-08-09
docs: update README and Makefile

justinhynes:
A PR to cleanup the Credentials README. Additionally, this removes some Makefile targets that are no longer valid, their functionality had been removed in an earlier PR.

- [ ] Run `make test-karma`
- [ ] All tests pass

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ TOX = ''
 	extract_translations dummy_translations compile_translations fake_translations pull_translations \
 	detect_changed_source_translations validate_translations check_translations_up_to_date \
 	check_keywords pii_check coverage \
-	quality_and_translations_tests_suite unit_tests_suite docs
+	docs
 
 ifdef TOXENV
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -10,7 +10,7 @@ TOX = ''
 	extract_translations dummy_translations compile_translations fake_translations pull_translations \
 	detect_changed_source_translations validate_translations check_translations_up_to_date \
 	check_keywords pii_check coverage \
-	quality_and_translations_tests_suite unit_tests_suite docs
```
justinhynes, line 13: These Makefile targets were used as part of an experimental feature that built a Docker container that mimics how tests were run in CI. The Dockerfile and related functionality were removed in an earlier PR, and the targets in the Makefile were already removed.

This is just some leftover cruft we missed in the earlier cleanup.

MaxFrank13 (approved):

PR #2644, opened 2024-11-26
feat: allowing the passing of a logging formatter

deborahgu:
* allow the configuration of logging to pass a custom format string for the logs
* adding type hints and a more complete docstring to the modified method

FIXES: APER-3805

### DIFF
--- a/credentials/settings/utils.py
+++ b/credentials/settings/utils.py
@@ -25,7 +25,8 @@ def get_logger_config(
     service_variant: str = "credentials",
     format_string: str = "",
 ):
-    """Return the appropriate logging config dictionary, to be assigned to the LOGGING var in settings.
+    """
+    Return the appropriate logging config dictionary, to be assigned to the LOGGING var in settings.
 
     Arguments:
         log_dir (str): Location for log files


credentials/settings/utils.py
```diff
@@ -16,22 +16,29 @@ def get_env_setting(setting):
...
+    dev_env: bool = False,
+    debug: bool = False,
+    local_loglevel: str = "INFO",
+    service_variant: str = "credentials",
+    format_string: str = "",
 ):
-    """
-    Return the appropriate logging config dictionary. You should assign the
-    result of this to the LOGGING var in your settings.
+    """Return the appropriate logging config dictionary, to be assigned to the LOGGING var in settings.
```
jsnwesson, line 28: Not really a nit, just wasn't sure if in fact you do want a new line for the comment, or at least a space between `"""Return`

```suggestion
    """
    Return the appropriate logging config dictionary, to be assigned to the LOGGING var in settings.
```

jsnwesson (approved):
Approved, other than that one bit about the documentation syntax

credentials/settings/utils.py
deborahgu: I was quite honestly struggling with this. I was following [Open edX Python Style Guide](https://docs.openedx.org/en/latest/developers/references/developer_guide/style_guides/python-guidelines.html#docstrings)  which gives an example using the line break format you suggest, but that doc references [PEP 257's](https://peps.python.org/pep-0257/)  multiline docstring format, which goes without a line break.

Oh, except now that I read more carefully, I see that this is called out as an explicit exception:

Note. There are some exceptions: The summary line is on the second line, including single-line comments (see below). Use the full word Arguments, although Args is also acceptable.

So that's a nice call out. Thank you!

PR #2648, opened 2024-12-03
chore: remove urllib3 constraint

justinhynes:
Removing the constraint on `urllib3` no longer downgrades boto and its dependencies. I think it is safe to remove this constraint now.

- [ ] Run `make test-karma`
- [ ] All tests pass

### DIFF
--- a/requirements/constraints.txt
+++ b/requirements/constraints.txt
@@ -11,20 +11,12 @@
 # Common constraints for edx repos
 -c common_constraints.txt
 
-# Pinning urllib3 to versions < 2.x as this conflicts with boto. This constraint will be re-evaluated as part of
-# APER-2422
-urllib3<2
-
-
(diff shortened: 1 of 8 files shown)


requirements/constraints.txt
```diff
@@ -11,20 +11,12 @@
 # Common constraints for edx repos
 -c common_constraints.txt
 
-# Pinning urllib3 to versions < 2.x as this conflicts with boto. This constraint will be re-evaluated as part of
-# APER-2422
-urllib3<2
-
-
-# Temporary to Support the python 3.11 Upgrade
-backports.zoneinfo;python_version<"3.9"  # Newer versions have zoneinfo available in the standard library
```
justinhynes, line 20: We no longer need this constraint either now that Credentials has been updated to use Python 3.11.

PR #2652, opened 2024-12-04
feat: management command allowing you to revoke certificates

deborahgu:
Management command to enable batch revocation of certificates. It defaults to program_certificates, which is the most  likely use case, but because some UserCredentials might not be synchronized from the LMS, does allow override of the credential type.

* Adds a new model and django admin so this script can be run in an automated fashion
* this required bringing in config_models

FIXES: APER-3799


credentials/apps/credentials/management/commands/revoke_certificates.py
```diff
@@ -0,0 +1,153 @@
...
+class Command(BaseCommand):
+    """
+    Management command to revoke certificates.
+
+    Given a certificate ID and a list of users, revoke that certificate ID
+    for those users.
+
+    Example usage:
+
+    $ ./manage.py revoke_certificates --lms_user_ids 867 5309 925
```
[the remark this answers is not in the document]
deborahgu: done.

credentials/apps/credentials/management/commands/tests/test_revoke_certificates.py
```diff
@@ -0,0 +1,242 @@
...
+        self.assertTrue(any(expected_substring in s for s in cm.output))
+
+    def test_validation(self):
+        """Test that the input validators are correct"""
+        # fake user
+        expected = "None of the given lms_user_ids maps to a real user*"
+        with self.assertRaisesRegex(CommandError, expected):
+            call_command(Command(), "--lms_user_ids", "8675309", "--credential_id", "123")
+
+        # real user, doesn't mapped to a real credential
```
justinhynes, line 202: nit: `map` over `mapped` in this comment

requirements/all.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with Python 3.11
+# This file is autogenerated by pip-compile with Python 3.12
```
justinhynes, line 2: Has Credentials been upgraded to use Python 3.12? I thought we were still only using 3.11 here.

deborahgu: regenerating with my devstack  version of Python, thanks

justinhynes (approved):

PR #2657, opened 2024-12-05
feat: public dockerfiles are deprecated, so removing

deborahgu:
This removal supports the  [deprecation of public dockerfiles](https://github.com/openedx/public-engineering/issues/263).

* this removes the public dockerfile
* the corresponding workflow action was removed in #2618

FIXES: APER-3435


justinhynes (approved):
Thanks for handling this! It looks the `push-docker-image` workflow was already removed, so this should be all that remains!

PR #2675, opened 2025-01-02
feat: [FC-0065] add Accredible integration

kyrylo-kh:
**Description:**
Add integration with Accredible service for badges


openedx-webhooks:
Thanks for the pull request, @kyrylo-kh!

### What's next?

*Please work through the following steps to get your changes ready for engineering review:*

#### :radio_button: Get product approval

If you haven't already, [check this list](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3875962884/How+to+submit+an+open+source+contribution+for+Product+Review#Does-my-contribution-require-Product-Review%3F) to see if your contribution needs to go through the product review process.

- If it does, you'll need to submit a product proposal for your contribution, and have it reviewed by the [Product Working Group](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3449028609/Product+Working+Group).
    - This process (including the steps you'll need to take) is documented [here](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3875962884/How+to+submit+an+open+source+contribution+for+Product+Review#Product-Review-Process).
- If it doesn't, simply proceed with the next step.

#### :radio_button: Provide context

To help your reviewers and other members of the community understand the purpose and larger context of your changes, feel free to add as much of the following information to the PR description as you can:

- Dependencies
  > This PR must be merged before / after / at the same time as ...
- Blockers
  > This PR is waiting for OEP-1234 to be accepted.
- Timeline information
  > This PR must be merged by XX date **because** ...
- Partner information
  > This is for a course on edx.org.
- Supporting documentation
- Relevant [Open edX discussion forum](https://discuss.openedx.org/) threads





#### :radio_button: Get a green build

If one or more checks are failing, continue working on your changes until this is no longer the case and your build turns green.



#### :radio_button: Let us know that your PR is ready for review:

### Who will review my changes?


This repository is currently maintained by `@2U-aperture`.  Tag them in a comment and let them know that your changes are ready for review.


### Where can I find more information?

If you'd like to get more details on all aspects of the review process for open source pull requests (OSPRs), check out the following resources:

- [Overview of Review Process for Community Contributions](https://docs.openedx.org/en/latest/developers/references/developer_guide/process/FAQ-about-pull-requests.html)
- [Pull Request Status Guide](https://docs.openedx.org/en/latest/developers/references/developer_guide/process/pull-request-statuses.html)
- [Making changes to your pull request](https://docs.openedx.org/en/latest/documentors/how-tos/make_changes_to_your_pull_request.html)

### When can I expect my changes to be merged?

Our goal is to get community contributions seen and reviewed as efficiently as possible.

However, the amount of time that it takes to review and merge a PR can vary significantly based on factors such as:

- The size and impact of the changes that it introduces
- The need for product review
- Maintenance status of the parent repository

:bulb: *As a result it may take up to several weeks or months to complete a review and merge your PR.*

deborahgu (commented):
Non-blocking concern: It seems sensible to have verifiable credentials as a django app in the IDA, but I would expect it to be generic, with the Accredible integration as a plugin (which could be provided in an `openedx` plugin repo). Otherwise it seems like we're recreating the same technical debt issues that we've been removing for the last couple of years, putting a specific 3rd party SaaS  provider's code, and a specific  implementation, directly into the IDA. 

Especially since we have a good plug-in framework now, I would think this would be an opportunity for a formal `openedx` plugin (which would also be modeling a great practice for the community).

deborahgu:
> Non-blocking concern

The more I think about it, the more I feel like this _is_ a blocking concern. Certainly on the `edx-platform` side, there's an acknowledgement that built in optional integrations to braze, hubspot, salesforce, etc. are grandfathered in for now but will inevitably be formally deprecated, and new third-party APIs should be written as plug-ins.

Unless there's a reason we can't, it seems to me that moving forward, all new integrations to a specific third-party vendor should be in a plugin. In this case, again, in a public, official, `openedx` plugin repo. Is there a reason we can't or shouldn't do that?

deborahgu:
> @deborahgu Does the Credentials IDA support plugins using the event driving model, or at all today?

I'm not positive. Certainly I know we can run [our standard edx_django_utils plug-in utilities](https://github.com/openedx/edx-django-utils/blob/master/edx_django_utils/plugins/docs/how_tos/how_to_create_a_plugin_app.rst).  And I can see that `openedx_events`  is already [used in the badges implementation on credentials](https://github.com/openedx/credentials/blob/cb51cc457bd9141edf62051ee378f6a7957b5b92/credentials/apps/badges/signals/handlers.py#L10), so I'd be surprised if we could  not use the hooks and filters implementation.

The OEP for hooks [specifically references Accredible as its very first use case](https://open-edx-proposals.readthedocs.io/en/latest/architectural-decisions/oep-0050-hooks-extension-framework.html#use-cases), and all of the existing filter  documentation on [design practices](https://github.com/openedx/openedx-filters/blob/main/docs/decisions/0007-filter-design-practices.rst) and [how-tos](https://github.com/openedx/openedx-filters/blob/main/docs/how-tos/using-filters.rst) use "third-party certificate verification service" as their use case.  (obviously  those examples all take place on `edx-platform`, where the course certificates are currently generated.) But that reference implantation is only a couple of years ago, so I assume that `credentials`  was part of the infrastructural plan  around it.

kyrylo-kh:
@deborahgu Answer for your concern
1. The proposal did not mention moving the customization for Credly into a separate plugin; therefore, the new integration with Accredible was added alongside the existing one. This approach allows for reusing existing functionality and avoids duplicating the same code (such as the badge issuance mechanism, base classes for badges, configurations, issuers, etc.).
2. All the new logic related to badges is located in its own separate app at credentials/apps/badges, and changes to the main codebase (excluding translations and documentation) outside of it are minimal. Therefore, in my opinion, it makes sense to keep everything in one place rather than splitting it between this location and a separate plugin.
3. We created a separate plugin for quick and convenient badge configuration, specifically designed for the current solution.

This implementation was agreed with the @e0d

e0d:
@deborahgu I agree with your overall argument here.  However, I'd like to propose that we capture the work as technical debt rather than block this change on the refactoring.  I'll do my best to prioritize getting in done in the next ~2 quarters or so.  @cmltaWt0 I believe you have done a back of the envelope estimate?  Could that be the basis of a ticket to capture the required work here?

deborahgu:
@e0d I'm amenable to  looking at this work with the plan to switch to a plugin based-system.  I will go ahead and review this code now on its own merits, and I  will look forward to seeing  the new scope of work. (cc @justinhynes @kbuchanan-2u)

credentials/apps/badges/signals/handlers.py
```diff
@@ -73,16 +73,22 @@ def handle_badge_completion(sender, username, badge_template_id, **kwargs):  # p
 
     logger.debug("BADGES: progress is complete for %s on the %s", username, badge_template_id)
 
-    CredlyBadgeTemplateIssuer().award(username=username, credential_id=badge_template_id)
+    if origin == CredlyBadgeTemplate.ORIGIN:
+        CredlyBadgeTemplateIssuer().award(username=username, credential_id=badge_template_id)
+    elif origin == AccredibleGroup.ORIGIN:
+        AccredibleBadgeTemplateIssuer().award(username=username, credential_id=badge_template_id)
```
deborahgu, line 79: seems like it's worth a test

credentials/apps/badges/signals/handlers.py
```diff
@@ -73,16 +73,22 @@ def handle_badge_completion(sender, username, badge_template_id, **kwargs):  # p
...
 
     - username
     - badge template ID
     """
 
-    CredlyBadgeTemplateIssuer().revoke(badge_template_id, username)
+    if origin == CredlyBadgeTemplate.ORIGIN:
+        CredlyBadgeTemplateIssuer().revoke(badge_template_id, username)
+    elif origin == AccredibleGroup.ORIGIN:
+        AccredibleBadgeTemplateIssuer().revoke(badge_template_id, username)
```
deborahgu, line 94: same.

deborahgu (approved):
this looks good to me! I'm approving, although I did put in some nonblocking feedback.

credentials/apps/badges/signals/handlers.py
kyrylo-kh: There are already existing tests for completion and regression, these tests check the effect from handlers
https://github.com/openedx/credentials/blob/master/credentials/apps/badges/tests/test_signals.py#L23

Probably you want to test directly handlers instead of effect from them?

credentials/apps/badges/accredible/api_client.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def sync_groups(self, site_id: int) -> int:
+        """
+        Pull all groups for a given Accredible API config.
+
+        Args:
+            site_id (int): ID of the site.
+
+        Returns:
+            int | None: processed items.
```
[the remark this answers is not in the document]
kyrylo-kh: Fixed
https://github.com/openedx/credentials/compare/9855ff4769de50e496ca6a69ca954322ecdf36d7..f081a9c83a7572d1b5cb846fffb4b428a157ec47

credentials/apps/badges/accredible/utils.py
```diff
@@ -0,0 +1,45 @@
...
+            This will be the URL for the sandbox environment if `USE_SANDBOX` is
+            set to a truthy value in the configuration;
+            otherwise, it will be the production environment's URL.
+    """
+    accredible_config = settings.BADGES_CONFIG["accredible"]
+
+    if accredible_config.get("USE_SANDBOX"):
+        return accredible_config["ACCREDIBLE_SANDBOX_API_BASE_URL"]
+
+    return accredible_config["ACCREDIBLE_API_BASE_URL"]
```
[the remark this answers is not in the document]
kyrylo-kh: Added tests for this method
https://github.com/openedx/credentials/compare/9855ff4769de50e496ca6a69ca954322ecdf36d7..f081a9c83a7572d1b5cb846fffb4b428a157ec47

credentials/apps/badges/management/commands/sync_accredible_groups.py
```diff
@@ -0,0 +1,54 @@
...
+        Handle the command.
+        """
+        DEFAULT_SITE_ID = 1
+        api_configs_to_sync = []
+
+        site_id = options.get("site_id")
+        api_config_id = options.get("api_config_id")
+
+        if site_id is None:
+            self.stdout.write(f"Side ID wasn't provided: assuming site_id = {DEFAULT_SITE_ID}")
```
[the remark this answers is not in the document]
kyrylo-kh: Fixed https://github.com/openedx/credentials/compare/9855ff4769de50e496ca6a69ca954322ecdf36d7..f081a9c83a7572d1b5cb846fffb4b428a157ec47

e0d:
@kyrylo-kh  I see your question.  @deborahgu do you have a feeling about the test coverage question?

I also notice that there are some test failures here that will need to be addressed.

credentials/apps/badges/signals/handlers.py
deborahgu: that makes sense to me.

credentials/apps/badges/signals/handlers.py
kyrylo-kh: The same from previous issue, we have tests for signals and for effect from handlers

credentials/apps/badges/signals/handlers.py
```diff
@@ -63,7 +63,7 @@ def handle_requirement_regressed(sender, username, **kwargs):
 
 
 @receiver(BADGE_PROGRESS_COMPLETE)
-def handle_badge_completion(sender, username, badge_template_id, **kwargs):  # pylint: disable=unused-argument
+def handle_badge_completion(sender, username, badge_template_id, origin, **kwargs):  # pylint: disable=unused-argument
```
GlugovGrGlib, line 66: @kyrylo-kh Can you please update docstring here and for the following function.
Please mention new parameter `origin`, and add a note that for now we select Issuer class manually based on the parameter `origin`, and in the future it is planned to be replaced with extensible implementation to support other issuers implemented using plugins.

GlugovGrGlib (changes requested):
Looks good, only several text updates required to emphasize that the updated solution is intended to support more badge providers in the future.

GlugovGrGlib:
@e0d @deborahgu Thank you for raising valid concerns about plugin capabilities in the badges app and the overall `credentials` IDA. I also agree that, eventually, extensions for badges and verifiable credentials sharing mechanisms should become installable djangoapps that will use extension points in the base apps to modify use cases without requiring code changes in credentials. This decision has already been documented https://github.com/openedx/credentials/blob/master/docs/decisions/0006-plugin-extensions.rst, but is yet to be completed.
With Credly, this backend was closely coupled with the base "Open edX" badges implementation. In this PR, some work was done to decouple Credly from the base implementation and provide a reference implementation for a separate, simpler badges backend. However, some additional work is still required to bring the plugin capabilities in `credentials` to the level of `edx-platform`.

I also believe that the best course of action is not to block the current solution for extending the badges backend, as it already provides more value to the community than the initial implementation of the Credly backend. But to finalize the plugin system in the `credentials` IDA and enable it for badges and verifiable credentials. After this, I believe we should even consider migrating both Credly and Accredible backends to separate djangoapps.

deborahgu:
this makes sense to me, and once all the tests that exist are passing I am comfortable merging this. I am made more comfortable by @e0d's  assurance that there will definitely be a  relatively short-term plan to migrate to the plug-in-based model.

credentials/apps/badges/models.py
```diff
@@ -88,11 +89,12 @@ class BadgeTemplate(AbstractCredential):
     uuid = models.UUIDField(unique=True, default=uuid.uuid4, help_text=_("Unique badge template ID."))
     name = models.CharField(max_length=255, help_text=_("Badge template name."))
     description = models.TextField(null=True, blank=True, help_text=_("Badge template description."))
-    icon = models.ImageField(upload_to="badge_templates/icons", null=True, blank=True)
+    icon = models.ImageField(upload_to="badge_templates/icons", null=True, blank=True, max_length=255)
     origin = models.CharField(max_length=128, null=True, blank=True, help_text=_("Badge template type."))
     state = StatusField(
         choices_name="STATES",
         help_text=_("Credly badge template state (auto-managed)."),
+        null=True,
```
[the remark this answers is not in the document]
kyrylo-kh: Done https://github.com/openedx/credentials/compare/1021bd7cc86ff77bdf4920abb690c62e7cf8d5c2..c3dc5c7e298a028d4bdbcf89d99b62679d582a26

credentials/apps/badges/signals/handlers.py
kyrylo-kh: Done
https://github.com/openedx/credentials/compare/1021bd7cc86ff77bdf4920abb690c62e7cf8d5c2..c3dc5c7e298a028d4bdbcf89d99b62679d582a26

GlugovGrGlib (approved):

GlugovGrGlib:
@kyrylo-kh Please rebase PR and regenerate translation files.

PR #2694, opened 2025-02-06
chore: update paragon 23

MaxFrank13:
Updates Paragon to v23

The two big changes to Paragon that will effect this repo are the deprecation of FontAwesome support and the change from SCSS to CSS variables.

There were only two variables being pulled from paragon: `$spacer` and `$card-border-radius`

It's worth noting that there are still SCSS variables present in this repo, but that is because there is a `_utilities.scss` file that houses a number of SCSS variables used throughout other files ([source](https://github.com/openedx/credentials/blob/master/credentials/static/sass/_utilities.scss))

Example print preview of cert when on Paragon 23 in local environment:
[Professional Certificate _ Open edX.pdf](https://github.com/user-attachments/files/18769379/Professional.Certificate._.Open.edX.pdf)

- [ ] Run `make test-karma`
- [ ] All tests pass

### DIFF
--- a/credentials/apps/credentials_theme_openedx/static/sass/_print-rendering.scss
+++ b/credentials/apps/credentials_theme_openedx/static/sass/_print-rendering.scss
@@ -36,7 +36,7 @@
             @include make-col-ready();
             @include make-col(8);
             margin-bottom: 0;
-            margin-top: $spacer * .5;
+            margin-top: calc(var(--pgn-spacing-spacer-base) * .5);
             display: inline-block;
             vertical-align: middle;
           }
@@ -87,7 +87,7 @@
 
         .signatory {
           margin: 0;
-          margin-bottom: $spacer * .5;
+          margin-bottom: calc(var(--pgn-spacing-spacer-base) * .5);
           width: 100%;
           &:last-child {
             margin-bottom: 0;
@@ -165,7 +165,7 @@
     .accomplishment-summary,
     .accomplishment-statement-detail {
       margin: 0;
-      margin-bottom: $spacer * 1.5;
+      margin-bottom: calc(var(--pgn-spacing-spacer-base) * 1.5);
     }
   }
 
--- a/credentials/static/sass/_components.scss
+++ b/credentials/static/sass/_components.scss
@@ -61,7 +61,7 @@
       @extend .p-1;
       display: inline-block;
       border: 1px solid $gray-100;
-      border-radius: $card-border-radius;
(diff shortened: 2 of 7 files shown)


credentials/static/sass/_lib.scss
```diff
@@ -3,5 +3,4 @@
 
 // About: third party libraries and dependencies import
 
-@import '~font-awesome/scss/font-awesome';
```
MaxFrank13, line 6: FontAwesome is no longer supported by Paragon which is where these fonts were coming from — we were never explicitly installing it as a dependency.

credentials/static/sass/_components.scss
```diff
@@ -61,7 +61,7 @@
       @extend .p-1;
       display: inline-block;
       border: 1px solid $gray-100;
-      border-radius: $card-border-radius;
+      border-radius: var(--pgn-size-card-border-radius-base);
```
MaxFrank13, line 64: This was one of the only variables being used from Paragon. In v23, Paragon has switched from SCSS variables (i.e. `$variable-name`) to CSS variables (i.e. `var(--pgn-variable-name)`)

credentials/apps/credentials_theme_openedx/static/sass/_print-rendering.scss
```diff
@@ -165,7 +165,7 @@
     .accomplishment-summary,
     .accomplishment-statement-detail {
       margin: 0;
-      margin-bottom: $spacer * 1.5;
+      margin-bottom: calc(var(--pgn-spacing-spacer-base) * 1.5);
```
MaxFrank13, line 168: See note about switching from SCSS to CSS variables

justinhynes: Any chance we can get a post-change screenshot of the "printed" cert? or a print preview? Just to confirm things look as expected after the changes?

credentials/static/sass/_lib.scss
justinhynes: Dumb question... does this mean we are no longer using FA? How does this affect the icons rendered for sharing/printing a program cert? Are the icons being replaced by something else?

MaxFrank13: Icons from Paragon are all React based so they were never used in these frontends. The only icons in Credentials are instead from something like [this import](https://github.com/openedx/credentials/blob/master/credentials/templates/base.html#L20-L22).

justinhynes (approved):

PR #2710, opened 2025-02-25
feat: adding pathway status

deborahgu:
course-discovery is adding the concept of retired  pathways, which it will expose with a status field through the API. We still want to ingest retired and unpublished pathways, because we don't want to break any of our existing foreign key relationships. However, we don't want to expose them to either the Learner Record MFE  or the program dashboard.

Still todo:
* verification
*  documentation updates
* ADR

FIXES: APER-3929

- [ ] Run `make test-karma`
- [ ] All tests pass


credentials/apps/records/api.py
```diff
@@ -100,7 +100,12 @@ def _get_transformed_pathway_data(program, user):
     program_pathways_set = frozenset(program_pathways)
 
     user_credit_pathways = (
-        UserCreditPathway.objects.select_related("pathway").filter(user=user, pathway__in=program_pathways_set).all()
+        UserCreditPathway.objects.select_related("pathway")
+        .filter(
+            user=user,
+            pathway__in=program_pathways_set,
+        )
+        .all()
```
justinhynes, line 108: Would we want to add a check on the status to be empty or published here?

I believe this would result is not including unpublished or retired pathways to a learner when fetching their program record details.

deborahgu: So the way I made the changes to `copy_catalog`, a retired or unpublished pathway is no longer linked to a program at all, which means  that it can't make a `UserCreditPathway`. it doesn't delete old UserCreditPathway objects, but I don't think it should, because those are effectively audit logs, as I understand it. I suppose technically there could be a race condition where:
1. The user has their learner record open so they can see the button to send the record to an un-retired pathway
2.  We retire the pathway while that page is open
3. They press the button and the query comes back and returns no longer valid `UserCreditPathway` objects.

 But I'm not sure I'm that worried about that race condition? What do you think?

justinhynes (approved):
Looks good.

PR #2712, opened 2025-02-28
feat: making pathway status non-nullable

deborahgu:
It gets rid of so many problems if we just do a data migration on old pathway  status, and don't make any assumptions that it might be null or blank.

- [x] Run `make test-karma`
- [x] All tests pass


credentials/apps/catalog/migrations/0017_pathway_status_never_empty.py
```diff
@@ -0,0 +1,40 @@
...
+        ("catalog", "0016_pathway_status"),
+    ]
+
+    operations = [
+        migrations.RunPython(populate_pathway_status, unpopulate_pathway_status),
+        migrations.AlterField(
+            model_name="pathway",
+            name="status",
+            field=models.CharField(
+                choices=[("retired", "retired"), ("unpublished", "unpublished"), ("published", "published")],
```
justinhynes, line 35: Would this end up with the tags being hard-coded english?

credentials/apps/catalog/migrations/0017_pathway_status_never_empty.py
```diff
@@ -0,0 +1,40 @@
+# Generated by Django 4.2.19 on 2025-02-28 16:28
+
+from django.db import migrations, models
+
+
+def populate_pathway_status(apps, schema_editor):
+    Pathway = apps.get_model("catalog", "Pathway")
+    for pathway in Pathway.objects.exclude(status__isnull=True).exclude(status__exact=""):
```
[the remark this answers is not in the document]
deborahgu: doh, that's just a bug (ie. typed the wong QuerySet operator), and it didn't turn up in my testing because the filter chain is _also_ a bug (didn't `or` them).

credentials/apps/catalog/migrations/0017_pathway_status_never_empty.py
deborahgu: that was following prior art, but also, it's generated directly from the `choices` attribute in the model, and is [hard-coded in discovery as well](https://github.com/openedx/course-discovery/blob/d505b0ace9bf5a40ee7631b91186e36181d1e954/course_discovery/apps/course_metadata/migrations/0347_pathway_status.py#L25).

justinhynes (approved):
Thank you!

PR #2722, opened 2025-03-14
chore: downgrading python in ci  for consistency

deborahgu:
the docs aren't building on github because a valid version of `didkit` won't install, and I think it might be because we upgraded Python to bleeding edge but only for the `ci` workflows.

- [ ] Run `make test-karma`
- [ ] All tests pass


sarina:
Ahhh this makes sense. 👍🏻

PR #2747, opened 2025-04-15
feat!: remove download_url from a UserCredential

deborahgu:
When the credentials IDA was created, two conditions were different from how they are now:

1. There was a plan to move all certificates functionality into the IDA.
1. PDF certificates were still being generated.

Neither of those conditions holds true today.

No code exists that has ever propagated `download_url` from `edx-platform` to the Credentials IDA.

The question was raised on both slack and [on discourse](https://discuss.openedx.org/t/removing-download-url-from-credentials-ida/15350)  as to whether any instances of EdX  using the Credentials IDA had implemented some  other feature that used `download_url`,  and there were no such instances. Removing this field removes a point of confusion.

Migration will be a second PR, to avoid timing problems.

FIXES: APER-3931

- [ ] Run `make test-karma`
- [ ] All tests pass


credentials/apps/credentials/migrations/0033_remove_download_url.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 4.2.20 on 2025-04-15 15:42
```
justinhynes, line 1: I don't think we want to include the migration at this point. If the guidance from "[Everything About Database Migrations](https://openedx.atlassian.net/wiki/spaces/AC/pages/23003228/Everything+About+Database+Migrations#EverythingAboutDatabaseMigrations-Howtodropacolumn)" still holds true, we want to split this change into two separate PRs: 

1. The first PR involves removing usage of the columns and deleting the field from the model. 
2. The second PR involves dropping the column with the migration.

justinhynes (approved):
LGTM! Thanks for breaking the PR up.

PR #2749, opened 2025-04-16
feat: cherry pick FA icons used in Credentials UIs

justinhynes:
Previous to this change, when collecting static assets, we would include ALL of the FA icons. This PR is an attempt to limit the icons in our bundle to only the icons used in the UI.

As part of this change, the `fontawesomefree` app is no longer needed, so it has been removed.

- [ ] Run `make test-karma`
- [ ] All tests pass

### DIFF
--- a/requirements/pip.txt
+++ b/requirements/pip.txt
@@ -10,7 +10,7 @@ wheel==0.45.1
 # The following packages are considered to be unsafe in a requirements file:
 pip==24.2
     # via
-    #   -c /home/runner/work/credentials/credentials/requirements/common_constraints.txt
+    #   -c /edx/app/credentials/credentials/requirements/common_constraints.txt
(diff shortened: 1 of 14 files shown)


requirements/pip.txt
```diff
@@ -10,7 +10,7 @@ wheel==0.45.1
 # The following packages are considered to be unsafe in a requirements file:
 pip==24.2
     # via
-    #   -c /home/runner/work/credentials/credentials/requirements/common_constraints.txt
+    #   -c /edx/app/credentials/credentials/requirements/common_constraints.txt
```
justinhynes, line 13: This changed because I ran "make upgrade" from my Credentials container running in devstack, and was not run from CI through a GHA.

deborahgu (approved):

PR #2780, opened 2025-05-22
feat: Remove newrelic package from base dependencies

timmc-edx:
`newrelic` is not directly used by this repository, and edx-django-utils provides a pluggable telemetry API for any deployers who use New Relic.

See DEPR for details: https://github.com/openedx/public-engineering/issues

This PR is composed of two commits. The first is the result of `make upgrade`, and the second removes `newrelic` from the `.in` files and runs `make upgrade` again (so you can see the actual effect).

----

- [x] Run `make test-karma`
- [ ] All tests pass

### DIFF
--- a/requirements/all.txt
+++ b/requirements/all.txt
@@ -535,10 +535,6 @@ mysqlclient==2.2.7
     # via
     #   -r requirements/dev.txt
     #   -r requirements/production.txt
-newrelic==10.12.0
-    # via
-    #   -r requirements/dev.txt
-    #   -r requirements/production.txt
 nodeenv==1.9.1
     # via -r requirements/production.txt
 oauthlib==3.2.2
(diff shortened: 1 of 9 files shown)


timmc-edx:
When I ran `make test-karma` I got the following error, which is... probably unrelated, given the nature of the changes?

```
  analytics module
    ✓ should pull attributes from data attributes and call window.analytics.track

  sharing module
    initializeFacebook
      ✓ should initialize the Facebook SDK
      ✓ should update the share button
      ✓ should log a page view
      ✓ should handle the share button not existing
Firefox 136.0 (Ubuntu 0.0.0) ERROR
  An error was thrown in afterAll
  SyntaxError: import declarations may only appear at top level of a module
  credentials/static/js/fontawesome.js:2
```

deborahgu (approved):

PR #2881, opened 2025-10-14
fix: forking work for workflows

deborahgu:
per our internal setup documentation:

* Changing the branch for the workflows to match the release branch
* disabling the  timed update workflow which we are no longer handling here

- [ ] Run `make test-karma`
- [ ] All tests pass


deborahgu:
fat finger again

PR #2895, opened 2025-11-07
feat: add pagination support to fetch all badge templates

luisfelipec95:
## Description

This PR enhances the synchronization process of badge templates from the Credly API by implementing pagination support.

Previously, the fetch_badge_templates function only retrieved the first page of results (up to 50 templates).
Now, the function iterates through all available pages to fetch every active badge template, ensuring no data is missed during synchronization.

## Supporting information

- No migrations; no operator action required for default deployments.

## Testing instructions

- Executed sync_organization_badge_templates for an organization with more than 50 templates.
- Confirmed that the total number of local records matches the total returned by the API.

### DIFF
--- a/credentials/apps/badges/tests/test_api_client.py
+++ b/credentials/apps/badges/tests/test_api_client.py
@@ -71,10 +71,13 @@ class CredlyApiClientTestCase(TestCase):
 
     def test_fetch_badge_templates(self):
         with mock.patch.object(CredlyAPIClient, "perform_request") as mock_perform_request:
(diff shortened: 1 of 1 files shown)


e0d:
@luisfelipec95 Thanks for this.  It looks like some of the tests are failing, can you have a look?

luisfelipec95:
> @luisfelipec95 Thanks for this. It looks like some of the tests are failing, can you have a look?

Sure, the failing tests were due to a mismatch between the expected return structure and the actual output. I updated the test accordingly, and the failures should now be resolved.

